PR #198, opened 2015-01-16
Connect code to GDAL and PROJ data within wheels.

sgillies:
The _drivers.pyx module has also been modernized. It's now the same as Rasterio's.

In setup, package that data when envvar PACKAGE_DATA=1. GDAL_CONFIG and PROJ_LIB envvars can be used to point to the sources.

Also switch over to setuptools.extension as in Rasterio.

### DIFF
--- a/fiona/_drivers.pyx
+++ b/fiona/_drivers.pyx
@@ -1,30 +1,38 @@
 # The GDAL and OGR driver registry.
 # GDAL driver management.
 
+import os
+import os.path
 import logging
+import sys
 
 from six import string_types
+
 
 cdef extern from "cpl_conv.h":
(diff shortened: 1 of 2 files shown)


sgillies:
@jwass will you review this one and merge if you concur? This makes batteries-included Fiona binaries possible, just like with Shapely and Rasterio. Example of the new build command: https://github.com/sgillies/frs-wheel-builds/blob/master/Makefile#L97

jwass:
Looks good. `setuptools` nuances and binary distribution usually not in my wheelhouse... but I was able to build the wheel and install in a fresh virtual env w/o any problems.

Issue #199, opened 2015-01-22
ValueError: No data available at path


allhailwesttexas:
I'm having trouble with Fiona 1.4.8 on OSX Yosemite. I get a `ValueError: No data available at path "test.shp"`. If I switch to version 1.4.7, I get a bunch of polygons printed to screen, as expected.

Here is my test program:
```python
import fiona

vecfile = "test.shp"
with fiona.open(vecfile, 'r') as fi:
    features = [x for x in fi]
print features
```
Here is the traceback:
```python
Traceback (most recent call last):
  File "test.py", line 4, in <module>
    with fiona.open(vecfile, 'r') as fi:
  File "/Users/stensmyr/.virtualenvs/fionatest/lib/python2.7/site-packages/fiona/__init__.py", line 139, in open
    encoding=encoding, layer=layer, vsi=vsi, archive=archive)
  File "/Users/stensmyr/.virtualenvs/fionatest/lib/python2.7/site-packages/fiona/collection.py", line 135, in __init__
    self.session.start(self)
  File "fiona/ogrext.pyx", line 359, in fiona.ogrext.Session.start (fiona/ogrext.c:6649)
ValueError: No data available at path 'test.shp'
```

I'm installing with pip. My best lead is that 1.4.8 installs from a wheel, whereas previous versions install from setup.py.

snorfalorpagus:
Do you experience the same issue if you provide an absolute path? e.g. `/Users/snorf/Desktop/test.shp`?

allhailwesttexas:
Yes, same error.

sgillies:
Until I'm able to confirm, the work around is `pip install --no-use-wheels fiona` to compile the extension modules locally using your system's GDAL.

sgillies:
@allhailwesttexas confirmed. Details soon.

sgillies:
So, this is only a problem with my wheel builds. I'm going to yank the wheels and re-upload once I've got them straightened out. Sorry for the trouble, @allhailwesttexas!

sgillies:
After https://github.com/sgillies/frs-wheel-builds/commit/f444f4a60af77383e11dada8d08c26e63c1041ce, I've built and uploaded new *tested* wheels.

allhailwesttexas:
No problem, happy to help! Thanks for the quick response.

PR #200, opened 2015-01-22
Make sure to pass dep libs to build_ext.

sgillies:
Towards a fix of #199.


sgillies:
Turns out 85252c2 is a terrible idea. ffaf83c is a good idea, we'll keep that.

Issue #192, opened 2014-12-10
Have to use OpenFileGDB


r-barnes:
Using Fiona 1.4.8 with GDAL 1.11.0, I get the following error message when trying to open and ESRI GDB directory:

    >>> fin      = fiona.open(filename)
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/usr/local/lib/python2.7/dist-packages/fiona/__init__.py", line 139, in open
        encoding=encoding, layer=layer, vsi=vsi, archive=archive)
      File "/usr/local/lib/python2.7/dist-packages/fiona/collection.py", line 148, in __init__
        self.guard_driver_mode()
      File "/usr/local/lib/python2.7/dist-packages/fiona/collection.py", line 160, in guard_driver_mode
        raise DriverError("unsupported driver: %r" % driver)
    fiona.errors.DriverError: unsupported driver: u'OpenFileGDB'

Opening `/usr/local/lib/python2.7/dist-packages/fiona/collection.py` and replacing `("FileGDB", "raw")` with `("OpenFileGDB", "r")` resolves the issue for me.

My expected behavior is that there would be no error and the file would simply open, as it does when using `("OpenFileGDB", "r")`. Therefore, I would suggest making the switch part of the codebase.

Are there Good Reasons not to? Or is there some way to use both drivers by having Fiona fall back `OpenFileGDB` if `FileGDB` fails?

sgillies:
@r-barnes @snorfalorpagus here's what I came up with: https://github.com/Toblerity/Fiona/blob/ef1d3f7f5e35c62e6d4ff157ec2725c279044683/tests/test_read_drivers.py#L20-L25. You can specify exactly one driver to try when opening or you can try several drivers in a specified order.

So, `fiona.open(filename, driver='GeoJSON')` is the equivalent of osgeo.ogr's `ogr.GetDriverByName('GeoJSON').Open(filename)` and `fiona.open(filename, enabled_drivers=['GeoJSON', 'ESRI Shapefile'])` is the equivalent of:

```python
for name in ['GeoJSON', 'ESRI Shapefile']:
    ds = ogr.GetDriverByName(name).Open(filename)
    if ds:
        break
```

I decided not to take on selective driver registration as it goes hard against the grain of OGR's C API. A call to `fiona.open()` continues to register all drivers if they haven't already been registered.

r-barnes:
[...]

PR #203, opened 2015-01-25
Control over drivers when opening datasets to read.

sgillies:
Closes issue #192.

See test_read_drivers.py for examples. Documentation forthcoming.


sgillies:
@snorfalorpagus yes? no?

PR #215, opened 2015-03-08
Added a BytesCollection class.

youngpm:
This class is a subclass of Collection whose contructor takes a buffer
that we'd like mapped to a virtual file.  The class creates such a file
and passes its path to the underlying Collection, where it can be read
from.  The class cleans up the virtual file when close is called (and by
contruction when the context manager exits).


youngpm:
Here's a working version of the feature.  With this PR, I'm just looking for feedback (I haven't added tests, etc).  To try it out, you can do something like this if you're in the Fiona root dir:

```python
import fiona

with open('tests/data/esri.json') as source:
    buffer = source.read()

with fiona.BytesCollection(buffer) as source:
    recs = [r for r in source]

It should print out something about opening and closing a file (I'll remove the debug statements when I finalize), and your records should be populated!

File: fiona/collection.py [hunk 1]
```diff
@@ -406,6 +407,33 @@ def __del__(self):
...
+    """BytesCollection takes a string buffer of bytes and maps that to
+    a virtual file that can then be opened by fiona.
+    """
+    def __init__(self, buffer):
+        """Takes string buffer whose contents is something we'd like
+        to open with Fiona and maps it to a virtual file.
+        """
+        # Hold a reference to the buffer, as bad things will happen if
+        # it is garbage collected while in use.
+        self.buffer = buffer
```
snorfalorpagus: Avoid using builtin names for variables.

sgillies:
Fantastic. Looks good to me!

```python
>>> import fiona
>>> with open('tests/data/esri.json') as f:
...     data = f.read()
...
>>> with fiona.BytesCollection(data) as src:
...     features = list(src)
...
Opened /vsimem/325a691522dd49fcaae21b45ac26c5d3
Closed /vsimem/325a691522dd49fcaae21b45ac26c5d3
>>> len(features)
2
>>> features[0]
{'geometry': {'type': 'Polygon', 'coordinates': [[(-13555360.4191, 5911556.581), (-13489311.5669, 5898227.932), (-13423477.4153, 5884426.3329), (-13357867.1993, 5870158.6064), (-13292489.9099, 5855431.779), (-13310286.7337, 5791381.4753), (-13327810.7467, 5727748.1244), (-13345069.9061, 5664524.4905), (-13362071.9117, 5601703.4638), (-13425742.0169, 5615460.9307), (-13489617.0175, 5628791.9699), (-13553688.6321, 5641690.4875), (-13617948.3761, 5654150.574), (-13602646.9571, 5717848.4135), (-13587119.9125, 5781976.6214), (-13571360.1713, 5846543.2654), (-13555360.4191, 5911556.581)]]}, 'type': 'Feature', 'id': '1', 'properties': OrderedDict([(u'OBJECTID', 1), (u'Name', u'p045r028_7t19991002.met'), (u'MinPS', 28.5), (u'MaxPS', 342.0), (u'LowPS', 28.5), (u'HighPS', 114.0), (u'Shape', None), (u'Shape_Length', 1058012.723772), (u'Shape_Area', 69884678121.7441)])}
```

I'm looking forward to the tests.

youngpm:
I fixed the use of buffer added a bunch of tests, mostly stolen from the tests for the Collections class.  I'm failing the following with the BytesCollection class, is this expected or should I dig a bit deeper?

```python
    def test_getitem_iter_combo(self):
        i = iter(self.c)
        f = next(i)
        f = next(i)
        self.failUnlessEqual(f['id'], "1")
        f = self.c[0]
        self.failUnlessEqual(f['id'], "0")
        f = next(i)
        self.failUnlessEqual(f['id'], "2")
```

sgillies:
@youngpm I see the same exception.

```
    def test_getitem_iter_combo(self):
        i = iter(self.c)
        f = next(i)
        f = next(i)
        self.failUnlessEqual(f['id'], "1")
        f = self.c[0]
        self.failUnlessEqual(f['id'], "0")
        f = next(i)
>       self.failUnlessEqual(f['id'], "2")
E       AssertionError: '1' != '2'
```

You can delete this test. OGR's JSON format driver simply can't support mixed use of `next()` and `__getitem__()` in a single session. There's no random access to records in a JSON text.

youngpm:
I removed the offending test, looks like Python 2.7 is good.  The Python 3 tests are failing, I need to deal with the string/bytes discrepancy!

sgillies:
FWIW, Fiona uses [six](http://pythonhosted.org/six/) as a 2/3 compatibility module. As at https://github.com/Toblerity/Fiona/blob/master/fiona/crs.py#L13.

youngpm:
Alright I think its all fixed up.  Turns out it was passing the virtual file paths to the wrapped VSI functions that was causing the problem, luckily you guys already dealt with that elsewhere from the looks of it!

sgillies:
Great! I'll take a closer look in the morning. Looks very mergable.

youngpm:
Sounds good, I'll get on it!

youngpm:
I pushed updates reflecting your comments, let me know what you think!

sgillies:
:+1:

PR #229, opened 2015-05-14
Restore Python 2.6 support

groutr:
The original nested with statement in inspector.py was syntax that was not introduced until Python 2.7.  This alternative, equivalent version is specified in PEP343.

Without these changes, this module will fail with syntax error on Python 2.6


sgillies:
:+1:

PR #231, opened 2015-05-29
Enhancements to setup.py

groutr:
Share the enhancement to setup.py that rasterio recently gained.


groutr:
Split into two commits so that way I can apply a patch to setup.py in the conda package for Fiona.

sgillies:
:+1:

Issue #47, opened 2013-06-01
ValueError: Null layer error on geojson?


blackmad:
playing with some osm extracts. 

converted shapefile to geojson
ogr2ogr -f GeoJSON paris-lines.json paris.osm-line.shp
copied to: http://dump.twofishes.net/paris-lines.json

blackmad@ns4005169:~/sites/the-neighborhoods-project$ ./make-holes.py ~/osm/paris-lines.json 
Traceback (most recent call last):
  File "./make-holes.py", line 12, in <module>
    source = fiona.open(sys.argv[1], 'r')
  File "/usr/local/lib/python2.7/dist-packages/Fiona-0.12.1-py2.7-linux-x86_64.egg/fiona/**init**.py", line 99, in open
    c = Collection(path, mode, encoding=encoding)
  File "/usr/local/lib/python2.7/dist-packages/Fiona-0.12.1-py2.7-linux-x86_64.egg/fiona/collection.py", line 92, in **init**
    self.session.start(self)
  File "ogrext.pyx", line 591, in fiona.ogrext.Session.start (src/fiona/ogrext.c:9701)
ValueError: Null layer

weirdly I'm hitting the same issue with geojson I'm constructing by hand that qgis is happy to render: http://dump.twofishes.net/paris.osm-roads.shp.json

PR #246, opened 2015-07-23
Properly handle JSON when reading OGR string fields

### LINKED ISSUE #245: Deserialize string field values that contain JSON text
Background: #244.

Given a GeoJSON feature property like `'foo': {'bar': true}`, OGR's field value accessor returns the string `"{\"bar\": true}"`. That string should be deserialized into a Python object. Approach: test if the string begins with a curly bracket and an escaped quote and try `json.loads()` if it does.

sgillies:
Closes #245.

Now (after we merge this) when you open a GeoJSON file which has features with object (not just number or string) properties using the OGR GeoJSON driver (ie with `fiona.open()`) those properties will return to you as Python objects (not JSON-encoded strings).

NB that there is no "dict" type in OGR, so when serializing, you must set the type in the schema to "str". See the example in the new tests.

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -1,6 +1,7 @@
 # These are extension functions and classes using the OGR C API.
 
 import datetime
+import json
 import locale
 import logging
 import os
@@ -165,6 +166,7 @@ cdef class FeatureBuilder:
                     key,
                     ograpi.OGR_Fld_GetType(fdefn))
                 continue
+
             # TODO: other types
             fieldtype = FIELD_TYPES_MAP[fieldtypename]
             if not ograpi.OGR_F_IsFieldSet(feature, i):
@@ -173,14 +175,26 @@ cdef class FeatureBuilder:
                 props[key] = ograpi.OGR_F_GetFieldAsInteger(feature, i)
             elif fieldtype is float:
                 props[key] = ograpi.OGR_F_GetFieldAsDouble(feature, i)
+
             elif fieldtype is text_type:
                 try:
                     val = ograpi.OGR_F_GetFieldAsString(feature, i)
-                    props[key] = val.decode(encoding)
+                    val = val.decode(encoding)
                 except UnicodeDecodeError:
                     log.warn(
                         "Failed to decode %s using %s codec", val, encoding)
-                    props[key] = val
+
+                # Does the text contain a JSON object? Let's check.
+                # Let's check as cheaply as we can.
+                if val.startswith('{'):
+                    try:
+                        val = json.loads(val)
+                    except ValueError as err:
+                        log.warn(str(err))
+
+                # Now add to the properties object.
+                props[key] = val
+
             elif fieldtype in (FionaDateType, FionaTimeType, FionaDateTimeType):
                 retval = ograpi.OGR_F_GetFieldAsDateTime(
                     feature, i, &y, &m, &d, &hh, &mm, &ss, &tz)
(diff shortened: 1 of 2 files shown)


sgillies:
@geowurster are you familiar enough with OGR's mangling of complicated GeoJSON to review this one?

geowurster:
@sgillies Not directly but my understanding is that GeoJSON supports complex fields while OGR does not, however the user can just stuff a string encoded JSON object into a string field, assuming the field is wide enough to not truncate the data.

If that's all the required prior knowledge, then I would be happy to review.

sgillies:
@geowurster yep, that's all there is to it.

geowurster:
@sgillies On it.

fiona/ogrext.pyx
```diff
@@ -173,14 +175,26 @@ cdef class FeatureBuilder:
...
-                    props[key] = val.decode(encoding)
+                    val = val.decode(encoding)
                 except UnicodeDecodeError:
                     log.warn(
                         "Failed to decode %s using %s codec", val, encoding)
-                    props[key] = val
+
+                # Does the text contain a JSON object? Let's check.
+                # Let's check as cheaply as we can.
+                if val.startswith('{'):
```
geowurster: @sgillies This feature is framed as being GeoJSON specific but it actually works with any driver.  Is this intentional?

On one hand I see this as another killer Fiona feature but also a potential source of confusion, especially because a mismatched field width will truncate the JSON object on write, making it impossible to decode.  Fiona speaks GeoJSON so it definitely makes sense from that perspective but AFAIK none of the other enabled drivers directly support complex fields.  Maybe some documentation is needed?

geowurster:
@sgillies This feature is framed as being GeoJSON specific but it actually works with any driver.  Is this intentional?

On one hand I see this as another killer Fiona feature but also a potential source of confusion, especially because a mismatched field width will truncate the JSON object on write, making it impossible to decode.  Fiona speaks GeoJSON so it definitely makes sense from that perspective but AFAIK none of the other enabled drivers directly support complex fields.  Maybe some documentation is needed?

sgillies:
@geowurster really good question. No, it's not my intention this work for all drivers and I'm glad you caught this.

Earlier in the history of this project we didn't enable the GeoJSON driver at all because its support for complex properties (especially when writing) is so poor, but I was convinced otherwise and enabled it (see #47).

I think it makes sense to only call `json.loads()` if we're using the GeoJSON driver.

Fiona programs like the CLI should continue to bypass OGR's GeoJSON driver when writing GeoJSON.

geowurster:
@sgillies I think thats the right solution for now, although `json.dumps()` should also only be called when working with the GeoJSON driver.

If this feature turns out to be more useful in general then it would be far easier to enable it for all drivers rather than disable it.

geowurster:
@sgillies Something else I have thought about a few times but never tried, is implementing a special Fiona GeoJSON driver to make some of the Fiona internals easier to implement.  There are certainly pros and cons to such a big diversion from OGR, and Fiona would need some internal refactoring to take full advantage, but it would also open the door to serialized formats like MsgPack and maybe ProtoBuf for faster I/O.

geowurster:
@sgillies Don't forget about this one too :smile:

geowurster:
@sgillies I should have been clearer, this PR was not fully ready for merge. My bad. 

The JSON deserialization happens for any driver rather than just GeoJSON. See the thread for more info.

sgillies:
Oops. I think the odds of this causing problems until we fix it in 1.6.2 are pretty small.

Issue #251, opened 2015-07-26
Generalize feature sequence I/O and RS support


geowurster:
Currently these features are implemented partially in `fio cat` and `fio load` but adding them to the Fiona API as something like `fiona.SequenceStream()` would make features like https://github.com/geowurster/fio-buffer/issues/4 easier to support.  

We could maybe have `fiona.open()` return this object with something like:

```python
with fiona.open(file or '-', mode, sequence=bool, rs=bool) as f:
    ...
```

although properties like `crs` and `bounds` would behave differently than `Collection()`. 

At the very least we could support:

```python
with fiona.SequenceStream(infile) if i_sequence else fiona.open(infile) as src:
    with fiona.SequenceStream(outfield, 'w', rs=use_rs) if o_sequence else fiona.open(outfile, 'w', **meta) as dst:
        for feature in src:
            dst.write(feature)
```

geowurster:
Closing this in favor of a more generic external implementation.  See comment in #252 for reference.

sgillies:
@geowurster what would you think about working on it in the https://github.com/geojson org?

geowurster:
Definitely seems like the most appropriate place to me!

PR #252, opened 2015-07-29
Consolidate GeoJSON feature stream I/O

geowurster:
In progress.  Feedback welcome.

@sgillies This is a rough draft (without unittests) of #251 that implements a new `fiona.collections.FeatureStream()` file-like object for working with GeoJSON feature streams.  If it is decided that this object doesn't really belong in Fiona, this about how I planned on implementing https://github.com/geowurster/fio-buffer/issues/4, so feel free to shoot this down.

**Generate some sample data**

```console
$ fio cat tests/data/coutwildrnp.shp > features.geojson
```

**Inspect it with Python**

```python
import fiona
with fiona.open('features.geojson', sequence=True) as src:
    print(next(src))
```

RS prefixes are supported as well as reading from `stdin` and writing to `stdout`:

```python
import fiona
with fiona.open('-', sequence=True) as src, fiona.open('-', 'w', sequence=True, use_rs=True) as dst:
    for feat in src:
        dst.write(feat)
```

### DIFF
--- a/fiona/__init__.py
+++ b/fiona/__init__.py
@@ -69,7 +69,7 @@ import logging
 import os
 from six import string_types
 
-from fiona.collection import Collection, BytesCollection, vsi_path
+from fiona.collection import Collection, BytesCollection, FeatureStream, vsi_path
 from fiona._drivers import driver_count, GDALEnv, supported_drivers
 from fiona.odict import OrderedDict
 from fiona.ogrext import _bounds, _listlayers, FIELD_TYPES_MAP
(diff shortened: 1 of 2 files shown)


sgillies:
@geowurster Let's develop an [IO-like class](https://docs.python.org/3/library/io.html#io.FileIO) for GeoJSON streams in a different project. It won't require GDAL/OGR, and might get wider use if it didn't entail GDAL. Additionally, there are a few things to figure out yet about feature sequences (https://github.com/geojson/draft-geojson/issues/64 for reference), and we can go faster and more recklessly outside of Fiona :)

geowurster:
@sgillies Sounds good.  I'll drop this in `fio-buffer` for now but feel free to find me when you get a repo set up for general purpose GeoJSON IO.

geowurster:
@sgillies Would it be helpful if I worked this into a PR for `py-feature-sequences` to start a discussion?

sgillies:
@geowurster :+1:

geowurster:
@sgillies I'm in.  Thanks!

PR #261, opened 2015-08-23
Add fixes to support gdal2

rbuffat:
This PR includes:
- Adding of 64bit types. (For Python 2.7 64bit Integers are mapped to the 32bit int type, not long)
- Session.sync resets previous errors before executing OGR_DS_SyncToDisk (*)
- If OGR_L_TestCapability of OLC_STRINGSASUTF8 is successful, utf-8 is chosen as encoding and not  OGR_DETECTED_ENCODING

(*) : The tests in test_issue177 (tests.test_schema.FieldTruncationTestCase) respectively test_schema (tests.test_schema.ShapefileSchema) return as last error message
"Normalized/laundered field name: 'a_fieldname' to 'a_fieldnam'" respectively "Value 521280485 of field TotalWages of feature 0 not successfully written. Possibly due to too larger number with respect to field width"

These errror messages do not have a connection with OGR_DS_SyncToDisk. Thus past errors are reset. A test was added for value 521280485 of total wages for test_schema (tests.test_schema.ShapefileSchema

### DIFF
--- a/fiona/ograpi.pxd
+++ b/fiona/ograpi.pxd
@@ -147,4 +147,5 @@ cdef extern from "ogr_api.h":
 
 cdef extern from "cpl_error.h":
     void CPLErrorReset()
+    char * CPLGetLastErrorMsg()
 
--- a/fiona/ograpi.pxd
+++ b/fiona/ograpi.pxd
@@ -5,6 +5,7 @@
 cdef extern from "gdal.h":
     char * GDALVersionInfo (char *pszRequest)
 
+
 cdef extern from "gdal_version.h":
     int    GDAL_COMPUTE_VERSION(int maj, int min, int rev)
 
@@ -143,3 +144,7 @@ cdef extern from "ogr_api.h":
     void *  OGROpenShared (char *path, int mode, void *x)
     int     OGRReleaseDataSource (void *datasource)
     OGRErr  OGR_L_SetNextByIndex (void *layer, long nIndex)
+
+cdef extern from "cpl_error.h":
+    void CPLErrorReset()
+
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -47,6 +47,8 @@ FIELD_TYPES = [
     'date',         # OFTDate, Date
     'time',         # OFTTime, Time
     'datetime',     # OFTDateTime, Date and Time
+    'int',          # OFTInteger64, Single 64bit integer
+    None,           # OFTInteger64List, List of 64bit integers
     ]
 
 # Mapping of Fiona field type names to Python types.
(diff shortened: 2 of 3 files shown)


rbuffat:
see  https://github.com/rbuffat/Fiona/tree/gdal2 for a branch with a fiona port to gdals rfc46 https://trac.osgeo.org/gdal/wiki/rfc46_gdal_ogr_unification

These changes will break gdal 1.x support.

sgillies:
Wow. I'm very interested in this. Despite it being a big breaking change and GDAL 2.0 not being widely available yet.

Perhaps it's time to start thinking about a Fiona 2.x that would require GDAL 2.x?

rbuffat:
@sgillies The big API changes are from RFC 46 (GDAL/OGR unification). This affects basically all OG_DS_* functions.  However the old API is still there, except marked as deprecated.  Thus it's still possible to support gdal2 for some time without breaking GDAL 1.x compatibility.  (See https://github.com/rbuffat/Fiona/commit/10fdeb8fc225b72c58d63ca389e68b05a8f77747 for the required changes to port to the new api)

The things that fail seem to be internal behaviour changes of gdal 2.0.  The commits above fixes the failed tests. All tests succeed with a self compiled gdal 2.0 froum source.


# UTF8
From the GDAL API:
> "OLCStringsAsUTF8 / "StringsAsUTF8": TRUE if values of OFTString fields are assured to be in UTF-8 format. If FALSE the encoding of fields is uncertain, though it might still be UTF-8."

I'm not sure why in 2 of 3 cases the encoding is set to OGR_DETECTED_ENCODING. The relevant commit seems to be https://github.com/Toblerity/Fiona/commit/9d06fd12d4ac1524419c3e76225ea65d001b4a75
There seems to be a related OGR Bug: https://trac.osgeo.org/gdal/ticket/4650
From my current point of view, the encoding can be set to utf-8, when StringsAsUTF8 is True.

# SyncToDisk
DS_SynctoDisk returns an error code for two tests out of test_schema.py. The error codes are not from synctodisk, but from earlier statements. I added a statement to reset the error codes befor exececuting synctodisk, The GDALFlushCache replacement of  synctodisk does also not return error codes.

The open question is, way this error codes remains in gdal2 and not gdal1. 

>"Normalized/laundered field name: 'a_fieldname' to 'a_fieldnam'"

This is the intended behaviour the test is looking for

>  "Value 521280485 of field TotalWages of feature 0 not successfully written

This sounds bad. But the value is actually written to the dataset. Or at least the test I added for the field TotalWages does not fail:
``` self.assertEqual(f['properties']['TotalWages'], 521280485.0)

# Fiona2

I think the thing to think about would be to unifiy the dataset managerment with rasterio (similar to gdal2). I think e.g. the virtual filesystem handling is implemented in fiona but not rasterio. Thus there could be some benefits from a shared code basis. However, I'm not too familiar with rasterio. A CI system with multiple gdal/ogr versions would be nice to test compability of new features with old versions.

rbuffat:
See https://github.com/Toblerity/Fiona/pull/275

Issue #263, opened 2015-08-30
Add CLI support for virtual file systems


sgillies:
The Python API supports zip and tar virtual file systems, but the CLI does not. I think it would go something like this, using a `--vfs` option.

```
fio info /testing/coutwildrnp.shp --vfs tar://tests/data/coutwildrnp.tar
```

sgillies:
Update: should this be done, we'll use URIs instead of the `--vfs` option:

`fio info tar://tests/data/coutwildrnp.tar!/testing/coutwildrnp.shp`

See also https://commons.apache.org/proper/commons-vfs/filesystems.html.

PR #264, opened 2015-08-31
Eager --vfs option and validators

sgillies:
Validators for `--vfs` and for the input argument.

Towards closing #263. About 30% of the way there.


sgillies:
Closing because we're not going the `--vfs` route after all.

PR #267, opened 2015-09-13
Enable travis ci to test for multiple gdal versions

rbuffat:
Initial commit to open discussion

Goal: 
- Enable travis to test fiona against different gdal versions
- Allow to test fiona against gdal 2.0

The PR includes:
- script to download, build from source and install different versions of gdal
- updated travis.yml
- travis chaching


rbuffat:
See also https://github.com/mapbox/rasterio/issues/478

rbuffat:
updated travis.yml from legacy to container to enable caching

rbuffat:
I think the PR https://github.com/Toblerity/Fiona/pull/267 would be ready to review. 

The last remaining issue is that coveralls seems to be in pending state for a long period of time.

Currently I only enabled to compile gdal 1.9.2 and 1.11.2. Gdal 2.0.0 compiles too on travis, but as Fiona currently does not support it and the tests fail I disabled it for now. 

I would suggest to test against the oldest gdal version that should be supported by Fiona as well as the newest of the Gdal1x, respectively 2x series.  Debian Squeeze (End of life February 2016) ships [gdal 1.6.3](https://packages.debian.org/squeeze/gdal-bin).

The build time is about 8 min when gdal needs to be compiled and around 4 min if the cache can be used. Possibly this could be improved by caching the python dependencies (==dist-packages folder) and removing a previous installed Fiona before it is newly built.

rbuffat:
Previously each job built only the gdal version that was used by this job. This lead to situations where already existing compiled gdal installations got deleted by travis. Now each job builds all gdal version if they are not yet cached.

Only compiled gdal installations are cached. This reduces cache size and improves installation speed.

Python dist-packages are cached now. This increases speed of the build process.

rbuffat:
The build time without cache is around 15min for each job, with cache, the time is reduced to well below 2 min. 

Regarding coveralls, do we need [PARALLEL BUILD WEBHOOK](https://coveralls.zendesk.com/hc/en-us/articles/203484329-Parallel-Build-Webhook)?

sgillies:
@rbuffat this looks great. When caching is working, the 6 element matrix runs in < 8 mins whereas the 3 element matrix from the master branch runs in 11.5 minutes. Seems like a win to me.

Any objections to me merging this @geowurster?

geowurster:
@sgillies @rbuffat :+1: This is great!  Would be good to get this in rasterio as well.

sgillies:
Engage!

PR #269, opened 2015-09-15
Add crs_wkt to meta property

### LINKED ISSUE #265: Datum information lost when converting file with datum D_Amersfoort
If I use Fiona to read a file in a coordinate system with D_Amersfoort as the datum and copy the metadata into a new file, I get D_Unknown as the datum.

Original .prj file:
```
PROJCS["RD_New",GEOGCS["GCS_Amersfoort",DATUM["D_Amersfoort",SPHEROID["Bessel_1841",6377397.155,299.1528128]],PRIMEM["Greenwich",0.0],UNIT["Degree",0.0174532925199433]],PROJECTION["Double_Stereographic"],PARAMETER["False_Easting",155000.0],PARAMETER["False_Northing",463000.0],PARAMETER["Central_Meridian",5.38763888888889],PARAMETER["Scale_Factor",0.9999079],PARAMETER["Latitude_Of_Origin",52.15616055555555],UNIT["Meter",1.0]]
```

.prj file after conversion:

```
PROJCS["Double_Stereographic",GEOGCS["GCS_Bessel 1841",DATUM["D_unknown",SPHEROID["bessel",6377397.155,299.1528128]],PRIMEM["Greenwich",0],UNIT["Degree",0.017453292519943295]],PROJECTION["Double_Stereographic"],PARAMETER["latitude_of_origin",52.1561605556],PARAMETER["central_meridian",5.38763888889],PARAMETER["scale_factor",0.9999079],PARAMETER["false_easting",155000],PARAMETER["false_northing",463000],UNIT["Meter",1]]
```

This leads to a pretty significant shift in the location of the geodata.

Using ogr2ogr to convert the same file behaves correctly, so it looks like this is a Fiona problem (or a bug in my use of Fiona). The code I'm using is [here](https://gist.github.com/mattwigway/18799e8d56e20f81a4ba).

Finally, the file I'm using is from Statistics Netherlands (page in dutch, but it's the buurt_2014.shp file inside the first link): http://www.cbs.nl/nl-NL/menu/themas/dossiers/nederland-regionaal/publicaties/geografische-data/archief/2015/wijk-en-buurtkaart-2014-art.htm .

sgillies:
Also extend fiona.open() and Collection so that a crs_wkt keyword argument will trump a crs keyword argument.

Closes #265

### DIFF
--- a/fiona/collection.py
+++ b/fiona/collection.py
@@ -30,6 +30,7 @@ class Collection(object):
             vsi=None,
             archive=None,
             enabled_drivers=None,
+            crs_wkt=None,
             **kwargs):
         
         """The required ``path`` is the absolute or relative path to
@@ -39,7 +40,9 @@ class Collection(object):
         a file.
         
         In ``mode`` 'w', an OGR ``driver`` name and a ``schema`` are
-        required. A Proj4 ``crs`` string is recommended.
+        required. A Proj4 ``crs`` string is recommended. If both ``crs``
+        and ``crs_wkt`` keyword arguments are passed, the latter will 
+        trump the former.
         
         In 'w' mode, kwargs will be mapped to OGR layer creation
         options.
@@ -55,6 +58,8 @@ class Collection(object):
             raise TypeError("invalid schema: %r" % schema)
         if crs and not isinstance(crs, (dict,) + string_types):
             raise TypeError("invalid crs: %r" % crs)
+        if crs_wkt and not isinstance(crs_wkt, string_types):
+            raise TypeError("invalid crs_wkt: %r" % crs_wkt)
         if encoding and not isinstance(encoding, string_types):
             raise TypeError("invalid encoding: %r" % encoding)
         if layer and not isinstance(layer, tuple(list(string_types) + [int])):
@@ -123,13 +128,12 @@ class Collection(object):
             elif 'geometry' not in schema:
                 raise SchemaError("schema lacks: geometry")
             self._schema = schema
-            
-            if crs:
-                if 'init' in crs or 'proj' in crs or 'epsg' in crs.lower():
-                    self._crs = crs
-                else:
-                    raise CRSError("crs lacks init or proj parameter")
-        
+
+            if crs_wkt:
+                self._crs_wkt = crs_wkt
+            elif crs:
+                self._crs = crs
+
         if driver_count == 0:
             # create a local manager and enter
             self.env = GDALEnv()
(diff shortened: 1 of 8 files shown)


sgillies:
@geowurster can review this for me? 

Existing usage isn't changed at all. I've just added the WKT representation of crs (`crs_wkt`) to a collection's `meta` and prioritized it over `crs`.

This raises the question: should WKT be our default CRS representation? It's more precise in some cases but is far too verbose and crufty for my taste. So I feel like it should not be the default.

geowurster:
@sgillies Will do.  Might take a few days though.

fiona/collection.py
```diff
@@ -123,13 +128,12 @@ def __init__(
             elif 'geometry' not in schema:
                 raise SchemaError("schema lacks: geometry")
             self._schema = schema
-            
-            if crs:
-                if 'init' in crs or 'proj' in crs or 'epsg' in crs.lower():
-                    self._crs = crs
-                else:
-                    raise CRSError("crs lacks init or proj parameter")
```
geowurster: @sgillies Looks like this `if/else` is the only check for an invalid `crs` parameter.  Without it any invalid `crs` is silently ignored for shapefiles and is silently changed to `EPSG:4326` for GeoJSON.  Didn't try any other formats and I'm guessing the GeoJSON change is actually happening in the OGR GeoJSON driver since that's technically the only supported CRS?

Additionally, a datasource that lacks CRS information has a `crs` property that is an empty dictionary and a `crs_wkt` property that raises a `ValueError`.  `fio info` doesn't have error handling for the `ValueError`, but it seems like that should probably be an empty string rather than an error.

geowurster:
@sgillies Agreed about WKT being too crufty for a default so I think this is probably a reasonable middle ground.  I feel like I have had issues in the past with ArcGIS generated `prj` files not being identical when round-tripped through OGR/QGIS or containing WKT that OGR couldn't parse, but when explicitly set with `ogr2ogr -a_srs` everything turns out alright.  I wonder if #265 is related?

Need to finish reviewing, but see my inline comment in the meantime.

geowurster:
@sgillies When a datasource doesn't have a CRS `Session.crs_wkt` raises an exception but `Session.crs` does not, which means any call to `Collection.meta` crashes.  [Looks like there's a check for null CRS in `Collection.get_crs()` but not for `Collection.get_crs_wkt()`](https://github.com/Toblerity/Fiona/blob/master/fiona/ogrext.pyx#L542).

To reproduce:

```console
$ ogr2ogr test-ds.shp tests/data/coutwildrnp.shp
$ rm test-ds.prj

$ fio insp test-ds.shp
Fiona 1.6.2 Interactive Inspector (Python 3.4.3)
Type "src.schema", "next(src)", or "help(src)" for more information.
>>> src.meta
ERROR:Fiona:CPLE_ObjectNull in b"Pointer 'hSRS' is NULL in 'OSRExportToWkt'.\n"
  File "<console>", line 1, in <module>
  File "/Users/wursterk/github/Fiona/fiona/collection.py", line 220, in meta
    'crs_wkt': self.crs_wkt}
  File "/Users/wursterk/github/Fiona/fiona/collection.py", line 211, in crs_wkt
    self._crs_wkt = self.session.get_crs_wkt()
  File "fiona/ogrext.pyx", line 599, in fiona.ogrext.Session.get_crs_wkt (fiona/ogrext.c:10930)
    raise ValueError("Null projection")
ValueError: Null projection
```

sgillies:
@geowurster thanks for catching that. I've added a test using a CRS-less dataset.

geowurster:
@sgillies :+1: Merge away.

As far as whether or not WKT should be our default representation, I think thats probably overkill.  Seems like CRS interoperability issues should really be solved by projection libraries.  Letting users drop in WKT definitions when needed should solve some edge cases and #270 will make this feel more welcomed.

sgillies:
Rats, wrong base branch. Replaced by #274.

PR #272, opened 2015-09-22
CLI for filtering features, fio filter

perrygeo:
I started writing this as a standalone script and realized it might be useful as a built in `fio` command.

The `fio filter` command reads GeoJSON features from stdin and writes the feature to 
stdout *if* the provided expression evalutates to `True` for that feature.

For example to create a geojson feature collection containing large polygons only:

    fio cat data.shp \
    | fio filter "f['properties']['area'] > 1000.0" \
    | fio collect > large_polygons.geojson

With the munch library, we can use dot notation for convenience 

    fio cat data.shp \
    | fio filter "f.properties.area > 1000.0" \
    | fio collect > large_polygons.geojson

The python expression is evaluated in a restricted namespace, what is allowed in that namespace is TBD. For now it's just (`sum`, `min`, `max`), the shapely `shape` constructor, the `math` module, and an object `f` representing the feature to be evaluated. 

The whole approach of eval in a restricted namespace seems safe enough for a command line tool but there might be other DSLs that could accomplish the same things (snuggs?)

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -20,6 +20,7 @@ Fiona's new command line interface is a program named "fio".
       collect  Collect a sequence of features.
       dump     Dump a dataset to GeoJSON.
       env      Print information about the rio environment.
+      filter   Filter GeoJSON features by python expression
       info     Print information about a dataset.
       insp     Open a dataset and start an interpreter.
       load     Load GeoJSON to a dataset in another format.
@@ -219,6 +220,26 @@ collection into a feature sequence.
     > 'each(data.features,function(o){console.log(JSON.stringify(o))})' \
     > | fio load /tmp/test-seq.shp --x-json-seq --driver Shapefile
 
+
+filter
+------
+The filter command reads GeoJSON features from stdin and writes the feature to 
+stdout *if* the provided expression evalutates to `True` for that feature. 
+
+The python expression is evaluated in a restricted namespace containing 3 functions 
+(`sum`, `min`, `max`), the `math` module, and an object `f` representing the
+feature to be evaluated. If the expression evaluates to a false-y value, the
+feature is not included in the output. Otherwise, the feature is printed
+verbatim. 
+
+For example 
+
+    fio cat data.shp \
+    | fio filter "f['properties']['area'] > 1000.0" \
+    | fio collect > large_polygons.geojson
+
+Would create a geojson file with only those features from `data.shp` where the
+area was over a given threshold.
 
 Coordinate Reference System Transformations
 -------------------------------------------
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -227,15 +227,17 @@ The filter command reads GeoJSON features from stdin and writes the feature to
 stdout *if* the provided expression evalutates to `True` for that feature. 
 
 The python expression is evaluated in a restricted namespace containing 3 functions 
-(`sum`, `min`, `max`), the `math` module, and an object `f` representing the
-feature to be evaluated. If the expression evaluates to a false-y value, the
-feature is not included in the output. Otherwise, the feature is printed
-verbatim. 
+(`sum`, `min`, `max`), the `math` module, the shapely `shape` function, 
+and an object `f` representing the feature to be evaluated. This `f` object allows
+access in javascript-style dot notation for convenience. 
+
+If the expression evaluates to a "truthy" value, the feature is printed verbatim.
+Otherwise, the feature is excluded from the output.
 
 For example 
 
     fio cat data.shp \
-    | fio filter "f['properties']['area'] > 1000.0" \
+    | fio filter "f.properties.area > 1000.0" \
     | fio collect > large_polygons.geojson
 
 Would create a geojson file with only those features from `data.shp` where the
--- a/fiona/fio/filter.py
+++ b/fiona/fio/filter.py
@@ -1,0 +1,59 @@
+import json
+import logging
+import math
+
+import click
+from cligj import use_rs_opt
+
+from .helpers import obj_gen
+
+
+@click.command(short_help="Filter GeoJSON features by python expression")
+@click.argument('filter')
+@use_rs_opt
+@click.pass_context
+def filter(ctx, filter, use_rs):
+    """
+    Filter GeoJSON objects read from stdin against the specified expression.
+
+    The expression is evaluated in a restricted namespace containing:
+        - sum
+        - min
+        - max
+        - math, imported module
+        - f, the feature in question
+
+    The expression will be evaluated for each feature and, if true,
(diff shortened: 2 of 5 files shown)


sgillies:
@perrygeo Awesome idea, definitely. What would you think about changing the query expressions to either 1) align with [jq filters](https://stedolan.github.io/jq/manual/#Basicfilters), so `'.properties.area > 1000.0' `instead of `"f['properties']['area'] > 1000.0"`, or 2) extend [snuggs expressions](https://github.com/mapbox/snuggs) in a Hy-like manner like `'(> (. (. f properties) area) 1000.0)'`?

perrygeo:
conceptually, I like the lisp approach. It's consistent with rio calc and and hy or extended snuggs gives us a built in parser.  We could directly leverage python functions, `math.sin`, `shape`, etc. But I worry the syntax is a bit esoteric and might scare off some users. Dot notation really doesn't translate to LISP well. 

With the jq expressions, they certainly look nice and clean. But could we extend them to support python functions? And how would we parse and evaluate them? At first glance I don't see any existing libraries for using that grammar in python.

sgillies:
Yeah, lisp isn't the right fit here. But I do think that dot accessors are much cleaner in this case than Python dicts and won't require as much attention to quoting and escaping. How about wrapping features in a class that provides `getattr()` access and surfacing that as `f`?

perrygeo:
@sgillies :+1:  to the `getattr()` idea, keeps things simple.

I'll give that a shot along with some tests here shortly.

perrygeo:
What do you think about adding a dependency to https://github.com/dsc/bunch ?

sgillies:
@perrygeo I think bunch looks very good. Maybe we should bring it all the way into the [core](https://github.com/Toblerity/Fiona/blob/master/fiona/ogrext.pyx#L217) for 1.7?

perrygeo:
Looks nice but tests don't pass on Python 3. bummer. And there are also performance concerns with copying a dict and creating a bunch of new objects - not a deal breaker for `filter` but it might be for including in the core.

geowurster:
@perrygeo Asked about my opinion on the use of Python expressions in #273 and since my answer applies to that issue and this in-progress PR, I'll answer here.  I recognize my comments below recommend giving the user mostly unsupervised access to `eval()` and almost the entire underlying Python environment, but I'm not trying to push these changes on this feature, so feel free to disagree and push back.  This is merely what I have learned in other projects.

I'm in favor of using Python expressions and wrote a [`sed`-like utility called `pyin`](https://github.com/geowurster/pyin), that is almost identical to this command, because I got sick of looking up `sed` syntax when I knew how to do it in Python but didn't want to write a script every time.  I went through several iterations in that project and came to some conclusions that I think apply here, although keep in mind that I don't think anyone else is using `pyin`.

1. Allowing multiple chained expressions that maintain state between each expression's execution is really powerful.
2. Don't spend too much time scoping `eval()`.  Python has all kinds of weird ways to access objects and its just not worth the time.  The benefits of allowing users to do whatever they want far outweigh the negatives.
3. Don't shy away from `eval()` because its "dangerous".  I have had far more mishaps with a poorly composed `rm something*/whatever*` than I have composing expressions for `pyin` and other tools that execute arbitrary code.  Any `setup.py` executes arbitrary code, so users can be trusted to safely compose Python expressions for filtering dictionaries.  Besides, they're in the terminal anyway where there are any number of bad things to do.
4. Don't try to be clever about exposing safe ways for the user to define an environment and import modules before evaluating expressions.  Just let them do whatever they want.  [`py`](https://github.com/Russell91/pythonpy/blob/master/pythonpy/__main__.py#L26-L34) has a good method for auto-importing objects.
5. Think about the scope of a command that uses `eval()` before thinking about all the bad stuff users could do.  Sure, the expression `__import__('os'); os.remove('*')` will do bad things, but why would I type that when I'm trying to filter features?  Anyone attempting to do something nefarious already has access to the computer at this point so taking it upon ourselves to block this syntax is pointless and unnecessarily constrains the user.
6. A lisp requires **all** users to learn a new syntax, but Fiona is advertised as a Python library, so a huge percentage of its users already know how they would filter the features in a script.  Expressions just let them do it without having to write the overhead.

As far as scoping the `eval()` environment goes, its worth [limiting direct access to some objects](https://github.com/geowurster/pyin/blob/master/pyin/core.py#L134-L142) to block the really obvious bad stuff and to help keep people from doing anything really strange.

Using `pyin` as an example, here's some weird stuff you can do - note that `json` is automatically imported:

```console
$ fio cat tests/data/coutwildrnp.shp \
    | pyin \
        "json.loads(line)" \
        "'Wilderness' in line['properties']['NAME']" \
        "int(line['properties']['STATE_FIPS']) < 9" \
        "{p: {k.lower(): v for k, v in val.items()} if p == 'properties' else val for p, val in line.items()}" \
        "json.dumps(line)" \
            | fio collect > OUT.geojson
```

Here's what's going on:

1. Use `fio cat` to print GeoJSON features to stdout and pipe them to pyin.
2. Convert each line to a dictionary.
3. Only process features where the substring 'Wilderness' appears in the 'NAME' property.
4. Only process features where the state FIPS code is < 9.
5. Convert all the field names to lowercase.
6. Serialize as a JSON string.
7. Use `fio collect` to dump to a GeoJSON file.

It's not pretty, but its easy to compose each step in the moment.  Making the property names lowercase took a minute to figure out though, so there's an argument for exposing properties in a separate variable in the `eval()` environment.  In [another project](https://github.com/SkyTruth/gpsdio/blob/master/gpsdio/core.py#L243-L301) that structures data as a dictionary, I use the data object as the [local scope](https://github.com/SkyTruth/gpsdio/blob/master/gpsdio/core.py#L288) for eval and then copy the entire dict inside an [additional variable](https://github.com/SkyTruth/gpsdio/blob/master/gpsdio/core.py#L289), but I have been considering removing the fields from the scope and just interacting with the data dictionary because its difficult to rebuild the feature after all the expressions have been evaluated.  This would be like doing `eval("'Wilderness' in properties['NAME']", global_scope, feature)`.  `Munch` is a good middle-ground.

perrygeo:
@geowurster  My intention with limiting the scope in eval was primarily to make it clear that no advanced data manipulations or anything else should be attempted in the `rio filter` expression - only an expression that evaluates to True/False with no side effects. If anyone would be tempted to, say, rename fields or manipulate the data in any way, I'd point them to `pyin` or the hypothetical future `rio calc` #273.

Security is a secondary concern but, as you point out, there are convoluted ways of recovering builtins and such that eval is never going to be safe. Oh well, you're at the command line. If you're calling this with arbitrary user input, you're doing it wrong! We'll just have to make that clear.

Issue #278, opened 2015-09-29
Check for FileGDB driver support from OGR


micahcochran:
Fiona needs to check with OGR if the FileGDB driver is installed. It reports that it supports the FileGDB driver.

```
$ fio env | grep FileGDB
FileGDB (modes 'r', 'a', 'w')
OpenFileGDB (modes 'r')
```

```
$ ogrinfo --formats | grep FileGDB
  -> "OpenFileGDB" (readonly)
```

This could also be present a problem for OpenFileGDB, if someone is using a version of GDAL of less than 1.11.

Otherwise trying to write to a FileGDB when the FileGDB driver software isn't installed causes errors like this.

```
$ fio dump global_cities_all.geojson | fio -v load gen/global.gdb --driver "FileGDB"
ERROR:fio:Exception caught during processing
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/fiona/fio/cat.py", line 569, in load
    schema=schema) as dst:
  File "/usr/local/lib/python3.4/dist-packages/fiona/__init__.py", line 169, in open
    enabled_drivers=enabled_drivers)
  File "/usr/local/lib/python3.4/dist-packages/fiona/collection.py", line 150, in __init__
    self.session.start(self, **kwargs)
  File "fiona/ogrext.pyx", line 734, in fiona.ogrext.WritingSession.start (fiona/ogrext.c:12921)
ValueError: Null driver
Aborted!
Exception ignored in: <bound method Collection.__del__ of <open Collection 'gen/global.gdb:global', mode 'w' at 0xb718194c>>
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/fiona/collection.py", line 407, in __del__
    self.__exit__(None, None, None)
  File "/usr/local/lib/python3.4/dist-packages/fiona/collection.py", line 402, in __exit__
    self.close()
  File "/usr/local/lib/python3.4/dist-packages/fiona/collection.py", line 386, in close
    self.flush()
  File "/usr/local/lib/python3.4/dist-packages/fiona/collection.py", line 375, in flush
    if self.session is not None and self.session.get_length() > 0:
  File "fiona/ogrext.pyx", line 455, in fiona.ogrext.Session.get_length (fiona/ogrext.c:8748)
ValueError: Null layer
```

I assume that the driver not being supported is what this error is telling me.   If any data is needed to replicate these I'm happy to hand over any of these files for testing.

[...]

PR #280, opened 2015-10-06
Filters supported_drivers to only those in OGR

micahcochran:
This is a fix to address Issue #278 that I reported.

It does an in-place scan of `supported_drivers`, removing any drivers that are not supported by OGR.

Most of the line addition & deletion is moving the `supported_drivers` to its own file.  I moved it so I could use Python to do this instead of recoding the algorithm in Cython.

I was a little unsure if the call to ` _filter_supported_drivers()` should be run in that module or should be called from `__init__.py`


sgillies:
Thanks! Sorry to be so late.

micahcochran:
Thanks Sean.  Not a problem.

Issue #285, opened 2015-10-26
Add support for deleting datasets


snorfalorpagus:
Fiona does not currently support deleting a dataset. I think this would be a useful feature, as when working with ESRI shapefiles there are a number of different files that need to be removed, which is a pain with `os.unlink`.

OGR/GDAL provides a method for this, `OGR_Dr_DeleteDataSource` (deprecated in 2.0) and `GDALDeleteDataset`.

http://www.gdal.org/ogr__api_8h.html#a4d2314e22acd846e7e31ff5a5c5ec27f
http://www.gdal.org/gdal_8h.html#a97939d588e2b350b9ec7a440f3e06a7c

If this feature was implemented, I'm not sure what the API should be. Perhaps `fiona.delete` as a counterpart to `fiona.open`?

rbuffat:
Finding the right files to delete with glob should be fairly easy.

jwass:
Maybe `fiona.remove` as an analog to [`os.remove`](https://docs.python.org/2/library/os.html#os.remove) - instead of `delete`? I've always preferred remove to unlink but they are identical.

sgillies:
@snorfalorpagus sure, why not? I think @jwass's analogy to `os.remove` is right on. There's already a `fiona.listlayers()` function that acts like `os.listdir`, so a function that removed a layer (think multi-layer datasets) would fit right in.

snorfalorpagus:
Closed by #288

MathiasBaumgartinger:
Hello - stumbled across this ancient issue when i was trying to remove a subdataset from a GPKG using fiona. Is there any way to remove a feature subdataset from a geopackage using fiona? If no, are there any ambitions for such an implementation? 

I wouldve suggested the syntax `f"GPKG:{geopackage}:{feature_layer_name}"`

sgillies:
@MathiasBaumgartinger that's a good idea. I think we'll do it for 2.0.

PR #288, opened 2015-10-27
Initial attempt at fiona.remove for deleting data sources

snorfalorpagus:
As suggested in #285.

This was slightly more tricky than I initially realised, as OGR needs to know which driver to use in order to delete the data source.

I've only implemented this for GDAL 1.x for now, and still need to write the function documentation. Posting here for comment.

### DIFF
--- a/fiona/__init__.py
+++ b/fiona/__init__.py
@@ -183,13 +183,30 @@ def open(
 collection = open
 
 
-def remove(path, driver=None):
-    # collection was passed instead of path
-    if isinstance(path, Collection):
-        collection = path
+def remove(path_or_collection, driver=None):
+    """Deletes an OGR data source
+
+    The required ``path`` argument may be an absolute or relative file path.
+    Alternatively, a Collection can be passed instead in which case the path
+    and driver are automatically determined. Otherwise the ``driver`` argument
+    must be specified.
+
+    Raises a ``RuntimeError`` if the data source cannot be deleted.
+
+    Example usage:
+
+      fiona.remove('test.shp', 'ESRI Shapefile')
+
+    """
+    if isinstance(path_or_collection, Collection):
+        collection = path_or_collection
         path = collection.path
         driver = collection.driver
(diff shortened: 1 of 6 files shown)


sgillies:
@snorfalorpagus looks like a good start. We'll need tests.

I'm going to do something about refactoring the .pyx files so we're not repeating ourselves. It might be possible to write `_remove()` just once for both versions of GDAL.

snorfalorpagus:
@sgillies I actually wrote some tests, but forgot to stage the file.

When `driver` isn't specified by the user, should we make an attempt at guessing based on file extension, rather than just assuming `ESRI Shapefile`? We can't just scroll through all the possible drivers like we do for `open` as the operation is destructive.

sgillies:
@snorfalorpagus no guessing, please. I could see that becoming a large amount of fragile code. We should make the driver argument required if the input isn't a collection.

snorfalorpagus:
I think this is working now. If it looks OK, I'll squash the commits together.

snorfalorpagus:
I forgot, this isn't quite finished as the call I'm using for delete is deprecated - need to explore when the new call became available.

rbuffat:
@snorfalorpagus 
OGR_Dr_DeleteDataSource is deprecated for gdal2x. Thus in ogrext1.pyx (for gdal1x in the current implementation) OGR_Dr_DeleteDataSource is the one to use, in ogrext2.pyx (gdal2x) you can use GDALDeleteDataset.

Maybe a check for ODrCDeleteDataSource would be nice?

snorfalorpagus:
@rbuffat Sounds good to me.

Note to self: `OGR_Dr_TestCapability`

snorfalorpagus:
I've added a check for the driver capability, but not sure how to test it. I tested locally with the GMT driver, but that doesn't seem to be available on Travis (and may not be on other users machines).

Also switched to `GDALDeleteDataset` in GDAL 2.x.

PR #291, opened 2015-11-06
Add fio -D option and ability to set json_lib

### LINKED ISSUE #271: Benchmarking json vs msgpack vs geobuf/protobuf
Relevant to some discussion we're having about processing pipelines. Results here: https://gist.github.com/sgillies/bfeb6b479bfd2e6a0a38.

The standard lib json is significantly faster than umsgpack or geobuf.

Nothing compares to marshal.

geowurster:
**Not ready**

Closes #271 

@sgillies I was thinking something like this.  As an MVP I only changed `fio cat`.  The assumption is that the user will supply a library that implements `json_lib.loads()` and `json_lib.dumps()`.   `ujson` doesn't like the `sort_keys` parameter so I moved it to an option so it can be disabled.  `simplejson` supports all of the optional `json.dumps()` options.

```console
$ fio -D json_lib=ujson cat tests/data/coutwildrnp.shp --no-sort-keys
```

### DIFF
--- a/fiona/fio/cat.py
+++ b/fiona/fio/cat.py
@@ -80,7 +80,7 @@ def id_record(rec):
               help="log errors but do not stop serialization.")
 @options.dst_crs_opt
 @use_rs_opt
-@click.option('--sort-keys/--no-sort-keys', is_flag=True, default=True,
+@click.option('--sort-keys/--no-sort-keys', is_flag=True, default=False,
               show_default=True, help="Sort JSON keys when serializing.")
 @click.option('--bbox', default=None, metavar="w,s,e,n",
               help="filter for features intersecting a bounding box")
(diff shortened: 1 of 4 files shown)


sgillies:
@geowurster what would you think about having sort_keys=False be the default so that ujson users don't have to disable? Would be in line with https://docs.python.org/2/library/json.html#json.dumps.

geowurster:
@sgillies Yeah thats fine with me.  The only reason I didn't have it default to `False` was to maintain the current behavior, but its JSON so we can probably assume our users are looking for keys by name rather than index.  Switched in latest commit.

geowurster:
@sgillies I played with this a bit more, implemented it on a few additional commands, and it just feels kind of awkward to use, so I vote we close this one.  Definitely better suited to a config file with an optional override.

sgillies:
:+1:

PR #305, opened 2015-12-20
Fix inconsistencies in fiona.Collection() iteration

### LINKED ISSUE #301: Inconsistent Collection() iteration
Iteration in `fiona.Collection()` is not "pinned" until the `__iter__()` method is called.  Opening a file and calling `next()` on the file produces the first feature, but entering a for loop inside the same context manager produces the first feature again on the first iteration.  Doing the reverse only produces each feature once.

/ cc @cjthomas730

**Examples**

Open a file, call `next()` a couple times, and then loop over the same collection:

``` python
print("Individual")
with fio.open('tests/data/coutwildrnp.shp') as src:
    print(next(src)['id'])
    print(next(src)['id'])
    print("Loop")
    for idx, feat in enumerate(src):
        print(feat['id'])
        if idx >= 2:
            break
Individual
0
1
Loop
0
1
2
```

Loop over the collection a few times and then call `next()` outside the loop:

``` python
print("Loop + individual")
with fio.open('tests/data/coutwildrnp.shp') as src:
    for idx, feat in enumerate(src):
        print(feat['id'])
        print(next(src)['id'])
        if idx > 3:
            break
    print(next(src)['id'])
    print(next(src)['id'])
Loop + individual
0
1
2
3
4
5
6
7
8
9
10
11
```

Python iterator:

``` python
print("Python iterator")
r = iter(range(5))
print(next(r))
print(next(r))
for i in r:
    print(i)
Python iterator
0
1
2
3
4
5
```

geowurster:
**Not ready**

Closes #301 

- [ ] Add test for `keys()` method
- [ ] Add test for `items()` method
- [ ] Add test for `filter()` method

/ cc @cjthomas730

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,6 @@ fiona/_drivers.c
 fiona/_err.c
 fiona/_geometry.c
 fiona/_transform.cpp
+VERSION.txt
+fiona/ograpi.pxd
+fiona/ogrext.pyx
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,6 @@ fiona/_transform.cpp
 VERSION.txt
 fiona/ograpi.pxd
 fiona/ogrext.pyx
+tests/data/coutwildrnp.json
+tests/data/coutwildrnp.tar
+tests/data/coutwildrnp.zip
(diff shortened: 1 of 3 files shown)


geowurster:
Need to work out some features in https://github.com/Toblerity/Fiona/issues/301, so I'm closing this.

PR #310, opened 2015-12-21
Ignore unicode and value errors

### LINKED ISSUE #293: fio cat --ignore-errors is ignored
The `--ignore-errors` flag for `fio cat` is not used.

sgillies:
Ignore the branch name :)

Closes #293


sgillies:
@geowurster ready for early review. I need to come up with a test or two. I wouldn't expect a single source file to spit out a broken feature here and there, but maybe an entire file in the input sequence could be rotten.

sgillies:
Nope. I'm going to close this PR and mark #293 as a won't fix.

PR #312, opened 2015-12-23
add a --no-parse option to fio collect

### LINKED ISSUE #306: a faster, dumber option for fio collect
`fio collect --record-buffered` is a good way handle streaming large datasets. But it still loads and dumps each feature which add significant overhead if the features are large.

Is there a place in Fiona for a more trusting variant that just echos each string in the sequence and assumes it's a feature? Maybe a `fio collect --trust`?

Proof of concept:
https://gist.github.com/perrygeo/fe756f9db0087a70ffb7

perrygeo:
This is working but needs some review. 

It should work with `--record-buffered` and `--no-record-buffered`, and with line-delimited and rs-delimited sequences. 

if we're not parsing the json, we can't transform it so the `--src-crs` option isn't compatible with `--no-parse` and raises a click usage error.

I'm not thrilled at adding lots of if statements everywhere; there might be an entirely different way to structure this logic that doesn't result in so many nested conditionals.

resolves #306

### DIFF
--- a/fiona/fio/cat.py
+++ b/fiona/fio/cat.py
@@ -167,44 +167,55 @@ def collect(ctx, precision, indent, compact, record_buffered, ignore_errors,
 
     first_line = next(stdin)
 
-    # If input is RS-delimited JSON sequence.
-    if first_line.startswith(u'\x1e'):
-        def feature_gen():
-            buffer = first_line.strip(u'\x1e')
-            for line in stdin:
-                if line.startswith(u'\x1e'):
-                    if buffer:
-                        if parse:
+    # If parsing geojson
+    if parse:
+        # If input is RS-delimited JSON sequence.
+        if first_line.startswith(u'\x1e'):
+            def feature_gen():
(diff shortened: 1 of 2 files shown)


sgillies:
Yeah, at the very least get the `if`s out of the `for` loops. It strikes me that the only thing done to features coming from the generator is serialization and writing... maybe there's a function that loads and transforms and dumps and yields GeoJSON text when `--parse` vs one that just yields the GeoJSON when `--no-parse`?

perrygeo:
So I took a stab at this but am not crazy about round 2 - basically I am stuck repeating all the `rs` delimiting logic in the parse vs no-parse variants of the function. Still, it improves the readability a bit so I'm pushing it.

sgillies:
@perrygeo this ready? I like it.

perrygeo:
Functionality wise, ready to go.

There is a cleaner way to organize the flow. Having 4 definitions of feature_text_gen is a bit strange. But we can refactor later.

PR #322, opened 2016-02-07
Don't copy gdalapi & data files in clean target.

sebastic:
The clean target should modify these files, it prevents builds of the Debian package due to unexpected changes to the source.


sgillies:
Thanks!

PR #336, opened 2016-04-14
Add --layer to $ fio info and $ fio load

### LINKED ISSUE #299: Problems Writing to FileGDB format with fio-load
I'm trying to use `fio` to convert a shapefile into a FileGDB.  FileGDB is a multilayer format.  These examples are really only treatingthe format as a single layer format.  That is so that I can try to figure out what is going wrong.

This example seems to work
```python
import fiona

with fiona.drivers():
    with fiona.open('tests/data/coutwildrnp.shp') as source:
        meta = source.meta
        meta['driver'] = 'FileGDB'
        meta['layer'] = 'layername'

        with fiona.open('out.gdb', 'w', **meta) as sink:
            for f in source:
                sink.write(f)
```

GDAL/OGR does it this way:
```
C:\...>ogr2ogr try1.gdb coutwildrnp.shp -f FileGDB
```
Slap `-nln layername` on the end to rename the layer to 'layername'.

So I tried this:
```
C:\...>fio dump coutwildrnp.shp | fio load try2.gdb --driver FileGDB
WARNING:Fiona:CPLE_AppDefined in Empty Spatial Reference
```
That crashes python on windows.  :-(

My first guess was this.  I looked in the source, but I didn't find a "layer" option.  
```
C:\...>fio dump coutwildrnp.shp | fio load try3.gdb --driver FileGDB --layer layername
Error: no such option: --layer
```
This doesn't cause a crash, but I'm guessing this never gets to GDAL/OGR due to the bad command line argument.

This was the other way I thought this could work.  (ArcGIS Desktop treats it GeoDatabases like directories in ArcCatalog.)
```
C:\...>fio dump coutwildrnp.shp | fio load try3.gdb/layername --driver FileGDB
ERROR:Fiona:CPLE_AppDefined in FGDB data source name must use 'gdb' extension.
```
Yet again, it crashes python on windows.  :-(   The same thing happens if I omit the `.gdb` extension.

My problem is I can't get `fio` to write to FileGDB format.    Is it just me?  Is it my installation? Is this something needing to be patched?

### LINKED ISSUE #316: Add --layer parameter to fio info
From https://github.com/Toblerity/Fiona/issues/299#issuecomment-170678550

We can use a click callback to check if the parameter is a layer index so we can use names or indexes.

/ cc @smnorris

geowurster:
Closes #299 
Closes #316 

@sgillies Ready for review.

This PR adds `--layer` to `$ fio info` to let users specify which layer they want to run a report on, and `--layer` to `$ fio load` so users can specify which layer they want to load data into.

**info example**
```console
$ fio info --layer routes tests/data/test_gpx.gpx --name
/Users/wursterk/code/Fiona/fiona/fio/info.py:83: RuntimeWarning: Layer does not support counting Setting 'count' to 'null'
  warnings.warn(msg, RuntimeWarning)
routes
```

**load example**

```console
$ fio cat tests/data/coutwildrnp.shp \
    | fio load OUT.gdb -f FileGDB --src-crs EPSG:4326 --sequence --layer new
$ fio ls OUT.gdb
["new"]
```

This also addresses an issue where `$ fio info` crashed when `len(collection)` crashed so now we catch the exception, issue a `RuntimeWarning`, and set `count=None`.

`$ ogrinfo`'s report sets the feature count to `0`, but `None/null` seems more appropriate and more likely to alert users that they may need to convert their data to a different format.

Previously:

```console
fio info tests/data/test_gpx.gpx 
  File "/Users/wursterk/code/Fiona/fiona/fio/info.py", line 61, in info
    info.update(bounds=src.bounds, count=len(src))
  File "/Users/wursterk/code/Fiona/fiona/collection.py", line 375, in __len__
    raise TypeError("Layer does not support counting")
TypeError: Layer does not support counting
Aborted!
```

/ cc @micahcochran @smnorris

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ script:
   - coverage run --source=fiona --omit='*.pxd,*.pyx,*/tests/*,*/docs/*,*/examples/*,*/benchmarks/*' -m nose --exclude test_filter_vsi --exclude test_geopackage
 
 after_success:
-  - coveralls
+  - coveralls || echo "!! intermittent coveralls failure"
 
 env:
 
--- a/fiona/fio/info.py
+++ b/fiona/fio/info.py
@@ -14,6 +14,7 @@ from cligj import indent_opt
 
 import fiona
 import fiona.crs
+from fiona.fio import options
 
 
 @click.command(short_help="Print information about the fio environment.")
@@ -36,9 +37,13 @@ def env(ctx, key):
 
 
 # Info command.
-@click.command(short_help="Print information about a dataset.")
+@click.command()
 # One or more files.
 @click.argument('input', type=click.Path(exists=True))
+@click.option('--layer', metavar="INDEX|NAME", callback=options.cb_layer,
+              help="Print information about a specific layer.  The first layer "
+                   "is used by default.  Layers use zero-based numbering when "
+                   "accessed by index.")
 @indent_opt
 # Options to pick out a single metadata item and print it as
 # a string.
@@ -54,19 +59,27 @@ def env(ctx, key):
 @click.option('--name', 'meta_member', flag_value='name',
               help="Print the datasource's name.")
 @click.pass_context
-def info(ctx, input, indent, meta_member):
+def info(ctx, input, indent, meta_member, layer):
+
+    """
+    Print information about a dataset.
+
+    When working with a multi-layer dataset the first layer is used by default.
+    Use the '--layer' option to select a different layer.
+    """
+
     verbosity = (ctx.obj and ctx.obj['verbosity']) or 2
     logger = logging.getLogger('fio')
     try:
         with fiona.drivers(CPL_DEBUG=verbosity>2):
-            with fiona.open(input) as src:
+            with fiona.open(input, layer=layer) as src:
                 info = src.meta
                 info.update(bounds=src.bounds, name=src.name)
                 try:
                     info.update(count=len(src))
                 except TypeError as e:
                     info.update(count=None)
-                    msg = str(e) + "  Setting 'count' to 'null'"
+                    msg = "{exc} Setting 'count' to 'null'".format(exc=str(e))
                     warnings.warn(msg, RuntimeWarning)
                 proj4 = fiona.crs.to_string(src.crs)
                 if proj4.startswith('+init=epsg'):
--- a/fiona/fio/info.py
+++ b/fiona/fio/info.py
@@ -7,6 +7,7 @@ import code
 import logging
 import json
 import sys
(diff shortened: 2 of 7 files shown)


fiona/fio/info.py
```diff
@@ -50,15 +56,31 @@ def env(ctx, key):
...
+
     verbosity = (ctx.obj and ctx.obj['verbosity']) or 2
     logger = logging.getLogger('fio')
     try:
         with fiona.drivers(CPL_DEBUG=verbosity>2):
-            with fiona.open(input) as src:
+            with fiona.open(input, layer=layer) as src:
                 info = src.meta
-                info.update(bounds=src.bounds, count=len(src))
+                info.update(bounds=src.bounds, name=src.name)
```
geowurster: @sgillies Forgot to mention that this PR also adds `Collection.name` to the info report to make it less ambiguous.

micahcochran:
It looks like exactly what I was asking for. Thanks @geowurster .  

I haven't had a chance to try it out.

sgillies:
@geowurster I'm struggling on cleaning up Rasterio's GDAL/AWS environment now, will :eyes: this closely tomorrow.

geowurster:
@sgillies No rush.  Didn't mean to distract.

.travis.yml
```diff
@@ -32,7 +32,7 @@ script:
   - coverage run --source=fiona --omit='*.pxd,*.pyx,*/tests/*,*/docs/*,*/examples/*,*/benchmarks/*' -m nose --exclude test_filter_vsi --exclude test_geopackage
 
 after_success:
-  - coveralls
+  - coveralls || echo "!! intermittent coveralls failure"
```
sgillies: Yes!

fiona/fio/info.py
```diff
@@ -50,15 +56,31 @@ def env(ctx, key):
...
+                try:
+                    info.update(count=len(src))
+                except TypeError as e:
+                    info.update(count=None)
+                    msg = "{exc} Setting 'count' to 'null'".format(exc=str(e))
+                    warnings.warn(msg, RuntimeWarning)
```
sgillies: @geowurster I think OGR considers a count of `-1` (or `None` in our case) "normal" for formats like GPX (see https://github.com/Toblerity/Fiona/pull/190), so how about `log.debug()` instead of a warning?

smnorris:
Thanks @geowurster this is great.

Any suggestions on a way to use `fio cat` with multi-layer sources? It looks like a `--layer` option can't really be added to that one. 

Sorry if this is the wrong venue for user questions.

PR #338, opened 2016-04-18
Warn users if they use gdal2 with fiona 1.x

rbuffat:
This PR includes two updates:

- The user is presented a warning if he uses fiona 1.x with gdal2 when he imports fiona
- The FIELD_TYPES list is extended with the 64bit Integer Types of gdal2.0.  64bit integers are thus internally treated as normal integers. 

This should have no effect when gdal1.x is used. For some drivers (e.g. shapefile), integer fields with a large enough length are treated as 64bit integer in gdal2.x while as "normal" 32bit integers in gdal1.x. Adding the 64bit Integer type to the FIELD_TYPES should thus fix the 'IndexError: list index out of range' present in some bug reports of the past.

### DIFF
--- a/fiona/__init__.py
+++ b/fiona/__init__.py
@@ -91,8 +91,8 @@ log.addHandler(NullHandler())
 
 # Warn user that they use fiona 1.x with gdal 2.0
 if get_gdal_version_num() >= calc_gdal_version_num(2, 0, 0):
-    warnings.warn("Fiona {} is not compatible with installed GDAL {}".format(__version__,
-                                                                             get_gdal_release_name()))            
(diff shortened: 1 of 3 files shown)


snorfalorpagus:
Should this raise a warning or an exception? A common pattern for optional dependencies is two wrap the import in a try-except block. This wouldn't work if the module raises a warning only.

```
try:
    import something
except:
    # do or import something else
```

rbuffat:
@snorfalorpagus My thought was that some functionality might still work, thus to just display a warning that if something goes bad the user has an indication for the source of the problem.

Raising an exception may be the better approach.

sgillies:
I'll take care of the exception in #341.

PR #342, opened 2016-05-05
Gdal version attr

sgillies:
Adds a `__gdal_version__` attribute and raises `ImportError` if GDAL>=2.0 (until the 1.7 release).


sgillies:
Ack. wrong target branch.

PR #348, opened 2016-05-21
fio calc

### LINKED ISSUE #273: fio command for calculating new properties
`fio calc` could take a stream of GeoJSON features on stdin calculate additional properties.

For each feature, it could evaluate one or more expressions,  place the derived variable(s) in the `properties` dict then print the feature to stdout.

So for this input:
```
{'type': 'Feature',
 'id': 1,
 'properties': {'a': 1, 'b': 2}}
```
You might have a CLI something like this
```
fio calc --new "c" "f['properties']['a'] + f['properties']['b'] * 2"
```
which would yield
```
{'type': 'Feature',
 'id': 1,
 'properties': {'a': 1, 'b': 2, 'c': 5}}
```

There are obviously some details to work out
* what is in the namespace when the expression is evaluated
* python expressions or some other DSL? 
* what's the best interface for args and options

If there's interest in the general idea, I can get started on a PR

perrygeo:
Property calculator for `fio`, resolves #273 

examples:

Calculating an area as `new_area` property using shapely
```
$ fio cat tests/data/coutwildrnp.shp | fio calc new_area "shape(f.geometry).area" | jq .properties
...
{
  "STATE_FIPS": "08",
  "AGBUR": "NPS",
  "AREA": 0.000817711,
  "FEATURE1": "Wilderness",
  "URL": "http://www.wilderness.net/index.cfm?fuse=NWPS&sec=wildView&wname=Mesa%20Verde",
  "WILDRNP020": 511,
  "new_area": 0.0008177111012628302,
  "PERIMETER": 0.120627,
  "NAME": "Mesa Verde Wilderness",
  "STATE": "CO",
  "FEATURE2": null
}
```

Numerical calcs - the area-perimeter ratio as `ap_ratio` property
```
$ fio cat tests/data/coutwildrnp.shp | fio calc ap_ratio "f.properties.AREA / f.properties.PERIMETER" | jq .properties
...
{
  "URL": "http://www.wilderness.net/index.cfm?fuse=NWPS&sec=wildView&wname=Mesa%20Verde",
  "NAME": "Mesa Verde Wilderness",
  "ap_ratio": 0.0067788389000804136,
  "WILDRNP020": 511,
  "STATE_FIPS": "08",
  "AGBUR": "NPS",
  "AREA": 0.000817711,
  "PERIMETER": 0.120627,
  "STATE": "CO",
  "FEATURE2": null,
  "FEATURE1": "Wilderness"
}
```

Booleans
```
$ fio cat tests/data/coutwildrnp.shp | fio calc is_big "f.properties.AREA > 0.01" | jq .properties
...
{
  "URL": "http://www.wilderness.net/index.cfm?fuse=NWPS&sec=wildView&wname=Mesa%20Verde",
  "NAME": "Mesa Verde Wilderness",
  "is_big": false,
  "WILDRNP020": 511,
  "STATE_FIPS": "08",
  "AGBUR": "NPS",
  "AREA": 0.000817711,
  "PERIMETER": 0.120627,
  "STATE": "CO",
  "FEATURE2": null,
  "FEATURE1": "Wilderness"
}
```

Strings
```
$ fio cat tests/data/coutwildrnp.shp | fio calc shouting_name "f.properties.NAME.upper()" | jq .properties
...
{
  "URL": "http://www.wilderness.net/index.cfm?fuse=NWPS&sec=wildView&wname=Mesa%20Verde",
  "NAME": "Mesa Verde Wilderness",
  "shouting_name": "MESA VERDE WILDERNESS",
  "WILDRNP020": 511,
  "STATE_FIPS": "08",
  "AGBUR": "NPS",
  "AREA": 0.000817711,
  "PERIMETER": 0.120627,
  "STATE": "CO",
  "FEATURE2": null,
  "FEATURE1": "Wilderness"
}
```

This overlaps `fio filter` quite a bit so we might want to think about abstracting some of those bits out.

@jacquestardie @dnomadb @sgillies @geowurster - what do you think?


sgillies:
@perrygeo I like it. Assigning to @geowurster.

How about putting shapely in a setuptools extras_require group named "calc" so that it gets installed if you do `pip install fiona[calc]`?

geowurster:
Cool - I'll check it out.

tests/test_fio_calc.py
```diff
@@ -0,0 +1,70 @@
...
+from fiona.fio.calc import calc
+from .fixtures import feature_seq
+
+
+def test_fail():
+    runner = CliRunner()
+    result = runner.invoke(calc,
+                           ["TEST", "f.properties.test > 5"],
+                           "{'type': 'no_properties'}")
+    assert result.exit_code == 1
```
geowurster: @perrygeo While this expression does fail because the input feature lacks a `properties` key, it is currently failing because the input is invalid JSON.  The quote type on the input JSON needs to be switched to: `'{"type": "no_properties"}'`.

Adding an `assert False, result.output` right below this line shows the failure:

```console
E       AssertionError: Exception caught during processing
E         Traceback (most recent call last):
E           File "/Users/wursterk/code/perrygeo-fiona/fiona/fio/calc.py", line 62, in calc
E             for i, obj in enumerate(source):
E           File "/Users/wursterk/code/perrygeo-fiona/fiona/fio/helpers.py", line 30, in gen
E             yield json.loads(first_line)
E           File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/json/__init__.py", line 319, in loads
E             return _default_decoder.decode(s)
E           File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/json/decoder.py", line 339, in decode
E             obj, end = self.raw_decode(s, idx=_w(s, 0).end())
E           File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/json/decoder.py", line 355, in raw_decode
E             obj, end = self.scan_once(s, idx)
E         json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
E         Aborted!
```

fiona/fio/calc.py
```diff
@@ -0,0 +1,79 @@
+import json
```
geowurster: @perrygeo Need a `from __future__ import division` so we get float division on Python 2.

```console
$ echo '{"properties": {"value": 3}}' | fio calc NEW "f.properties.value / 5"
{"properties": {"NEW": 0, "value": 3}}
```

fiona/fio/calc.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    Create a new property on GeoJSON features using the specified expression.
+
+    \b
+    The expression is evaluated in a restricted namespace containing:
+        - sum
+        - min
+        - max
+        - math (imported module)
+        - shape (optional, imported from shape.geometry if available)
```
geowurster: @perrygeo Typo: `shape.geometry` should be `shapely.geometry`.

fiona/fio/calc.py
```diff
@@ -0,0 +1,79 @@
...
+import click
+from cligj import use_rs_opt
+
+from .helpers import obj_gen
+
+
+@click.command(short_help="Calculate GeoJSON property by Python expression")
+@click.argument('property_name')
+@click.argument('expression')
+@click.option('--overwrite', '-o', is_flag=True, default=False,
```
geowurster: @perrygeo I wonder if we shouldn't reserve `-o` for future use?  It's a pretty standard flag for specifying output, and while Fiona currently almost exclusively writes to `stdout`, it will likely create problems if we switch to writing directly to disk.  The CLI is pretty consistent as-is, so I don't feel too strongly about this.

fiona/fio/calc.py
```diff
@@ -0,0 +1,79 @@
...
+    \b
+    $ fio cat data.shp | fio calc sumAB  "f.properties.A + f.properties.B"
+    """
+    logger = logging.getLogger('fio')
+    stdin = click.get_text_stream('stdin')
+
+    def calc_func(feature):
+        safe_dict = {'f': munchify(feature)}
+        safe_dict['sum'] = sum
+        safe_dict['pow'] = sum
```
geowurster: @perrygeo Typo: both `sum` and `pow` are mapped to `sum()`.

tests/test_fio_calc.py
```diff
@@ -0,0 +1,70 @@
+import json
```
geowurster: @perrygeo Need a `from __future__ import division` here too.

fiona/fio/calc.py
```diff
@@ -0,0 +1,79 @@
...
+        return eval(expression, {"__builtins__": None}, safe_dict)
+
+    try:
+        source = obj_gen(stdin)
+        for i, obj in enumerate(source):
+            features = obj.get('features') or [obj]
+            for j, feat in enumerate(features):
+
+                if not overwrite and property_name in feat['properties']:
+                    raise ValueError(
```
geowurster: @perrygeo A `click.ClickException()` may produce a better error for the user.

geowurster:
@perrygeo This looks like a good MVP.  To help with type conversions and other calculations I think adding these builtins to the calc scope would be useful:

* `bool`
* `float`
* `int`
* `str`
* `len`

These functions could be given a totally valid `None`, so in another project I have been doing something similar to:

```python
import functools

def nullable(val, cast):
    if val is None:
        return None
    else:
        return cast(val)

safe_dict = {
    'str': functools.partial(nullable, str),
    'int': functools.partial(nullable, int),
    'float': functools.partial(nullable, float)}
```

perrygeo:
👍  to all those suggestions - thank you @geowurster ! 

All merged up with the `cli-files` branch (now in master)

The only significant modification was to pull the expression evaluation out into a helper function that is shared by both `fio calc` and `fio filter`.

Issue #349, opened 2016-05-25
Add --layer option to fio-cat


sgillies:
To allow access to layers within a file GDB, etc.

smnorris:
Thanks. This seemed trivial at first so I was happy to take it on before I made the comment in the other thread. I started on the `cli-files` branch, just copying the patterns from `info` and `load`... but of course `cat` is concatenate!

This clearly doesn't work:  `fio cat input.gdb --layer inlayer1 input.gdb --layer inlayer2` 

So maybe a comma separated list is the way to go? 
`fio cat input.gdb input.gdb --layers "inlayer1,inlayer2"`
But then handling a mix of multi and single layer input files needs to be considered.

sgillies:
@smnorris as you can see, multi-layer files haven't been our our radar. Let's take of care them now!

The usage snippet above is a lot like that of [`rio-stack`](https://github.com/mapbox/rasterio/blob/master/docs/cli.rst#stack). What would you think about adapting that? Simplest usage would be 

`fio cat input1.gdb --layer "1.a"`

Two layers from a single input:

`fio cat input1.gdb --layer "1.a,1.b"`

Five layers from two inputs:

`fio cat input1.gdb --layer "1.a,1.b" input2.gdb --layer "2.a,2.c,2.z"`

smnorris:
👍  
I didn't realize it was possible to specify options in between the input layer args, I've only scratched the surface of `click`.

geowurster:
@smnorris A click callback will be useful here too.  Something like:

```python
from collections import defaultdict

import click

def _cb_layer(ctx, param, value):

    out = defaultdict(list)
    for raw in value:
        for v in raw.split(','):
            ds, name = v.split('.')
            out[ds].append(name)

    return out


@click.option('--layer', callback=_cb_layer)
```

to process the CLI values like:

```python
cli_values = ["1.a,1.b", "2.a,2.c,2.z"]
print(_cb_layer(None, None, cli_values))
```

so in the function you get:

```python
{
    '1': ['a', 'b'],
    '2': ['a', 'c', 'z']
}
```

smnorris:
It seems tricky to get consistency between the various layer and path indexing systems:
- `rio stack --bidx`:  1 based indexing
- `fio info --layer`: 0 based indexing
- proposed `fio cat --layer` syntax: 1 based index for the paths/files, and I think layer index support would be nice so `fio cat --layer` is consistent with `fio info --layer`

But then we get this:

[...]

PR #354, opened 2016-06-07
Multilayer

smnorris:
For #349, adds `--layer` to `cat` and `dump`.  `fio cat` layer indexing is by name only to keep things simple.

The layer test function in `test_fio_cat.py` isn't very comprehensive but I didn't want to add a new test data source and couldn't seem to get any features out of `test_gpx.gpx`.  I did a better feature count test against a temp .gdb created with:
```
$ fio cat coutwildrnp.shp \
| fio filter "f.properties.NAME[:5] == 'Mount'" \
| fio collect \
| fio load multilayer.gdb --driver FileGDB --layer mountains

$ fio cat coutwildrnp.shp \
| fio filter "f.properties.NAME[:5] != 'Mount'" \
| fio collect \
| fio load multilayer.gdb --driver FileGDB --layer others

$ fio cat multilayer.gdb --layer "1.mountains,1.others" | wc -l 
67
```

### DIFF
--- a/fiona/fio/cat.py
+++ b/fiona/fio/cat.py
@@ -19,6 +19,11 @@ warnings.simplefilter('default')
 # Cat command
 @click.command(short_help="Concatenate and print the features of datasets")
 @cligj.files_in_arg
+@click.option('--layer', default=None, multiple=True,
+              callback=options.cb_multi_layer,
+              help="Comma separated string of layer name(s) or index(es). The "
+                   "first layer is used by default. Layers use zero-based "
+                   "numbering when accessed by index.")
 @cligj.precision_opt
 @cligj.indent_opt
 @cligj.compact_opt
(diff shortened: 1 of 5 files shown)


sgillies:
@smnorris I tested this and it's looking good. It can be used on a directory of files as well as a geodatabase.

```bash
$ fio cat --layer 1.coutwildrnp tests/data/
```

I'd like to see `':'` as the separator between dataset and layer because `'.'` is generally valid in directory and file names (not true for `':'`) and the split will fail in some cases.

I just tried an unrealistic file index and got GeoJSON out:

```bash
$ fio cat --layer 200000.coutwildrnp tests/data/
```

This should probably fail, yes?

smnorris:
Hmm, that should definitely fail. Probably want to raise one of these: http://click.pocoo.org/5/api/#click.BadParameter

Too bad the indexes can't be validated in the callback.
Thanks for your patience!

sgillies:
Great work @smnorris!

PR #360, opened 2016-06-13
Support GDAL 1.x and 2.x from a single sdist

### LINKED ISSUE #359: Ensure sdists can be built with GDAL 1 or 2
Our current setup script will produce an sdist requiring the version of GDAL found on the packaging system. We need to make sure both `ogrext1.c` and `ogrext2.c` are included in the sdist and that the right one is copied to `ogrext.c` when building.

sgillies:
Resolves #359

Achieved by refactoring so that ogrext1 and ogrext2 contain the only parts of the API that vary between GDAL versions. When making an sdist, we rename each in turn to `ogrext.pyx` and translate to ogrext1.c and ogrext2.c source files. Each of these go in the sdist and the right one is determined by the output of the gdal-config program or other configuration.


sgillies:
Last change: in order to test our packaging of the multi-version extension modules (`ogrext1.pyx` and `ogrext2.pyx`) I've scripted our Travis builds to `python setup.py sdist` and then `pip install` that sdist instead of `pip install -e .`

@geowurster see anything in this that you don't like? There are a lot of changes, but our tests and the new checks on the correctness of our packaging give me a lot of confidence.

sgillies:
Merging later this morning if there are no objections.

geowurster:
@sgillies No time for a full review today, but I skimmed the diff and nothing jumped out at me.

PR #361, opened 2016-06-15
Don't copy GDAL API C files in clean target.

sebastic:
da85e9521e98e05bfa6033a127b48e38069bebf0 broke the clean target again, because it changes the sources when it shouldn't. This was fixed earlier via #322.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -177,6 +177,7 @@ ext_options = dict(
     extra_link_args=extra_link_args)
 
 # When building from a repo, Cython is required.
+ext_modules = []
 if source_is_repo and "clean" not in sys.argv:
     log.info("MANIFEST.in found, presume a repo, cythonizing...")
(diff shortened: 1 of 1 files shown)


sgillies:
It's a little more complicated than this. The source distribution contains an `ogrext1.c` and an `ogrext2.c` file now and we use the one suited to the builder's GDAL API version. I'm going to make a new PR that does this without copying files.

Can you remind me how the clean target is used within a Debian build @sebastic?

sebastic:
The separate targets are called in the various stages of the build:
```
python setup.py clean
python setup.py config
python setup.py build
python setup.py install --root <DESTDIR>
```
The clean target is first executed outside the build chroot, and then again inside the chroot as part of the sequence above (which is executed for every Python version).

sgillies:
Replaced by #362.

PR #362, opened 2016-06-15
Don't copy files for clean target

sgillies:
Debian builds were broken by this regression.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -3,6 +3,11 @@ Changes
 
 All issue numbers are relative to https://github.com/Toblerity/Fiona/issues.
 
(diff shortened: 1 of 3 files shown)


sebastic:
Because `ext_modules` is defined in the the if/elif blocks, if none of the conditions match the variable remains undefined causing an error:
```
$ python2.7 setup.py clean 
  File "setup.py", line 248, in <module>
    ext_modules=ext_modules,
NameError: name 'ext_modules' is not defined
```
This was explicitly handled by my PR, see: [setup.py line 180](https://github.com/Toblerity/Fiona/pull/361/files#diff-2eeaed663bd0d25b7e608891384b7298R180)

PR #367, opened 2016-06-22
Let dict-like subclasses through CRS type checks

geowurster:
https://github.com/mapbox/rasterio/issues/736 introduced a `CRS()` class to Rasterio that subclasses `collections.UserDict()`, so the code below no longer works.  This PR allows `UserDict()` anywhere `dict` is allowed for CRS.

```python
import collections

import fiona
import rasterio


with rasterio.open('tests/data/RGB.byte.tif') as src:
    profile = {
        'crs': src.crs,
        'schema': {
            'properties': {},
            'geometry': 'Point'
        },
        'driver': 'GeoJSON'
    }
    print(src.crs)
    print(type(src.crs))
    print(isinstance(src.crs, collections.UserDict))
    with fiona.open('something.geojson', 'w', **profile) as dst:
        pass
```

output:

```
CRS({'init': 'epsg:32618'})
<class 'rasterio.crs.CRS'>
True
  File "buh.py", line 15, in <module>
    with fiona.open('something.geojson', 'w', **profile) as dst:
  File "/Users/kwurster/code/rasterio/venv/lib/python3.5/site-packages/fiona/__init__.py", line 178, in open
    enabled_drivers=enabled_drivers, crs_wkt=crs_wkt)
  File "/Users/kwurster/code/rasterio/venv/lib/python3.5/site-packages/fiona/collection.py", line 54, in __init__
    raise TypeError("invalid crs: %r" % crs)
TypeError: invalid crs: CRS({'init': 'epsg:32618'})
Exception ignored in:
```


fiona/collection.py
```diff
@@ -50,7 +52,8 @@ def __init__(self, path, mode='r', driver=None, schema=None, crs=None,
             raise TypeError("invalid driver: %r" % driver)
         if schema and not hasattr(schema, 'get'):
             raise TypeError("invalid schema: %r" % schema)
-        if crs and not isinstance(crs, (dict,) + string_types):
+        if crs and not isinstance(
+                crs, (dict, collections.UserDict) + string_types):
```
perrygeo: I'm not a fan of `isinstance` checks for specific subclasses; makes code brittle. What if a user a user constructs the crs as an `OrderedDict` or some other mapping? Unless there is a specific reason not to support them, all mappings could be valid here. 

How about testing against the parent class - all dictionaries should inherit from `collections.Mapping`:

```
import collections
assert isinstance({},collections.Mapping)
assert isinstance(collections.UserDict({}), collections.Mapping)
assert isinstance(collections.OrderedDict({}), collections.Mapping)
```

geowurster: @perrygeo Yep - good point.

PR #368, opened 2016-06-24
Cast int/float/str like values to these types

### LINKED ISSUE #365: Writing np.int64 or np.int32 scalars raises an error
Hi, this is on python3 with fiona `1.7.0.post2`. Attempting to write properties as numpy scalars can fail for certain types. The following example reproduces the problem:

```python
import fiona
import numpy as np
    
feat = {'geometry': {'coordinates': (((21.0, 22.0), (22.0, 23.0), 
                                      (24.0, 20.0), (21.0, 22.0)),),
        'type': 'Polygon'},
        'id': '0',
        'properties': {'PROP': np.int64(12)},  # this line is the problem
        'type': 'Feature'}
schema = {'geometry': 'Polygon', 'properties':{'PROP':'str'}}

with fiona.open('test.shp', 'w', driver="ESRI Shapefile", schema=schema) as c:
    c.write(feat)
```

will raise:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-84-e58056597894> in <module>()
     11 
     12 with fiona.open('test2.shp', 'w', driver="ESRI Shapefile", schema=schema) as c:
---> 13     c.write(feat)

/home/mowglie/.pyvirtualenvs/py3/lib/python3.4/site-packages/fiona/collection.py in write(self, record)
    331     def write(self, record):
    332         """Stages a record for writing to disk."""
--> 333         self.writerecords([record])
    334 
    335     def validate_record(self, record):

/home/mowglie/.pyvirtualenvs/py3/lib/python3.4/site-packages/fiona/collection.py in writerecords(self, records)
    325         if self.mode not in ('a', 'w'):
    326             raise IOError("collection not open for writing")
--> 327         self.session.writerecs(records, self)
    328         self._len = self.session.get_length()
    329         self._bounds = self.session.get_extent()

fiona/ogrext.pyx in fiona.ogrext.WritingSession.writerecs (fiona/ogrext1.c:17022)()

fiona/ogrext.pyx in fiona.ogrext.OGRFeatureBuilder.build (fiona/ogrext1.c:6615)()

ValueError: Invalid field type <class 'numpy.int64'>
```

Note that `np.int` or `np.float` would work. (see https://github.com/geopandas/geopandas/issues/348 and https://github.com/pydata/pandas/issues/13468 for more background).

sgillies:
Resolves #365

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -9,5 +9,6 @@ recursive-include tests *.py
 recursive-include tests/data *
 include fiona/*.c fiona/*.cpp
 exclude fiona/ogrext.c
+exclude fiona/ogrext.pyx
 include CHANGES.txt CREDITS.txt LICENSE.txt VERSION.txt README.rst
 include benchmark.py setup.py requirements.txt
--- a/fiona/compat.py
+++ b/fiona/compat.py
@@ -1,10 +1,17 @@
 import collections
+import sys
+
 from six.moves import UserDict
+
 try:
     from collections import OrderedDict
-except ImportError:    
+except ImportError:
     from ordereddict import OrderedDict
 
+if sys.version_info[0] > 2:
+    long = int
+else:
+    long = long
 
 # Users can pass in objects that subclass a few different objects
 # More specifically, rasterio has a CRS() class that subclasses UserDict()
--- a/fiona/ogrext1.pyx
+++ b/fiona/ogrext1.pyx
@@ -313,7 +313,7 @@ cdef class OGRFeatureBuilder:
                 hh, mm, ss = value.hour, value.minute, value.second
                 ogrext1.OGR_F_SetFieldDateTime(
                     cogr_feature, i, 0, 0, 0, hh, mm, ss, 0)
-            elif schema_type == 'str': #isinstance(value, string_types):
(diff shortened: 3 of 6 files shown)


sgillies:
The gist of this one: cast feature property values to their schema type before sending to OGR. This allows one to write features having a property like `"foo": np.float64(1.0)`.

Any comments on this one? Good idea? Bad idea?

geowurster:
@sgillies It will be a little while until I get to this one, so if its blocking a release someone else should look at it.

That said, interesting idea.  I have occasionally wished I could define a layer's properties with the builtin Python types, but not enough to try.  Numpy makes a certain amount of sense too because you know way more about the intended field width and precision.

geowurster:
> ... if it makes interfacing scientific libraries with Fiona easier.

Apparently editing the text in a PR review is not possible.

sgillies:
Thanks for the review @geowurster. I think I'm going to back burner this one. Developers have the option of converting, so they're not really blocked. And there are bigger questions about schemas that I think we should ponder before adding new schema features. Like, are Fiona's schemas good enough? Standard enough? Why not jsonchema? Etc.

PR #375, opened 2016-07-29
Fix bug, update sphinx-apidoc source path in doc makefile

qinfeng:


sgillies:
Looks good, I'm going to give it a test this weekend.

sgillies:
This is great, thanks!

PR #382, opened 2016-08-28
WIP: add support for binary data fields

### LINKED ISSUE #381: Add support for binary field type
Today I needed to read some binary data stored in a BLOB column of an Esri geodatabase. I had to resort to using the SWIG OGR bindings (*gasp*).

https://gis.stackexchange.com/questions/208182/reading-blob-data-from-an-esri-gdb-with-ogr

I don't think this would be too difficult to implement in Fiona. The data would be returned as a `bytes` type, something like:

```
feature = {"name": "some data", "geometry": {...}, "blob_col": b"..."}
```

I'll give this a go when I next find some time.

snorfalorpagus:
This PR adds support for binary data fields.

Closes #381.


fiona/ogrext1.pyx
```diff
@@ -151,6 +153,8 @@ cdef class FeatureBuilder:
         cdef int mm = 0
         cdef int ss = 0
         cdef int tz = 0
+        cdef unsigned char *data
+        cdef int l
```
sgillies: `cdef Py_ssize_t l = 0`

fiona/ogrext1.pyx
```diff
@@ -210,6 +214,11 @@ cdef class FeatureBuilder:
                 else:
                     props[key] = datetime.datetime(
                         y, m, d, hh, mm, ss).isoformat()
+
+            elif fieldtype is bytes:
+                data = ogrext1.OGR_F_GetFieldAsBinary(feature, i, &l)
+                props[key] = PyBytes_FromStringAndSize(<char*>data, <Py_ssize_t>l)
```
sgillies: @snorfalorpagus let's let Cython handle the bytes object creation for us, like

```python
blob = data[:l]
props[key] = blob
```

See http://cython.readthedocs.io/en/latest/src/tutorial/strings.html#passing-byte-strings.

fiona/ogrext1.pyx
```diff
@@ -322,6 +331,9 @@ cdef class OGRFeatureBuilder:
                     raise
                 string_c = value_bytes
                 ogrext1.OGR_F_SetFieldString(cogr_feature, i, string_c)
+            elif isinstance(value, bytes):
+                ogrext1.OGR_F_SetFieldBinary(cogr_feature, i, len(value),
+                    <unsigned char*>PyBytes_AsString(value))
```
sgillies: Likewise, 

```python
string_c = value
ogrext1.OGR_F_SetFieldBinary(cogr_feature, i, len(value), string_c)
```

sgillies:
@snorfalorpagus I added a few suggestions. Overall this looks great and will be a nice feature.

fiona/ogrext1.pyx
snorfalorpagus: I wasn't aware of this syntax. I'll update it.

snorfalorpagus:
@sgillies Any suggestion on how we test this on the CI without GPKG support?

sgillies:
@snorfalorpagus there's a `featureRT()` method in `ogrext1.pyx` and `ogrext2.pyx` that takes a feature object (dict) and returns another feature object that has been converted to/from an OGR feature. I don't feel like we need to test the geopackage or geodatabase drivers here.

snorfalorpagus:
`featureRT()` looks perfect. I'll use that.

snorfalorpagus:
@sgillies `featureRT` isn't as perfect as I though. It still needs a `collection` object which supports binary fields, which the CI does not have.

fiona/ogrext1.pyx
snorfalorpagus: This will need casting to an `int` later, as that's what `OGR_F_GetFieldAsBinary` expects. Is that OK?

sgillies: Yes

snorfalorpagus:
@sgillies Any idea how we can progress this?

sgillies:
@snorfalorpagus the difficulty of testing this is all my fault. After you resolve the conflicts, I will merge and then will look at the refactoring needed to make this easier to test in a different PR.

fiona/ogrext1.pyx
snorfalorpagus: This doesn't work. OGR_F_GetFieldAsBinary needs a pointer to an int so that it can write the output to it. I think it's correct as it is.

snorfalorpagus:
I've rebased against master.

sgillies:
🙏 @snorfalorpagus

PR #385, opened 2016-10-03
Protect against bad dates

sposs:
In case the date in the input isn't correct (yielding any of y=0, and/or m=0, and/or d=0), there is an error in datetime.date(). Protect against such case by defaulting to unix timestamp 0 (Jan 1st, 1970)

### DIFF
--- a/fiona/ogrext1.pyx
+++ b/fiona/ogrext1.pyx
@@ -203,6 +203,12 @@ cdef class FeatureBuilder:
             elif fieldtype in (FionaDateType, FionaTimeType, FionaDateTimeType):
                 retval = ogrext1.OGR_F_GetFieldAsDateTime(
                     feature, i, &y, &m, &d, &hh, &mm, &ss, &tz)
+                if not y:
+                    y = 1970
+                if not m:
+                    m = 1
+                if not d:
+                    d = 1
                 if fieldtype is FionaDateType:
                     props[key] = datetime.date(y, m, d).isoformat()
                 elif fieldtype is FionaTimeType:
--- a/fiona/ogrext1.pyx
+++ b/fiona/ogrext1.pyx
@@ -203,8 +203,6 @@ cdef class FeatureBuilder:
             elif fieldtype in (FionaDateType, FionaTimeType, FionaDateTimeType):
                 retval = ogrext1.OGR_F_GetFieldAsDateTime(
                     feature, i, &y, &m, &d, &hh, &mm, &ss, &tz)
-                if not y or not m or not d:
-                    log.warning("Invalid date")
                 try:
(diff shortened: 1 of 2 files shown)


snorfalorpagus:
I think this goes against the zen "errors should never pass silently". The user should be warned that the data is invalid (`warnings.warn()`?) and perhaps return a `None` rather that a date object.

sposs:
Fair enough. I'm up for anything that would avoid the error. As for the return, I don't know what the impact of such decision is...

fiona/ogrext1.pyx
```diff
@@ -203,13 +203,18 @@ cdef class FeatureBuilder:
...
                     feature, i, &y, &m, &d, &hh, &mm, &ss, &tz)
-                if fieldtype is FionaDateType:
-                    props[key] = datetime.date(y, m, d).isoformat()
-                elif fieldtype is FionaTimeType:
-                    props[key] = datetime.time(hh, mm, ss).isoformat()
-                else:
-                    props[key] = datetime.datetime(
-                        y, m, d, hh, mm, ss).isoformat()
+                if not y or not m or not d:
+                    log.warning("Invalid date")
```
sgillies: Let's remove these two lines and (see next comment)...

fiona/ogrext1.pyx
```diff
@@ -203,13 +203,18 @@ cdef class FeatureBuilder:
...
+                try:
+                    if fieldtype is FionaDateType:
+                         props[key] = datetime.date(y, m, d).isoformat()
+                    elif fieldtype is FionaTimeType:
+                         props[key] = datetime.time(hh, mm, ss).isoformat()
+                    else:
+                         props[key] = datetime.datetime(
+                             y, m, d, hh, mm, ss).isoformat()
+                except ValueError:
+                    props[key] = None
```
sgillies: ... and log the actual exception that occurred, which will be something like `ValueError: year is out of range`, using https://docs.python.org/2/library/logging.html#logging.Logger.exception before setting the property value to `None`.

fiona/ogrext2.pyx
```diff
@@ -217,13 +217,18 @@ cdef class FeatureBuilder:
...
+                try:
+                    if fieldtype is FionaDateType:
+                        props[key] = datetime.date(y, m, d).isoformat()
+                    elif fieldtype is FionaTimeType:
+                        props[key] = datetime.time(hh, mm, ss).isoformat()
+                    else:
+                        props[key] = datetime.datetime(
+                            y, m, d, hh, mm, ss).isoformat()
+                except ValueError:
+                    props[key] = None
```
sgillies: Same here as in comments above.

sposs:
Both files have been modified...

PR #386, opened 2016-10-04
Single gdal version jobs

sgillies:
This PR builds only one version of GDAL per Travis job and prunes all Python versions but 2.7 and 3.5. Fiona doesn't use any 3-only features and so 3.5 is as good as any other Python 3.x version. Benefits: faster builds when we have no caches, such as when we want to update GDAL versions, and smaller caches.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ addons:
     - gfortran
 python:
   - "2.7"
-  - "3.4"
+  - "3.5"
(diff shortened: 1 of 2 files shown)


snorfalorpagus:
This looks good to me.

Building GDAL takes up the majority of the total build time, but I was surprised at how little that is in absolute terms (a few minutes).

PR #398, opened 2016-11-17
Added support for layer creation options and a test.

### LINKED ISSUE #388: Error writing to FileGDB "The spatial index grid size is invalid"
I've hit a bit of an interesting edge case with creating FileGDBs using Fiona.

**tl;dr** I have a dataset of MultiPolygons of widely ranging sizes that I need to write to a FileGDB with a single layer. When the first feature in the dataset is several orders of magnitude smaller than the largest feature Fiona encounters an `ERROR:Fiona:CPLE_AppDefined in Failed at writing Row to Table in CreateFeature. (The spatial index grid size is invalid.)` error.

It appears that the first feature written is used to setup the internal spatial domain of a FileGDB[1], though I've not been able to find any documentation from ESRI, the OpenFileGDB reverse engineering, or through general research that confirms this. It is documented[2] that exceeding this domain causes issues in spatial indexing though.

In practice however, we see that the first feature does appear to dictate the bounds of the internal spatial domain.

See https://gist.github.com/keithamoss/e4c15cb8ecef143901b6131baad91bbd for the three test datasets.

**NB: I've used `fio` to debug the issue, but in practice this is part of a larger Flask+Fiona+GeoPandas app.**
### **small-geom-first.json: A GeoJSON dataset with two polygons (small polygon first, large polygon second)**

`fio dump small-geom-first.json | fio load test.gdb --driver FileGDB --src-crs EPSG:4326`

**Creates FileGDB containing the small polygon, then aborts.**

```
WARNING:Fiona:CPLE_NotSupported in dataset /app/tmp/filegdb-bug/test.gdb does not support layer creation option ENCODING
ERROR:Fiona:CPLE_AppDefined in Failed at writing Row to Table in CreateFeature. (The spatial index grid size is invalid.)
  File "/usr/local/lib/python2.7/site-packages/fiona/fio/load.py", line 103, in load
    dst.writerecords(source)
  File "/usr/local/lib/python2.7/site-packages/fiona/collection.py", line 327, in writerecords
    self.session.writerecs(records, self)
  File "fiona/ogrext.pyx", line 1038, in fiona.ogrext.WritingSession.writerecs (fiona/ogrext2.c:17532)
RuntimeError: Failed to write record: {u'geometry': {u'type': u'Polygon', u'coordinates': [[[127.29999999999998, -29.01], [119.49, -32.28], [118.43, -27.1], [125.9, -21.44], [127.29999999999998, -29.01]]]}, u'type': u'Feature', u'id': u'1', u'properties': {}}
Aborted!
```
### **first-geom-wa-extent.json: A GeoJSON dataset with three polygons (polygon with Western Australia's extent first, small polygon second, large polygon third)**

`fio dump /app/tmp/filegdb-bug/first-geom-wa-extent.json | fio load /app/tmp/filegdb-bug/test.gdb --driver FileGDB --src-crs EPSG:4326`

**Success! Creates FileGDB containing both records.**
### **first-geom-empty.json: A GeoJSON dataset with three polygons (polygon with empty geometry first, small polygon second, large polygon third)**

This fails via `fio dump | fio load` with an `ValueError: Record's geometry type does not match collection schema's geometry type: u'Polygon' != u'MultiPolygon'` error, but does successfully create a FileGDB if run through Fiona in Python (see code below).
## ogr2ogr

Running all of these GeoJSON files through `ogr2ogr` will successfully create a FileGDB.

I'm unsure (hence filing this bug) if the issue here lays with Fiona's interfacing with GDAL, or if GDAL via `ogr2ogr` simply has a few more smarts to handle setting the correct internal spatial domain by inspecting the features as it's creating the FileGDB.
## FileGDB Layer Creation Options

_This may be worth a separate issue, but for context I'll note it here._

Having gotten this far I was hoping to test my theory further by passing the **XORIGIN, YORIGIN, and XYSCALE** layer creation options[2] available in GDAL's FileGDB driver in order to define a grid sufficient to hold all of the features (essentially, the extent of Western Australia).

According to #90 I should just be able to pass them to to `fiona.drivers()`, but this appears to have no effect when it comes to FileGDBs and / or these specific layer creation options.

``` python
import fiona
with fiona.drivers(GEOMETRY_NAME="SHAPEY", FID="OID"):
        with fiona.collection(os.path.join(TMP_DIR, "filegdb-bug/first-geom-wa-extent.json"), "r") as source:
            print source.meta
            with fiona.open(os.path.join(TMP_DIR, "test.gdb"), "w", driver="FileGDB", schema=source.meta["schema"], crs=source.meta["crs"]) as sink:
                for f in source:
                    print f
                    sink.write(f)
```

The resulting FileGDB run through `ogrinfo` reports the defaults for geom name and fid are still in use:

```
FID Column = OBJECTID
Geometry Column = SHAPE
```

Fiona 1.7.0-post2, GDAL 2.1.1, Python 2.7

[1] http://webhelp.esri.com/arcgisdesktop/9.2/index.cfm?id=2059&pid=2057&topicname=Setting_spatial_indexes
[2] http://www.gdal.org/drv_filegdb.html (See the _Important_ note under "XORIGIN, YORIGIN, ZORIGIN, XYSCALE, ZSCALE".)

sgillies:
Resolves #388

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -8,7 +8,7 @@ All issue numbers are relative to https://github.com/Toblerity/Fiona/issues.
 
 Bug Fixes:
 
-- Prevent Fiona from stumbling over '*Z', '*M', and '*ZM' geometry types
+- Prevent Fiona from stumbling over 'Z', 'M', and 'ZM' geometry types
   introduced in GDAL 2.1 (#384). Fiona 1.7.1 doesn't add explicit support for
   these types, they are coerced to geometry types 1-7 ('Point', 'LineString',
   etc.)
--- a/fiona/__init__.py
+++ b/fiona/__init__.py
@@ -100,7 +100,8 @@ def open(
         layer=None,
         vfs=None,
         enabled_drivers=None,
-        crs_wkt=None):
+        crs_wkt=None,
+        **kwargs):
     """Open file at ``path`` in ``mode`` "r" (read), "a" (append), or
     "w" (write) and return a ``Collection`` object.
 
@@ -165,7 +166,7 @@ def open(
             raise IOError("no such file or directory: %r" % path)
         c = Collection(path, mode, driver=driver, encoding=encoding,
                        layer=layer, vsi=vsi, archive=archive,
-                       enabled_drivers=enabled_drivers)
+                       enabled_drivers=enabled_drivers, **kwargs)
     elif mode == 'w':
         if schema:
             # Make an ordered dict of schema properties.
@@ -175,7 +176,8 @@ def open(
             this_schema = None
         c = Collection(path, mode, crs=crs, driver=driver, schema=this_schema,
                        encoding=encoding, layer=layer, vsi=vsi, archive=archive,
-                       enabled_drivers=enabled_drivers, crs_wkt=crs_wkt)
+                       enabled_drivers=enabled_drivers, crs_wkt=crs_wkt,
+                       **kwargs)
     else:
         raise ValueError(
             "mode string must be one of 'r', 'w', or 'a', not %s" % mode)
--- a/fiona/ogrext1.pyx
+++ b/fiona/ogrext1.pyx
@@ -693,7 +693,7 @@ cdef class WritingSession(Session):
     
     cdef object _schema_mapping
 
-    def start(self, collection):
+    def start(self, collection, **kwargs):
         cdef void *cogr_fielddefn
         cdef void *cogr_driver
         cdef void *cogr_ds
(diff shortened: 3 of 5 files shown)


sgillies:
The test that this works is calling

```python
with fiona.open('test.json', 'w', driver='GeoJSON', coordinate_precision=2, ...) as dst:
    # write records.
```

And then checking the output to make sure that coordinate precision is indeed 2.

sgillies:
Appveyor error looks spurious to me. Merging.

PR #411, opened 2017-01-25
Add extension to VSI filenames so we can access zipped bytes

### LINKED ISSUE #318: How to extract features from in-memory zip files?
Hi,

I have a Flask application in which at a certain point the user uploads a zip file containing all the required shapefiles and accompanying files for the project. Currently, I extract the zip file to the server's file system and fetch the features using Fiona. However, I would like to NOT use the server's file system. In Python, I  can use ZipFile to avoid using the file system and yet managing to do as I please with the files. In Fiona, I know about the ability to use Virtual File Systems (VFS), but it requires me to provide a full path within the server which would still require me to use the file system.

Any ideas?

Thank you!

sgillies:
Resolves #318

### DIFF
--- a/fiona/collection.py
+++ b/fiona/collection.py
@@ -478,9 +478,10 @@ def vsi_path(path, vsi=None, archive=None):
     # an OGR VSI path (see cpl_vsi.h).
     if vsi:
         if archive:
-            result = os.path.join('/', 'vsi{}'.format(vsi), archive.strip('/'), path.strip('/'))
+            result = '/'.join([
+                '/vsi{}'.format(vsi), archive.strip('/'), path.strip('/')])
         else:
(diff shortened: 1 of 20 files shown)


sgillies:
A lot of test files were touched to shhh the debug logging, but nevermind those changes.

mwtoews:
a future tip for supporting UNIX-style paths on all OSes, including Windows:
```
import posixpath
posixpath.join('/', 'a', 'b')  # '/a/b'
```
See more at the second note at the top of https://docs.python.org/2/library/os.path.html

However, that said, there's nothing wrong with `'/'.join(['/root', 'a', 'b'])`

sgillies:
@mwtoews thanks, I'm going to switch to that.

Issue #414, opened 2017-01-25
Appveyor build is failing


snorfalorpagus:
The current appveyor builds are failing. It looks like just a single test:

```
================================== FAILURES ===================================
______________________________ test_multi_layer _______________________________
    def test_multi_layer():
        layerdef = "1:coutwildrnp,1:coutwildrnp"
        runner = CliRunner()
        result = runner.invoke(
            cat.cat, ['--layer', layerdef, 'tests/data/'])
>       assert result.output.count('"Feature"') == 134
E       assert 0 == 134
E        +  where 0 = <built-in method count of str object at 0x00000000047F68B0>('"Feature"')
E        +    where <built-in method count of str object at 0x00000000047F68B0> = 'Aborted!\n'.count
E        +      where 'Aborted!\n' = <Result SystemExit(1,)>.output
tests\test_fio_cat.py:80: AssertionError
---------------------------- Captured stderr call -----------------------------
DEBUG:Fiona:Creating a not-responsible GDALEnv in drivers()
DEBUG:Fiona:Option CPL_DEBUG=b'OFF'
ERROR:fio:Exception caught during processing
Traceback (most recent call last):
  File "C:\projects\fiona\fiona\fio\cat.py", line 69, in cat
    with fiona.open(path, layer=lyr) as src:
  File "C:\projects\fiona\fiona\__init__.py", line 169, in open
    enabled_drivers=enabled_drivers, **kwargs)
  File "C:\projects\fiona\fiona\collection.py", line 147, in __init__
    self.session.start(self)
  File "fiona\ogrext.pyx", line 435, in fiona.ogrext.Session.start (fiona/ogrext.c:9107)
    raise FionaValueError(
fiona.errors.FionaValueError: No dataset found at path 'tests/data/' using drivers: *
========= 1 failed, 525 passed, 44 skipped, 4 xpassed in 9.56 seconds =========
```

https://ci.appveyor.com/project/sgillies/fiona/build/1.0.178

rbuffat:
Looks like replacing tests/data/ with os.path.join('tests', 'data') did the trick (See https://github.com/Toblerity/Fiona/pull/415).

However there are a bunch of other tests in tests/test_fio_cat.py disabled for windows.

rbuffat:
I started looking into the on windows disabled tests in this branch: https://github.com/rbuffat/Fiona/tree/fix_windows_tests
It looks that most of them fail due to the different path separator.

rbuffat:
[...]

PR #415, opened 2017-01-26
Fixappveyor

rbuffat:
this fixes for me the test on appveyor. Let's see if it really works.

See https://github.com/Toblerity/Fiona/issues/414


sgillies:
🙏 @rbuffat

PR #417, opened 2017-01-29
Fix windows tests

rbuffat:
This is a PR for https://github.com/Toblerity/Fiona/issues/414

Especially two of the commits should be reviewed:

In 90846683f8cfe35103a193e31367d990f10a43ab, / is used a path separator for /vsimem/ paths. Unfortunately my editor removed some whitespace, only line `vsi_filename = os.path.join('/vsimem', uuid.uuid4().hex)` changed to `vsi_filename = '/vsimem/{}'.format(uuid.uuid4().hex)`. 

In 9c1b59ac2b83d77a9cc2120164acff8d9059d709, the encoding of a bytescollection is fixed to utf8 instead of  the encoding of the current session. 

There are still FIXME Windows in test_multiconxn.py. 
Currently a lot of tests use the same "hardcoded" path. This could be refactored in a future step.

### DIFF
--- a/fiona/collection.py
+++ b/fiona/collection.py
@@ -441,7 +441,7 @@ class BytesCollection(Collection):
         self.virtual_file = buffer_to_virtual_file(self.bytesbuf)
 
         # Instantiate the parent class.
-        super(BytesCollection, self).__init__(self.virtual_file)
+        super(BytesCollection, self).__init__(self.virtual_file, encoding='utf-8')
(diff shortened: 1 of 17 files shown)


sgillies:
@rbuffat Thank you! I've left a few comments.

I need to merge the maint-1.7 branch before this and there will be a few conflicts introduced, but nothing major.

Issue #422, opened 2017-02-14
BytesCollection cannot open ESRI flavored JSON 


youngpm:
I've got some strange behavior when trying to get a `BytesCollection` to open some ESRI flavored JSON.  Using [esri.json](https://gist.githubusercontent.com/youngpm/347b8a0a3ea724afa616ea2387104972/raw/1b95631f684d0bf03d585a7c52e9c31dbf8bacc6/esri.json) as my input file and running 
```
fio dump esri.json
```
I get the expected result. Similarly, 
```
In [1]: import fiona
In [2]: with fiona.open('esri.json') as src:
   ...:     recs = [r for r in src]
In [3]: len(recs)
Out[3]: 93
```
works great.  However, when trying to do it with `BytesCollection`, 
```
In [1]: import fiona
In [2]: with open('esri.json') as src:
   ...:     buf = src.read()
In [3]: fiona.BytesCollection(buf)
---------------------------------------------------------------------------
FionaValueError                           Traceback (most recent call last)
<ipython-input-3-1ccda61140ae> in <module>()
----> 1 fiona.BytesCollection(buf)

.../lib/python2.7/site-packages/fiona/collection.pyc in __init__(self, bytesbuf)
    427
    428         # Instantiate the parent class.
--> 429         super(BytesCollection, self).__init__(self.virtual_file)
    430
    431     def close(self):

.../lib/python2.7/site-packages/fiona/collection.pyc in __init__(self, path, mode, driver, schema, crs, encoding, layer, vsi, archive, enabled_drivers, **kwargs)
    138             self.encoding = encoding
    139             self.session = Session()
--> 140             self.session.start(self)
    141
    142             # If encoding param is None, we'll use what the session

fiona/ogrext.pyx in fiona.ogrext.Session.start (fiona/ogrext.c:7197)()

FionaValueError: No dataset found at path '/vsimem/30332495866d43b3b7e8f63f74d68474' using drivers: *
```
Interestingly, if you first convert the esri stuff to geojson via `fio dump esri.json > not-esri.json`, then everything is copacetic: 
```
In [1]: import fiona
In [2]: with open('not-esri.json') as src:
   ...:     buf = src.read()
In [3]: c = fiona.BytesCollection(buf)
In [4]: len(c)
Out[4]: 93
```
[...]

PR #423, opened 2017-02-18
Geojson driver and bytes buffer

youngpm:
This addresses the issue discussed in #422.

By default, the BytesCollection class maps the buffer to a filename that has no extension. GDAL will happily check each driver in its registry to see if it can open `/vsimem/<somerandomhash>`.  The GeoJSON driver has logic in it to quickly see if it should open the file, as you can see [here](https://github.com/OSGeo/gdal/blob/901ce5be4d8a3ba85224134e80375ae70bbd7104/gdal/ogr/ogrsf_frmts/geojson/ogrgeojsonutils.cpp#L128-L148).  In that snippet, you can see that if passed a file path whose extension is not recognized as json, it will go ahead and [check the first 6000 bytes to see if its json-like](https://github.com/OSGeo/gdal/blob/901ce5be4d8a3ba85224134e80375ae70bbd7104/gdal/ogr/ogrsf_frmts/geojson/ogrgeojsonutils.cpp#L68-L79) by searching for the presence of certain keys that one would expect to find in GeoJSON, ESRI JSON, or TopoJSON.

The newly minted `tests/data/grenada.geojson` [file as seen here](http://bl.ocks.org/d/18ca7e12cc0f89f0d3180c969c4b826b) is an example of valid GeoJSON where the keys that GDAL is searching for are found past the first 6000 bytes.  This causes the GeoJSON driver to be skipped and a failure to occur when trying to open it with fiona despite of its validity. 

The fix is to append an appropriate extension to the `vsimem` path; I have done so only in the case when the `GeoJSON` driver is explicitly called out.  The new test demnstrates both that the expected exception is raised when trying to read this file when the GeoJSON driver is not specified, and that we succeed when it is.

What do you think?  My feelings won't be hurt if this is too esoteric of a change.  In #422, a hack is described where you can feed the json straight to `fiona.Collection` class directly, but that work around is not guaranteed to work in the future.


sgillies:
@youngpm this PR is great. Would you be willing to aim it at the main-1.7 branch so I can get it into a 1.7.4 release? 1.8 is still a little ways out.

youngpm:
:+1: The appveyor test failed with this switch to the maintenance branch, but I think from looking at the history it was already failing.

Issue #437, opened 2017-04-03
Handle OGROpen errors in session.start()


sgillies:
## Expected behavior and actual behavior.

Opening an existing GeoJSON to overwrite should succeed or raise an IO exception. Instead, we see `ValueError: null layer inside session.start()`. See #436.

## Steps to reproduce the problem.

Modification of tests/test_geojson.py's `WriteTest.test_json()` demonstrates the issue.

## Operating system

Any

## Fiona version and provenance

1.7.5

sgillies:
Raising this exception in `start()` also causes problems for the context manager protocol. The collection and session are left in a error state that trips up code called when the manager exits. I see double errors in some cases.

sgillies:
Resolved by #444.

PR #444, opened 2017-04-25
Context errors

sgillies:
Towards a fix of #437


sgillies:
Summary:

I've ported `exc_wrap_pointer()` from Rasterio and am using it to get exceptions from GDAL in a way that's better than using `with cpl_errs:` (which I'd like to remove in the future).

And, I'm making sure that if `self.session.start()` fails within `Collection()` or `fiona.open()`, the collection's session is set back to `None`.

PR #445, opened 2017-05-11
Add GDAL 2.2.0 to Travis build matrix

geowurster:
Released yesterday: http://download.osgeo.org/gdal/2.2.0/

### DIFF
--- a/fiona/ogrext2.pyx
+++ b/fiona/ogrext2.pyx
@@ -35,6 +35,10 @@ from cpython cimport PyBytes_FromStringAndSize, PyBytes_AsString
 
 
 log = logging.getLogger("Fiona")
+class NullHandler(logging.Handler):
+    def emit(self, record):
+        pass
+log.addHandler(NullHandler())
 
 # Mapping of OGR integer field types to Fiona field type names.
 #
@@ -136,6 +140,12 @@ def get_gdal_release_name():
     """Return release name of gdal"""
     return ogrext2.GDALVersionInfo("RELEASE_NAME")
 
+# check compiled version against runtime version
+# only compare the major and minor version number
+GDAL_RUNTIME_VERSION = get_gdal_version_num()
+if GDAL_VERSION_NUM // 1000 != GDAL_RUNTIME_VERSION // 1000:
+    log.warning("Fiona was compiled against GDAL {} but run with GDAL {}".format(GDAL_VERSION_NUM, GDAL_RUNTIME_VERSION))
+
 
 # Feature extension classes and functions follow.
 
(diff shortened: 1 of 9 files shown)


snorfalorpagus:
I tried restarting the Travis build, but it's still falling over on something related to `pip` and the `six` module:

https://travis-ci.org/Toblerity/Fiona/jobs/231022585

rbuffat:
@snorfalorpagus looks like this bug here: https://github.com/pypa/setuptools/issues/951

sgillies:
I can't tell whether pinning setuptools back to 33.0 will fix this or not, but I'll give it a try.

Nevertheless, I see in one test failure that the default name of a GeoJSON file's single layer is no longer 'OgrGeoJSON'. Seems to be the only problem :)

micahcochran:
Could this be closed and reopened to retest the PR?

snorfalorpagus:
The previous commit fixes the existing errors related to GeoJSON. I'm working on a fix for #460.

snorfalorpagus:
I can't figure out how to get Cython to compile something that works in both GDAL < 2.2 and GDAL >= 2.2 at the same time. I tried using the compile time directives but they don't seem to work the way I expected. If we can figure this out it'll be the same answer for how to merge ogrext1 and ogrext2 into a single source. Can anyone else figure this out?

snorfalorpagus:
This is working now. It was a stupid mistake in checking the version numbers.

geowurster:
@snorfalorpagus Thanks for picking this up!

fiona/ogrext2.pyx
```diff
@@ -35,6 +35,10 @@ from cpython cimport PyBytes_FromStringAndSize, PyBytes_AsString
 
 
 log = logging.getLogger("Fiona")
+class NullHandler(logging.Handler):
+    def emit(self, record):
+        pass
+log.addHandler(NullHandler())
```
snorfalorpagus: Not sure if this should be here? I was a little confused with the rebase.

sgillies: Yes, it belongs. The way I did logging early on was ignorant of how Python's hierarchical logger works and is wrong. In the future we should remove all these `NullHandler` classes, but keep it for now.

tests/test_feature.py
```diff
@@ -110,3 +111,18 @@ def test_properties(self):
         g = featureRT(f, self.c)
         self.assertEqual(g['properties']['title'], 'foo')
 
+def test_feature_null_field(tmpdir):
+    """
+    In GDAL 2.2 the behaviour of OGR_F_IsFieldSet slightly changed.
+    See GH #460.
```
sgillies: Fantastic, @snorfalorpagus!

Issue #450, opened 2017-06-02
ImportWarning in python 3.6


perrygeo:
## Expected behavior and actual behavior.

Running the `fio` command line gives an `ImportWarning`, expected no warnings.

```
⭆ python /Users/mperry/env/dev36/bin/fio
/Users/mperry/env/dev36/lib/python3.6/importlib/_bootstrap.py:205: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
```

## Steps to reproduce the problem.

If we tell the Python interpreter to raise an exception on this warning, we get a more useful traceback

```
⭆ python -W error:can\'t /Users/mperry/env/dev36/bin/fio
Traceback (most recent call last):
  File "/Users/mperry/env/dev36/bin/fio", line 11, in <module>
    load_entry_point('Fiona==1.7.5', 'console_scripts', 'fio')()
  File "/Users/mperry/env/dev36/lib/python3.6/site-packages/pkg_resources/__init__.py", line 560, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/Users/mperry/env/dev36/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2648, in load_entry_point
    return ep.load()
  File "/Users/mperry/env/dev36/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2302, in load
    return self.resolve()
  File "/Users/mperry/env/dev36/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2308, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/Users/mperry/env/dev36/lib/python3.6/site-packages/fiona/__init__.py", line 69, in <module>
    from fiona.collection import Collection, BytesCollection, vsi_path
  File "/Users/mperry/env/dev36/lib/python3.6/site-packages/fiona/collection.py", line 9, in <module>
    from fiona.ogrext import Iterator, ItemsIterator, KeysIterator
  File "fiona/_geometry.pxd", line 35, in init fiona.ogrext (fiona/ogrext2.c:24090)
  File "fiona/_geometry.pyx", line 3, in init fiona._geometry (fiona/_geometry.c:8052)
ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
```

## Operating system, Fiona version and provenance

OSX Sierra, Fiona 1.7.5 both wheel and `--no-binary`, Python 3.6.1 from python.org

[...]

PR #452, opened 2017-06-03
New maint 1.7

sgillies:


sgillies:
Resolves #450 and some other deprecation warnings.

Issue #277, opened 2015-09-28
Refactor extension classes


sgillies:
After #275 is finished, I'd like to try refactoring to eliminate redundant code. Ideally, ogrext1.pyx would have GDAL 1.x compatible Session and Iterator classes that derive from a base classes and ogrext2.pyx would have GDAL 2.x compatible classes that derive from the same base classes.

snorfalorpagus:
What's the best way to do this? Concatenate an `ogrext_base.pyx` and the relevant `ogrext1.pyx` or `ogrext2.pyx` in `setup.py`?

snorfalorpagus:
There is only 1 line different in the `writerec` method between the two versions. Does this mean the entire function needs to be duplicated? In C you would use an `#ifdef` but there isn't an equivalent in Cython. Do we just have to put up with this?

https://github.com/Toblerity/Fiona/blob/master/fiona/ogrext1.pyx#L936
https://github.com/Toblerity/Fiona/blob/master/fiona/ogrext2.pyx#L1000

micahcochran:
There are two approaches that I can think of. Write a common base class with all the functions that will compile in both versions of GDAL.  Then have your GDAL version specific classes derive from a those common base classes.

The other is to write an emulation layer.  So, the Fiona code may target GDAL 2.0, but have wrapper functions for GDAL 1.x in the version 2 API's names. This could present a problem in the future if there are functions that Fiona wants to use in GDAL 2.x that aren't easily supported by GDAL 1.x API.

There are a few other ways to do this that add more layers of abstraction. There are probably a few other ways to go.  Both of the approaches should reduce the lines of code.  Which approach is less confusing?

snorfalorpagus:
It looks like Cython does support some preprocessor-like functionality. I wasn't aware of this.

http://docs.cython.org/src/reference/language_basics.html#compile-time-definitions

sgillies:
I'm going to push this one off to a future milestone unless there are objections. It's time to get 1.7 out there!

snorfalorpagus:
I agree that this shouldn't hold back the release of 1.7 as there is a real need to have GDAL 2.x support now.

snorfalorpagus:
Closed by #461.

PR #461, opened 2017-07-11
Merge ogrext1.pyx and ogrext2.pyx into ogrext.pyx

snorfalorpagus:
This is an attempt to remove the need for separate source files for GDAL 1.x vs 2.x (#277). It uses Cython's compile time definitions feature (used in #445) to allow the same code to be used for either library version. The majority of changes can be accounted for with just a few function alias (`from ogrext1 cimport X as Y`). Where function signatures have changed these sections are wrapped in `IF GDAL_VERSION_NUMBER >= 2000000` sections. I was surprised at how few of these were needed. This is a fairly major change and needs some serious scrutiny but I think it could be the way forward. I've not thought about users compiling from source but without cython - do we really need to support that anyway?


sgillies:
> I've not thought about users compiling from source but without cython - do we really need to support that anyway?

@snorfalorpagus source distributions with .c and no .pyx code are still the state of the art. At the very least they cut down build time. Let's keep the separate files for now and look at refactoring them like I've done for rasterio: we could have a core ogrext.pxi that's included into smaller ogrext1.pyx ad ogrext2.pyx files.

(Note to myself: our .c files are filled with extraneous comments and can probably be optimized in future releases)

PR #462, opened 2017-07-11
Revert "Add GDAL 2.2.0 to Travis build matrix"

sgillies:
Reverts Toblerity/Fiona#445

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ env:
     - GDALVERSION="1.11.5"
     - GDALVERSION="2.0.3"
     - GDALVERSION="2.1.1"
-    - GDALVERSION="2.2.0"
 addons:
   apt:
     packages:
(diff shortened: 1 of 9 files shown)


sgillies:
@snorfalorpagus after merging this and then commenting on #461, I realized that we can't do this: the Cython directive is inconsistent with our approach to making C-only source distibutions.

Instead of making you redo this, how about I adapt what you've done in a new branch and then submit a PR for maint-1.7? I'd like to be able to support 2.2.0 there and null fields as well. I think the thing to do is to use GDAL's IsFieldNull when possible (2.2) and otherwise replace it with a Cython function.

snorfalorpagus:
@sgillies I now understand the approach you've taken in rasterio using the shim files. That seems the way forward to preserve the C-only source distributions. I don't mind making the changes but probably don't have time until the weekend.

PR #463, opened 2017-07-13
Support for GDAL 2.2.x and merge of ogrext1/2

snorfalorpagus:
Following the comments in #461 this is an attempt to merge ogrext1 and ogrext2 into a single file, using the shim approach taken in rasterio.


fiona/_shim.pyx
```diff
@@ -0,0 +1,55 @@
+cdef extern from "ogr_api.h":
```
snorfalorpagus: This file needs removing.

sgillies: I don't understand. Because it's taken care of elsewhere?

snorfalorpagus: Yes. I shouldn't have added it into the repo. It's a copy of either shim1, shim2 or shim22, in the same way that ogrext.pyx was a copy of ogrext1.pyx/ogrext2.pyx.

fiona/_shim.pyx
```diff
@@ -0,0 +1,55 @@
...
+from fiona.ogrext2 cimport *
+from fiona._err import cpl_errs
+
+cdef bint is_field_null(void *feature, int n):
+    if OGR_F_IsFieldNull(feature, n):
+        return True
+    elif not OGR_F_IsFieldSet(feature, n):
+        return True
+    else:
+        return False
```
sgillies: Great!

snorfalorpagus:
I've been thinking about this issue some more. If we were to do a refactor like this, how do we keep up with the maint- branches? It becomes much more difficult to merge commits as the main files have moved around.

sgillies:
@snorfalorpagus yes, it does become more difficult. The good news is that we have no more than 1 bug to fix in maint-1.7 and port to 1.8 and we don't have to maintain 1.7 once 1.8 is released since the latter is compatible.

snorfalorpagus:
@sgillies I have some more time to look at this again now. I'll apply this approach to the master branch now, if that's OK?

sgillies:
@snorfalorpagus yes please!

PR #474, opened 2017-09-04
update README to include enum34 dependency

egorf:


sgillies:
Hmm, some CI failures related to zlib, but this is still mergeable.

PR #480, opened 2017-10-22
Remove unintended output in fio cat

### LINKED ISSUE #478: fio cat --layers has unintended output
The bug is here: https://github.com/Toblerity/Fiona/blob/maint-1.7/fiona/fio/options.py#L42.

sgillies:
Resolve #478


snorfalorpagus:
The Travis error is fixed in #482.

Issue #369, opened 2016-06-30
Catch BrokenPipeError in CLI and don't log exception


geowurster:
When piping multiline outputs like feature sequences to another utility that does not fully consume the output an exception is raised and logged to `stderr`.  If its easy, handling `KeyboardInterrupt` would be nice too.  `BrokenPipeError` is ignored by default on Python 3, but Python 2 takes a bit of extra handling like:

``` python
try:
    # CLI logic
except BrokenPipeError as e:
    if sys.version_info[0] == 2 and 'broken pipe' in str(e):
        ctx.exit()  # Exit the click context
    else:
        raise
except Exception:
    logger.exception("Exception encountered ...")
    raise
```

For example:

``` console
$ fio cat tests/data/coutwildrnp.shp --precision 1 | head -1
[...]

PR #483, opened 2017-10-26
Add head option to 'fio cat'

timtroendle:
This is something rather useful to me and so I thought it might be worth sharing: It limits the number of features returned by `fio cat`. When dealing with large datasets that is helpful.

### DIFF
--- a/fiona/fio/cat.py
+++ b/fiona/fio/cat.py
@@ -33,9 +33,11 @@ warnings.simplefilter('default')
 @cligj.use_rs_opt
 @click.option('--bbox', default=None, metavar="w,s,e,n",
               help="filter for features intersecting a bounding box")
+@click.option('--head', default=None, type=click.INT,
+              help="limit each input to given amount of features")
 @click.pass_context
 def cat(ctx, files, precision, indent, compact, ignore_errors, dst_crs,
-        use_rs, bbox, layer):
+        use_rs, bbox, layer, head):
 
     """
     Concatenate and print the features of input datasets as a sequence of
@@ -73,7 +75,7 @@ def cat(ctx, files, precision, indent, compact, ignore_errors, dst_crs,
                                 bbox = tuple(map(float, bbox.split(',')))
                             except ValueError:
                                 bbox = json.loads(bbox)
-                        for i, feat in src.items(bbox=bbox):
+                        for i, feat in src.items(head, bbox=bbox):
                             if dst_crs or precision >= 0:
                                 g = transform_geom(
                                     src.crs, dst_crs, feat['geometry'],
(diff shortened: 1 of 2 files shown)


sgillies:
Thanks for sharing this @timtroendle ! I have a question about what the behavior should be if we pass multiple files to fio cat (which it supports and I use this feature regularly): do we want to get the first n features from each or first n total?

It occurs to me that we could side step that issue if we had a `fio head` command that would exit after printing n features. It could be based on `fio filter`, which already knows how to parse GeoJSON text sequences, or maybe we could add the `--head` option to fio filter.

geowurster:
@timtroendle Any reason why `$ fio cat file | head -10` doesn't work?

timtroendle:
@sgillies Agreed. I had thought about these options as well. The changes here were basically the simplest to be made to make it work for me and as they were almost no effort I don't mind throwing them away and building it in somewhere else. I just started using the cli a few days ago, so I am not in a good position to judge which place would be the best.

@geowurster You are right, that works. Well _I thought_ this approach here should be faster on large files. But I just had a second try and that does not seem to be true. Piping to `head` throws a `BrokenPipeError` at the moment for me, [but that should be easy to fix](https://stackoverflow.com/questions/26692284/brokenpipeerror-in-python#26738736).

So there are two ways I see to handle this:

1. Use linux `head` and print the first n total. Would need to fix the `BrokenPipeError`.
2. Use a build-in approach to print first n from each file; similar to the way it is now.

Both work for me, but the first looks like the cleaner way.

geowurster:
@timtroendle The `BrokenPipeError` is tracked in https://github.com/Toblerity/Fiona/issues/369, but its sent to `stderr`, so `$ fio cat file | head -10` is functional right now.  This command prints a feature sequence, redirects `stderr` to `/dev/null`, pipes to `head` to grab the first 10, and then counts the number of lines in `stdout`.  Fixing #369 is a bit more involved than this PR as it requires touching a few more commands, but if you're up for it a PR would be welcomed!

```
$ fio cat tests/data/coutwildrnp.shp 2>/dev/null | head -10 | wc -l
      10
```

Output below without the `stderr` redirect.  The output is directly below the second `PROJ_LIB` warning.

```
$ fio cat tests/data/coutwildrnp.shp | head -10 | wc -l
WARNING:Fiona:PROJ data files not located, PROJ_LIB not set
WARNING:Fiona:PROJ data files not located, PROJ_LIB not set
      10
  File "/Users/kevin.wurster/Library/Python/2.7/lib/python/site-packages/fiona/fio/cat.py", line 87, in cat
    click.echo(json.dumps(feat, **dump_kwds))
  File "/Users/kevin.wurster/Library/Python/2.7/lib/python/site-packages/click/utils.py", line 260, in echo
    file.flush()
IOError: [Errno 32] Broken pipe
Aborted!
```

sgillies:
Unix head is great, but it only works on compact GeoJSON text sequences.

timtroendle:
In either case, it's not going to be integrated in this way so I am closing this PR now. If you have a clear idea of how to build it in let me know and I might be able to do so.

Issue #429, opened 2017-03-07
Selecting a subset of fields while reading (for speed up)


oztalha:
This is a feature request. Porting a SO question: [Only read specific attribute columns of a shapefile with [...] Fiona](http://gis.stackexchange.com/questions/129414/only-read-specific-attribute-columns-of-a-shapefile-with-geopandas-fiona) to speed up reading from a shapefile.

snorfalorpagus:
This could be supported using the `SetIgnoredFields` functionality in OGR. The argument is a blacklist of fields to ignore. A whitelist (more akin to usecols in pandas) could be supported using a difference from a list of all the fields. Additionally, it's possible to ignore the special geometry field - I've had a situation before were I wasn't really interested in the geometry, just the attributes.

http://www.gdal.org/classOGRLayer.html#a5e0c3427f64249d1c35cefb487546b10

sgillies:
@oztalha @snorfalorpagus It looks like we have a choice between specifying fields to get or fields to skip. A big advantage of the latter seems to be that it can easily be a no-op when the field doesn't exist, whereas there's a decision to be made if I ask for field "foo" and it's not in the file: pass or raise an exception?

snorfalorpagus:
I agree that blacklisting would be easier to implement and is actually the interface that OGR provides. I think I'd find whitelisting more useful. Can we do both? Or is that confusing?

snorfalorpagus:
This was implemented in #488.

#469 is related.

PR #488, opened 2017-10-30
First attempt at ignoring fields

snorfalorpagus:
This PR is my first attempt at ignoring fields using the OGR API, as discussed in #429.

Is there a better way to create a `NULL` terminated `char **` array?

The feature is exposed as a keyword argument on `fiona.open`, e.g.

```
collection = fiona.open(WILDSHP, "r", ignore_fields=["AREA", "STATE"])
```

I don't think it would be to difficult to implement a `usecols` option for whitelisting, as the OGR layer is already open by the time the API method is called.

The fields are not *hidden* from the user, but their values are not returned. Instead the fields are populated with `None`. This isn't something I've decided - it's just what it's doing.

I haven't done any benchmarking on this to see if it actually makes a difference for large datasets.

### DIFF
--- a/fiona/collection.py
+++ b/fiona/collection.py
@@ -31,7 +31,8 @@ class Collection(object):
 
     def __init__(self, path, mode='r', driver=None, schema=None, crs=None,
                  encoding=None, layer=None, vsi=None, archive=None,
-                 enabled_drivers=None, crs_wkt=None, **kwargs):
+                 enabled_drivers=None, crs_wkt=None, ignore_fields=None,
+                 **kwargs):
 
         """The required ``path`` is the absolute or relative path to
         a file, such as '/data/test_uk.shp'. In ``mode`` 'r', data can
@@ -87,6 +88,7 @@ class Collection(object):
         self._crs_wkt = None
         self.env = None
         self.enabled_drivers = enabled_drivers
+        self.ignore_fields = ignore_fields
 
         self.path = vfs.vsi_path(path, vsi, archive)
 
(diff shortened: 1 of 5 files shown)


fiona/ogrext.pyx
```diff
@@ -460,6 +460,17 @@ cdef class Session:
...
+        if self.collection.ignore_fields:
+            ignore_fields = <char **>malloc((len(self.collection.ignore_fields)+1) * sizeof(char **))
+            ignore_fields_bytes = []
+            for n in range(len(self.collection.ignore_fields)):
+                ignore_fields_bytes.append(self.collection.ignore_fields[n].encode("utf-8"))
+                ignore_fields[n] = ignore_fields_bytes[n]
+            ignore_fields[n+1] = NULL  # array must be NULL-terminated
+            OGR_L_SetIgnoredFields(self.cogr_layer, ignore_fields)
+            free(ignore_fields)
+            del(ignore_fields_bytes)
```
sgillies: @snorfalorpagus `CSLAddString()` is the function to use like this:

```
cdef char **ignore_fields = NULL

for name in self.collection.ignore_fields:
    name = name.encode('utf-8')
    ignore_fields = CSLAddString(ignore_fields, <const char *>name)

try:
    OGR_L_SetIgnoredFields(self.cogr_layer, ignore_fields)
finally:
    CSLDestroy(ignore_fields)
```

sgillies:
Might be less surprising to remove the ignored fields from the schema and features entirely? Would it be hard to do? Is there any reason not to do this @snorfalorpagus ?

fiona/collection.py
```diff
@@ -87,6 +89,8 @@ def __init__(self, path, mode='r', driver=None, schema=None, crs=None,
         self._crs_wkt = None
         self.env = None
         self.enabled_drivers = enabled_drivers
+        self.ignore_fields = ignore_fields
+        self.ignore_geometry = bool(ignore_geometry)
```
sgillies: Great. Given a collection it will sometimes be good to know if fields have been ignored.

PR #490, opened 2017-11-04
Build gdal-trunk on travis (allow failure)

### LINKED ISSUE #479: Add GDAL 2.2 and trunk (allowed to fail) to the testing matrix
And remove 1.9 (as in Rasterio).

snorfalorpagus:
Also dropped build of 1.9.2 (released 2012). Oldest build is now 1.11.2 (released 2015).

There are a few failures against trunk we should look into.

Closes #479.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ cache:
   directories:
     - $GDALINST
     - ~/.cache/pip
+
 env:
   global:
     - PIP_WHEEL_DIR=$HOME/.cache/pip/wheels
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -26,29 +32,50 @@ addons:
...
-  - "pip install wheel coveralls>=1.1 --upgrade"
-  - pip install setuptools==36.0.1
-  - pip install wheel
-  - . ./scripts/travis_gdal_install.sh
-  - export PATH=$GDALINST/gdal-$GDALVERSION/bin:$PATH
-  - export LD_LIBRARY_PATH=$GDALINST/gdal-$GDALVERSION/lib:$LD_LIBRARY_PATH
-  - export GDAL_DATA=$GDALINST/gdal-$GDALVERSION/share/gdal
-  - export PROJ_LIB=/usr/share/proj
-  - gdal-config --version
+  - |
```
sgillies: I've never used this bit of YAML syntax before. Thanks for introducing me to it @snorfalorpagus ! All the same, I think I'd like to keep these commands as discrete steps for the present.

Issue #476, opened 2017-09-25
Writing to GeoPackage is very slow


culebron:
Working in GeoPandas I found an issue with GeoPackage format. It seems that it's both GeoPandas' and lower level's fault. Can performance be improved at Fiona's level?

I have SSD. OS is Ubuntu 17.04, Fiona version 1.7.9.post1. I installed it as recommended in #400 :

    sudo pip3 install -I fiona --no-binary fiona

Here's my test suite for a proof: https://github.com/culebron/geodata  Run `python3.6 few.py` and `python3.6 multiple.py` to compare. `few.py` opens a file with a lot of data, but only 2.7K records as GeoDataFrame. It writes them into GeoJSON and GPKG. In this case, GPKG driver outperforms GeoJSON. `multiple.py` creates a 100K records dataframe and then saves it to GeoJSON and GPKG. Here, GPKG is incredibly slow.

My results with default GeoPandas:

	$ python3.6 few.py 
	writing 2.7K records to geojson 36.283805477003625
	writing 2.7K records to gpkg    20.792497718997765
	$ python3.6 multiple.py 
	100%|████████████████████████████████████████████████████████| 100000/100000 [00:03<00:00, 29996.25it/s]
	writing 100K records to geojson 61.62079200500011
	writing 100K records to gpkg    260.4413645050008

And notice that in case of `multiple.py`, the resulting GeoPackage file is only 9 megs. Which is times smaller than the file produced by `few.py`.

As I understand, the problem is that Fiona opens a session in Sqlite and creates a lock file, and it takes some time. And inspecting the code, I see GeoPandas naively writes everything 1 record at a time, which means Sqlite honestly locks it, then writes, then unlocks:

https://github.com/geopandas/geopandas/blob/master/geopandas/io/file.py#L107

    with fiona.drivers():
        with fiona.open(filename, 'w', driver=driver, crs=df.crs,
                        schema=schema, **kwargs) as colxn:
            for feature in df.iterfeatures():
                colxn.write(feature)

I tried batching records:

    with fiona.drivers():
        with fiona.open(filename, 'w', driver=driver, crs=df.crs,
                        schema=schema, **kwargs) as colxn:
            buf = []
            for feature in df.iterfeatures():
                buf.append(feature)
                if len(buf) > 9999:
                    colxn.writerecords(buf)
                    buf = []

            colxn.writerecords(buf)

    $ python3.6 multiple.py 
[...]

PR #491, opened 2017-11-04
Initial implementation of transactions in writerecords

snorfalorpagus:
This PR implements transactions when using `collection.writerecords` in GDAL 2.x, as discussed in #476. Currently it's hard-coded to use a transaction size of 20,000 (the same as `ogr2ogr`). This significantly speeds up writing GPKG files.

I'm not sure of the best way to expose the transaction size to the user. Is it a property of the collection, or an argument to the `writerecords` call?

This should have no impact on drivers which don't support transactions.

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -24,7 +24,8 @@ from fiona._err import cpl_errs
 from fiona._geometry import GEOMETRY_TYPES
 from fiona import compat
 from fiona.errors import (
-    DriverError, DriverIOError, SchemaError, CRSError, FionaValueError)
+    DriverError, DriverIOError, SchemaError, CRSError, FionaValueError,
+    TransactionError)
 from fiona.compat import OrderedDict
 from fiona.rfc3339 import parse_date, parse_datetime, parse_time
 from fiona.rfc3339 import FionaDateType, FionaDateTimeType, FionaTimeType
@@ -70,6 +71,8 @@ FIELD_TYPES_MAP = {
     'datetime': FionaDateTimeType,
     'bytes':    bytes,
    }
+
+DEFAULT_TRANSACTION_SIZE = 20000
 
 # OGR Driver capability
 cdef const char * ODrCCreateDataSource = "CreateDataSource"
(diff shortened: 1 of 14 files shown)


fiona/ogrext.pyx
```diff
@@ -1017,6 +1022,20 @@ cdef class WritingSession(Session):
                 raise RuntimeError("Failed to write record: %s" % record)
             _deleteOgrFeature(cogr_feature)
 
+            features_in_transaction += 1
+            if features_in_transaction == 20000:  # TODO: this shouldn't be hard-coded
```
sgillies: @snorfalorpagus how about this for a compromise: until we make transactions a Fiona software feature, say in 1.8, let's use a constant value here. `DEFAULT_TRANSACTION_SIZE = 20000` or something like that defined in the module. Then we'd have the same performance characteristics as ogr2ogr in many cases and users would have two work-arounds for different transaction sizes:

1. a user could monkey patch the constant in code `fiona.ogrext.DEFAULT_TRANSACTION_SIZE = 1000`.
2. better: a user could call `writerecords()` with batches smaller than 20,000 and each of these would be done in their own transaction, right?

snorfalorpagus:
I've added a test for this. It uses a custom logging handler which catches the debug messages in `writerecords` to test if the transactions are working. I've not done anything like this before so I'm not sure if it's sane.

Also, Travis isn't configured to build GDAL with GPKG support. Is there a reason for this? I had trouble enabling it.

sgillies:
Seems sane to me. When I did this, I used a temporary file handler, but your custom handler looks like something I might try next time.

We've omitted GPKG support only because I'm not very experienced with it and it seemed time consuming to get it right.

I'm excited to get this merged. Let me know when you're done.

snorfalorpagus:
I think this is ready for merge now. I've improved the geopackage tests/fixture a little. I've also added a header to pytest with the supported drivers which makes it easier to see if GPKG support is enabled.

tests/conftest.py
```diff
@@ -90,6 +94,25 @@ def path_coutwildrnp_json():
...
+    file path."""
+    has_gpkg = "GPKG" in fiona.supported_drivers.keys()
+    if not has_gpkg:
+        raise RuntimeError("GDAL has not been compiled with GPKG support")
+    path = os.path.join(data_dir(), 'coutwildrnp.gpkg')
+    if not os.path.exists(path):
+        filename_shp = _COUTWILDRNP_FILES[0]
+        path_shp = os.path.join(data_dir(), filename_shp)
+        with fiona.open(path_shp, "r") as src:
+            meta = copy.deepcopy(src.meta)
```
sgillies: The `meta` property is a new dict and changes to it don't affect the src, we don't need to copy it here.

sgillies: I'll update after the merge.

snorfalorpagus: Good to know. I'm not sure where I picked this habit up from.

Issue #500, opened 2017-11-11
Copy MemoryFile and ZipMemoryFile from Rasterio


sgillies:
See https://github.com/mapbox/rasterio/blob/master/rasterio/io.py. Fiona's `BytesCollection` would be deprecated in a future version of Fiona.

sgillies:
Done!

PR #501, opened 2017-11-13
Initial impl of MemoryFile for Fiona

sgillies:
Towards resolution of #500.

Reading bytes from a `MemoryFile` is not yet working.


sgillies:
In Rasterio, `MemoryFile` uses numpy to get the size of the buffer written to the memory file by rasterio. We want to avoid, for now, requiring numpy for Fiona and will need a different way.

snorfalorpagus:
Looks like you forgot to add `fiona.io` (and maybe `fiona._io`?) to the commit.

sgillies:
Woo hoo, it's working 🎉

snorfalorpagus:
This looks good to me. I think it would be useful to add some examples in the tests for using this with a `BytesIO` and a Python file object (from regular `open()`).

Issue #510, opened 2017-11-27
Fiona > 1.7.5 does't open my shapefile - two different errors


miker985:
I've run in to issues using fiona that I suspect are hitting some edge-case somewhere. I'm not clear if this is the shapefile's issue and only newer versions of fiona are hitting them, or if this is a problem with one of the recent commits to fiona.

## Expected behavior and actual behavior.

My `test-fiona.py` script (run on python 2.7 and 3.6)
```python
#!/usr/bin/env python
from __future__ import print_function
import fiona
print(fiona.__version__)
# J is a shared filesystem I have symlink'd for convenience
collection = fiona.open('/Users/miker985/J/WORK/11_geospatial/05_survey shapefile library/Shapefile directory/IDN_DHS_1994.shp')
print("\tOpened")
```

My test script
 
```bash
pip install fiona==1.7.0
./test-fiona.py

for MINOR_VER in {1..10}; do
    pip install --upgrade fiona==1.7.${MINOR_VER} &>/dev/null
    ./test-fiona.py
done
```

For both pythons 2.7 and 3.6:
* 1.7.5 and below `fiona.open` succeeds and I get a collection object back.
* 1.7.6 and 1.7.7 I get `TypeError: __init__() takes exactly 4 positional arguments (2 given)`
* 1.7.8 and newer I get `fiona._err.CPLE_AppDefinedError: b'Recode from ANSI 1252 to UTF-8 failed with the error: "Invalid argument".'`
    * I have tried adding an `encoding=` keyword-arg to `fiona.open` and see no change. Specifically - I see the same error message regardless of the encoding (even if the encoding is invalid e.g., `encoding='WAT'`)

## Steps to reproduce the problem.

I've included the above scripts as an example. I'm still trying to get permission to share `IDN_DHS_1994.shp`, but as it's a shapefile distributed by DHS I'm not clear if this is possible.

## Operating system

Mac OS X 10.12.5.

## Fiona version and provenance

As stated above, I've tested fiona 1.7.0 thru 1.7.10. If you need the specific dist that pip is installing I can get you all 10 - just let me know.

jdmcbr:
@miker985 This looks similar to an issue @ozak observed in #486. @ozak are you still seeing this issue? And, if so, do you have a shapefile you can share that shows the issue?

ozak:
[...]

PR #511, opened 2017-12-10
Remove unnecessary if/else

jdmcbr:
Though in looking at the blame, this was just edited/reviewed, so
perhaps I'm just missing some black magic beyond my understanding.

This caught my eye while trying to see if I could fathom any explanation for the behavior I mentioned in #510 regarding different behavior in opening a file using `encoding='utf-8'`.


jdmcbr:
@sgillies Sounds good. It obviously isn't causing any issues as is, it was just a little distracting/confusing for me.

PR #516, opened 2017-12-13
WIP: fix open/session encoding keyword argument

### LINKED ISSUE #510: Fiona > 1.7.5 does't open my shapefile - two different errors

sgillies:
This PR implements layer opening and creation options and adds a test for overriding of shapefile encoding.

Resolves #510.


fiona/_shim1.pyx
```diff
@@ -1,25 +1,41 @@
...
 
+
 cdef void gdal_flush_cache(void *cogr_ds):
     retval = OGR_DS_SyncToDisk(cogr_ds)
     if retval != OGRERR_NONE:
         raise RuntimeError("Failed to sync to disk")
 
-cdef void* gdal_open_vector(const char *path_c, int mode, drivers):
+
+cdef void* gdal_open_vector(const char *path_c, int mode, drivers, options):
```
sgillies: I'm extending two shim functions with a new positional argument: a Python dict of the creation or open options.

fiona/_shim1.pyx
```diff
@@ -37,15 +53,36 @@ cdef void* gdal_open_vector(const char *path_c, int mode, drivers):
...
         cogr_ds = exc_wrap_pointer(
             OGR_Dr_CreateDataSource(
-                cogr_driver, path_c, NULL))
+                cogr_driver, path_c, opts))
+        return cogr_ds
     except Exception as exc:
         raise DriverIOError(str(exc))
-    return cogr_ds
+    finally:
+        CSLDestroy(opts)
```
sgillies: @snorfalorpagus this work benefits enormously from your rewrite! In maint-1.7 I had to sprinkle `CSLDestroy(opts)` throughout because there were so many places we could exit.

fiona/ogrext.pyx
```diff
@@ -395,7 +395,7 @@ cdef class Session:
     def __dealloc__(self):
         self.stop()
 
-    def start(self, collection):
+    def start(self, collection, **kwargs):
```
sgillies: Here's where open options come in now.

fiona/ogrext.pyx
```diff
@@ -895,6 +883,15 @@ cdef class WritingSession(Session):
...
 
+            for k, v in kwargs.items():
+                k = k.upper().encode('utf-8')
+                if isinstance(v, bool):
+                    v = ('ON' if v else 'OFF').encode('utf-8')
+                else:
+                    v = str(v).encode('utf-8')
+                log.debug("Set option %r: %r", k, v)
+                options = CSLAddNameValue(options, <const char *>k, <const char *>v)
+
```
sgillies: We don't have a shim for `GDALDatasetCreateLayer`, so the option encoding goes here.

geowurster:
@sgillies I won't be able to get to this until next week at the earliest.

fiona/_shim2.pyx
```diff
@@ -27,23 +38,44 @@ cdef void* gdal_open_vector(char* path_c, int mode, drivers):
...
 
+    for k, v in options.items():
+        k = k.upper().encode('utf-8')
+        if isinstance(v, bool):
+            v = ('ON' if v else 'OFF').encode('utf-8')
+        else:
+            v = str(v).encode('utf-8')
+        log.debug("Set option %r: %r", k, v)
+        open_opts = CSLAddNameValue(open_opts, <const char *>k, <const char *>v)
+
```
snorfalorpagus: This section should be pulled out into a function and reused in `gdal_create`. I don't think there is a nice way to share it between shim2/shim22, but I still think it should come out.

tests/test_encoding.py
```diff
@@ -0,0 +1,56 @@
...
+import pytest
+
+import fiona
+
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+# Uncomment when merged to master, where we're using pytest.
+# We're still holding our noses in maint-1.7.
```
snorfalorpagus: You're planning to cherry-pick this into maint-1.7?

snorfalorpagus:
LGTM.

fiona/_shim2.pyx
sgillies: @snorfalorpagus I think I'm going to follow the rule of 3 here: it's okay to duplicate code 2x, but on the third time we refactor.

tests/test_encoding.py
sgillies: Not a cherry pick, but a separate PR.

sgillies:
BTW, complaints about `AttributeError: 'Collection' object has no attribute 'session'` come from `Collection` instances in tests that are being garbage collected after the class has already been torn down. If you stick to  `with fiona.open(...) ` in your code, you will never see this error.

sgillies:
@geowurster No problem! I'm going to be checking here only sporadically until 1/2.

PR #519, opened 2017-12-14
[WIP] Use C++11. Closes #514.

### LINKED ISSUE #514: Support C++11
Relates to https://github.com/mapbox/rasterio/issues/1230 and https://github.com/sgillies/frs-wheel-builds/issues/25

GDAL 2.3 will [require C++11](https://trac.osgeo.org/gdal/changeset/40984).  I don't think a Fiona build has been triggered on Travis since the change, but [Rasterio's GDAL trunk builds are failing with a bunch of warnings](https://travis-ci.org/mapbox/rasterio/jobs/315171560) like:

```
/home/travis/gdalinstall/gdal-trunk/include/ogr_geometry.h:974:9: warning: override controls (override/final) only available with -std=c++11 or -std=gnu++11 [enabled by default]
             const CPL_OVERRIDE;
             ^
```

The [Travis builds against GDAL trunk](https://travis-ci.org/Toblerity/Fiona/jobs/316454091) are currently failing with the same issues.

snorfalorpagus:
This PR builds the `_transform` module using the C++11 standard.

I've not tested this on Windows. I suspect it will break, as MSVC uses a different flag? `/std:c++11`?

Also, VS2008 (which must be used to build Python 2.7 modules) does not support C++11. If I understand this correctly, this means a Python 2.7 + GDAL 2.3 build will be impossible. If we include this, it will also mean the .cpp source distribution wont work for Python 2.x. https://stackoverflow.com/questions/20302891/visual-studio-2008-with-c11

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -180,6 +180,11 @@ ext_options = dict(
     libraries=libraries,
     extra_link_args=extra_link_args)
 
+ext_options_cpp = ext_options.copy()
+# GDAL 2.3+ requires C++11
+ext_options_cpp["extra_compile_args"] = ["-std=c++11"]
+
+
 # Define the extension modules.
 ext_modules = []
 
@@ -206,7 +211,7 @@ if source_is_repo and "clean" not in sys.argv:
 
     ext_modules = cythonize([
         Extension('fiona._geometry', ['fiona/_geometry.pyx'], **ext_options),
-        Extension('fiona._transform', ['fiona/_transform.pyx'], **ext_options),
+        Extension('fiona._transform', ['fiona/_transform.pyx'], **ext_options_cpp),
         Extension('fiona._crs', ['fiona/_crs.pyx'], **ext_options),
         Extension('fiona._drivers', ['fiona/_drivers.pyx'], **ext_options),
         Extension('fiona._err', ['fiona/_err.pyx'], **ext_options),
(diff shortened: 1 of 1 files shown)


snorfalorpagus:
Apparently MSVC just ignores arguments it doesn't understand...
```
cl : Command line warning D9002 : ignoring unknown option '-std=c++11'
```

snorfalorpagus:
https://stackoverflow.com/questions/19372151/what-is-vs2012-cl-exe-equivalent-to-gcc-std-c11

>Is there a command line option I can pass to VS2012 or VS2013 cl.exe to specify if I want to use C++11 or C++98 syntax? Like GCC's -std=c++11 option.

> Not to my knowledge. VS always recognizes C++11 syntax; well, the parts it supports, at least.

sgillies:
I'm rebuilding this PR on Appveyor since the drive name issue is fixed.

sgillies:
The single test failure is related to this log message: `ERROR    CPLE_AppDefined in b"Failed to create field name 'num' : cannot convert to gb18030`.

Until I pushed the commit that enabled encoding, we weren't truly encoding shapefiles, and that test passed incorrectly. I wonder if our GDAL library builds lack iconv support?

snorfalorpagus:
> I wonder if our GDAL library builds lack iconv support?

There is a `iconv.lib` in the files downloaded which suggests it does, although I've not check if it was actually linked.

snorfalorpagus:
I'm getting the same exception on macOS with current master branch, so this is not specific to appveyor/windows.

screen shot 2017-12-27 at 12 19 11 pm

snorfalorpagus:
I'm using the libiconv from conda-forge. I can see this in the build log:

```
checking for iconv... yes
checking for working iconv... no
checking for iconv declaration... 
         extern size_t iconv (iconv_t cd, char * *inbuf, size_t *inbytesleft, char * *outbuf, size_t *outbytesleft);
```

I don't know if gisinternals (where appveyor binaries are from) have the same problem. I don't think it's related to the C++11 issue though. I'll create a separate issue for this.

PR #520, opened 2017-12-16
Fix handling of paths on Windows. Closes #505.

### LINKED ISSUE #505: VFS changes broke build on Windows
Some changes in #426 broke the build on Windows. I can reproduce this locally. https://ci.appveyor.com/project/sgillies/fiona

The error occurs when trying to open any file:

```
        if vsi:
            if not isinstance(vsi, string_types) or not vfs.valid_vsi(vsi):
>               raise TypeError("invalid vsi: %r" % vsi)
E               TypeError: invalid vsi: 'c'
fiona\collection.py:69: TypeError
```

I think the problem might be in `parse_paths` and the handling of Windows-style paths.

snorfalorpagus:
This PR fixes a regression introduced in the handling of paths beginning with a drive letter on Windows. Previously a path like `"C:\hello\world.shp"` would be interpreted as a URL with scheme `"C"` and a path `"/hello/world"`, which is incorrect. This type of path is now detected on Windows and handled correctly. True VFS paths should still work as expected.

Closes #505.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,12 +21,14 @@ environment:
         - PYTHON: "C:\\Python34-x64"
           PYTHON_VERSION: "3.4.3"
           PYTHON_ARCH: "64"
+          GDAL_VERSION: "1.11.4"
           GIS_INTERNALS: "release-1600-x64-gdal-1-11-4-mapserver-6-4-3.zip"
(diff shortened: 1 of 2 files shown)


snorfalorpagus:
Tagged another fix onto this PR which fixes a bug in the declaration of which GDAL version to build against on Appveyor.

PR #527, opened 2017-12-31
Fixed WGS 84 CRS regular expression.

snorfalorpagus:
This PR corrects a mistake in #495. The regex wasn't actually matching the new CRS in GDAL 2.3 - there are other issues still causing tests to fail in trunk, so it's difficult to spot.


sgillies:
🙏 @snorfalorpagus

Issue #528, opened 2018-01-04
Number formatting is not consistent among drivers


galak75:
According to the [documentation](http://toblerity.org/fiona/manual.html#format-drivers-crs-bounds-and-schema), a schema can be provided when writing records to a file.

e.g.:
```
    def _write_collection(self, driver):
        writing_schema = {
            'geometry': 'Point',
            'properties': [('integer', 'int'), ('one_digit', 'float:15.1')]
        }

        with fiona.open(
                self.filename,
                "w",
                driver=driver,
                schema=writing_schema,
                crs='epsg:4326',
                encoding='utf-8'
        ) as c:
            c.writerecords([
                {
                    'geometry': {'type': 'Point', 'coordinates': (0.0, 0.1)},
                    'properties': {'integer': 12.22, 'one_digit': 12.22}
                },
                {
                    'geometry': {'type': 'Point', 'coordinates': (0.0, 0.2)},
                    'properties': {'integer': 12.88, 'one_digit': 12.88}
                }
            ])
```

Unfortunately, all drivers do not behave the same:

- `ESRI Shapefile` truncates values to get an `int`, and rounds values to limit digits numbers (`float:15.1`)
- `GeoJSON` also truncates values to get an an `int`, but does not respect float formatting (`float:15.1`)

2 questions : 

1. Is it possible to make `GeoJSON` driver respect float formatting?
1. Shouldn't we have a common behavior for `int` and `float:15.1` types? (either rounding or truncating)

sgillies:
@galak75 The float formatting is done by GDAL/OGR and I'm inclined to let it continue as it is. As you've noticed, support for field widths and decimal precision is uneven in GDAL/OGR. As far as I can tell, the OGR concepts of field width and precision derive more from the designs of older spatial data formats than from the desire to have first class support for decimals (like Python's decimal modules).

The integer conversion is done by Fiona and only exists to trigger exceptions when we mistakenly try to copy non-numeric values to numeric fields, so that we don't get an exception deeper in OGR. Calling `int(round())` instead of `int()` goes too far for me. 

Fiona schemas aren't intended to validate and coerce your data into conformation. Their only purpose is to tell OGR how to format already-conforming data appropriately for a specific spatial data format. It only works at the OGR <-> data file interface.

[...]

PR #529, opened 2018-01-04
Add some unit tests on number formatting (#528)

galak75:
As discussed in #528

### DIFF
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -407,12 +407,12 @@ class PropertiesNumberFormattingTest(unittest.TestCase):
         )
 
         with fiona.open(self.filename, driver=driver, encoding='utf-8') as c:
-            self.assertEqual(len(c), 2)
+            self.assertEqual(2, len(c))
 
             rf1, rf2 = list(c)
 
-            self.assertEqual(rf1['properties']['property1'], 12)
-            self.assertEqual(rf2['properties']['property1'], 12)
+            self.assertEqual(12, rf1['properties']['property1'])
+            self.assertEqual(12, rf2['properties']['property1'])
 
     def test_shape_driver_rounds_float_property_to_requested_digits_number(self):
         driver = "ESRI Shapefile"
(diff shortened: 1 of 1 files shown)


tests/test_collection.py
```diff
@@ -349,6 +349,76 @@ def test_no_filter(self):
...
+        with fiona.open(self.filename, driver=driver, encoding='utf-8') as c:
+            self.assertEqual(len(c), 2)
+
+            rf1, rf2 = list(c)
+
+            # integers are truncated
+            self.assertEqual(rf1['properties']['integer'], 12)
+            self.assertEqual(rf2['properties']['integer'], 12)
+
+            # floats are rounded
```
sgillies: @galak75 would you be willing to make the float assertions into a separate test function? The truncation and rounding effects have different causes.

tests/test_collection.py
```diff
@@ -349,6 +349,76 @@ def test_no_filter(self):
...
+            self.assertEqual(len(c), 2)
+
+            rf1, rf2 = list(c)
+
+            # integers are truncated
+            self.assertEqual(rf1['properties']['integer'], 12)
+            self.assertEqual(rf2['properties']['integer'], 12)
+
+            # ****************************************
+            # FLOAT FORMATTING IS NOT RESPECTED...
```
sgillies: Here we should also have a different test since there is a split between expected and unexpected results.

galak75:
@sgillies thank you for your comments. I split assertions in separate tests. Feel free to suggest different test function names (English is not my mother tongue), or additional changes.

Also, I added some new tests, based on [your comment](https://github.com/Toblerity/Fiona/issues/528#issuecomment-355327527) : 

> The integer conversion is done by Fiona and only exists to trigger exceptions when we mistakenly try to copy non-numeric values to numeric fields

- `"12.22"` is converted to `12.22` float value before being written by the target driver
- `"invalid number"` is converted to `0`

But there is actually no exception (which might be the correct behavior)

These new tests might be not necessary if already tested else where. I can remove them if you prefer. Just let me know.

sgillies:
Thanks @galak75 . I'm fixing up the Windows builds and will see about merging this after AppVeyor gives us the green light.

tests/test_collection.py
```diff
@@ -392,15 +392,22 @@ def test_shape_file_properties_number_formatting(self):
...
-            # floats are rounded
+    def test_shape_driver_rounds_float_property_to_requested_digits_number(self):
+        driver = "ESRI Shapefile"
+        self._write_collection(driver)
+
+        with fiona.open(self.filename, driver=driver, encoding='utf-8') as c:
+            self.assertEqual(len(c), 2)
+
+            rf1, rf2 = list(c)
+
```
sgillies: Excellent !

galak75:
Glad to see that ! these are only tests, but it is my first contribution to an open source project. :smiley:

PR #531, opened 2018-01-06
Added support for bool field subtype in GDAL 2.x

### LINKED ISSUE #524: cannot specify boolean type in schema
JSON and GeoJSON seem to know the type 'boolean' with values true and false.
However, when specifying a schema from scratch I cannot specify a boolean. 'bool', 'Boolean' and 'boolean' are not recognized.

in pythonanywhere.com, python 3.6:
```
from shapely.geometry import mapping, Polygon, shape
import fiona

filename = '/home/raalst/output/buurtparts.shp'

# Define a GeoJSON polygon feature geometry with attributes
schema = {
    'geometry': 'Polygon',
    'properties': {'inw': 'int', 'urban': 'boolean'}
}

def create_output_record(vorm, inwoneraantal, urban):
    if vorm.is_valid and (vorm.geom_type == 'Polygon' or vorm.geom_type == 'MultiPolygon') and inwoneraantal >= 0 and urban in [True,False]:
        return {'geometry' : mapping(vorm), 'properties' : {'inw' : inwoneraantal, 'urban': urban}}
    else:
        print("error in creating record from shape {} and attribute {}. record with empty polygon returned".format(vorm.wkt,inwoneraantal))
        return {'geometry' : mapping(Polygon()), 'properties' : {'inw' : 0, 'urban': False}}

def write_shapes(features, crs):
    """Write a new Shapefile using constant filename"""
    with fiona.open(filename, 'w', driver='ESRI Shapefile', crs=crs , schema=schema) as c:
        for f in features:
            #print(f)
            c.write(f)

def main():
    """test write_shapes() on testfile"""
    write_shapes([create_output_record(Polygon([(0, 0), (1, 1), (1, 0)]), 13,  True),
                  create_output_record(Polygon([(0, 1), (1, 1), (1, 0)]),  0, False)
                 ],
                 {'proj': 'sterea', 'lat_0': 52.15616055555555, 'lon_0': 5.38763888888889,
                 'k': 0.9999079, 'x_0': 155000, 'y_0': 463000, 'ellps': 'bessel', 'units': 'm', 'no_defs': True
                 }
                )


if __name__ == '__main__':
    main()
```
results in 

  File "/home/raalst/IGAssignment2/write_shape.py", line 38, in <module>
    main()
  File "/home/raalst/IGAssignment2/write_shape.py", line 32, in main
    'k': 0.9999079, 'x_0': 155000, 'y_0': 463000, 'ellps': 'bessel', 'units': 'm', 'no_defs': True
  File "/home/raalst/IGAssignment2/write_shape.py", line 21, in write_shapes
    with fiona.open(filename, 'w', driver='ESRI Shapefile', crs=crs , schema=schema) as c:
  File "/usr/local/lib/python3.6/dist-packages/fiona/__init__.py", line 179, in open
    enabled_drivers=enabled_drivers, crs_wkt=crs_wkt)
  File "/usr/local/lib/python3.6/dist-packages/fiona/collection.py", line 158, in __init__
    self.session.start(self, **kwargs)
  File "fiona/ogrext.pyx", line 964, in fiona.ogrext.WritingSession.start (fiona/ogrext2.c:16842)
ValueError: 'boolean' is not in list 

I was unable to determine if a boolean was or was not supported by the underlying OGR C code. it is in JSON and GeoJSON I believe (after a mild googling)
- either add to the docs the fact that booleans should be recoded as ints or strings
- or add a boolean variable to the example of a schema in the docs

'Linux-4.4.0-101-generic-x86_64-with-Ubuntu-14.04-trusty'
>>> 
'1.7.4'

snorfalorpagus:
This PR adds support for the "bool" field subtype when using GDAL 2.x. When using GDAL 1.x the subtype is simply ignored - you can still pass "bool" in the schema but an "int" will be written.

Closes #524.

### DIFF
--- a/fiona/_shim1.pxd
+++ b/fiona/_shim1.pxd
@@ -1,5 +1,12 @@
 include "ogrext1.pxd"
 
+ctypedef enum OGRFieldSubType:
+    OFSTNone = 0
+    OFSTBoolean = 1
+    OFSTInt16 = 2
+    OFSTFloat32 = 3
+    OFSTMaxSubType = 3
+
 cdef bint is_field_null(void *feature, int n)
 cdef void gdal_flush_cache(void *cogr_ds)
 cdef void* gdal_open_vector(char* path_c, int mode, drivers, options)
@@ -7,6 +14,8 @@ cdef void* gdal_create(void* cogr_driver, const char *path_c, options) except *
 cdef OGRErr gdal_start_transaction(void *cogr_ds, int force)
 cdef OGRErr gdal_commit_transaction(void *cogr_ds)
 cdef OGRErr gdal_rollback_transaction(void *cogr_ds)
+cdef OGRFieldSubType get_field_subtype(void *fielddefn)
+cdef void set_field_subtype(void *fielddefn, OGRFieldSubType subtype)
 
 from fiona._shim cimport OGR_F_GetFieldAsInteger as OGR_F_GetFieldAsInteger64
 from fiona._shim cimport OGR_F_SetFieldInteger as OGR_F_SetFieldInteger64
--- a/fiona/_shim1.pyx
+++ b/fiona/_shim1.pyx
@@ -91,3 +91,9 @@ cdef OGRErr gdal_commit_transaction(void* cogr_ds):
     return OGRERR_NONE
 cdef OGRErr gdal_rollback_transaction(void* cogr_ds):
     return OGRERR_NONE
+
+# field subtypes are not supported in GDAL 1.x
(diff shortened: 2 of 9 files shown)


sgillies:
Looks good to me so far @snorfalorpagus. Seems to be a problem with the 1.x shim, though.

snorfalorpagus:
> Seems to be a problem with the 1.x shim, though.

Yeah. Looks like it needs a little more work. I'll keep trying.

snorfalorpagus:
I think this is working now.

PR #538, opened 2018-01-20
Added `fio rm` command

### LINKED ISSUE #401: Expose OGR_Dr_DeleteDataSource() / GDALDeleteDataset()
Maybe something like `$ fio rm` and `fiona.rm()` or `fiona.delete()`.  I think prompting the user for confirmation by default for `$ fio rm` would be responsible if we provide a `--force` or `--yes` flag for scripting purposes.

This would be useful for general data management and dealing with drivers that do not support update, like GeoJSON.  Currently, running this twice produces an exception.:

```python
import fiona as fio

profile = {
    'crs': 'EPSG:4326',
    'schema': {
        'geometry': 'Point',
        'properties': {}
    },
    'driver': 'GeoJSON'
}


with fio.open('test.geojson', 'w', **profile) as dst:
    dst.write({
        'type': 'Feature',
        'properties': {},
        'geometry': {
            'type': 'Point',
            'coordinates': [0, 0]
        }
    })
```

Dropping this above `fio.open()` would alleviate the problem:

```python
fio.rm('test.geojson')
```

snorfalorpagus:
This PR adds the `rm` command to `fio`, which can be used to delete a datasource or a specific layer in a multi-layer datasource.

```
$ fio rm countries.shp
$ fio rm --layer forests land_cover.gpkg
```

The `fiona.remove` function has been extended to support a `layer` keyword argument.

Closes #401.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -10,23 +10,27 @@ Fiona's new command line interface is a program named "fio".
       Fiona command line interface.
 
     Options:
-      -v, --verbose  Increase verbosity.
-      -q, --quiet    Decrease verbosity.
-      --version      Show the version and exit.
-      --help         Show this message and exit.
+      -v, --verbose     Increase verbosity.
+      -q, --quiet       Decrease verbosity.
+      --version         Show the version and exit.
+      --gdal-version    Show the version and exit.
+      --python-version  Show the version and exit.
+      --help            Show this message and exit.
 
     Commands:
       bounds   Print the extent of GeoJSON objects
-      buffer   Buffer geometries on all sides by a fixed distance.
+      calc     Calculate GeoJSON property by Python expression
       cat      Concatenate and print the features of datasets
       collect  Collect a sequence of features.
-      distrib  Distribute features from a collection
+      distrib  Distribute features from a collection.
       dump     Dump a dataset to GeoJSON.
       env      Print information about the fio environment.
-      filter   Filter GeoJSON features by python expression
+      filter   Filter GeoJSON features by python expression.
       info     Print information about a dataset.
       insp     Open a dataset and start an interpreter.
       load     Load GeoJSON to a dataset in another format.
+      ls       List layers in a datasource.
+      rm       Remove a datasource or an individual layer.
 
 It is developed using the ``click`` package and is new in 1.1.6.
 
@@ -259,6 +263,19 @@ For example
 
 Would create a geojson file with only those features from `data.shp` where the
 area was over a given threshold.
+
+rm
+--
+The ``fio rm`` command deletes an entire datasource or a single layer in a
+multi-layer datasource. If the datasource is composed of multiple files
+(e.g. an ESRI Shapefile) all of the files will be removed.
+
+.. code-block:: console
+
+    $ fio rm countries.shp
+    $ fio rm --layer forests land_cover.gpkg
+
+New in 1.8.0.
 
 Coordinate Reference System Transformations
 -------------------------------------------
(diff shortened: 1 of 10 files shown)


docs/cli.rst
```diff
@@ -260,6 +264,19 @@ For example
 Would create a geojson file with only those features from `data.shp` where the
 area was over a given threshold.
 
+rm
+--
+The ``fio rm`` command deletes an entire datasource or a single layer in a
+multi-layer datasource. If the datasource is composed of multiple files
+(e.g. an ESRI Shapefile) all of the files will be removed.
```
sgillies: 👍

fiona/fio/rm.py
```diff
@@ -0,0 +1,25 @@
...
+@click.option("--yes", is_flag=True)
+@click.pass_context
+def rm(ctx, input, layer, yes):
+    if layer is None:
+        kind = "datasource"
+    else:
+        kind = "layer"
+
+    if not yes:
+        click.confirm("The {} will be removed. Are you use?".format(kind), abort=True)
```
sgillies: There's a typo at "Are you use"

fiona/ogrext.pyx
```diff
@@ -1270,12 +1270,20 @@ def _remove(path, driver=None):
...
+    cdef void *cogr_ds
     cdef int result
+    cdef char *driver_c
 
     if driver is None:
-        driver = 'ESRI Shapefile'
+        # attempt to identify the driver by opening the dataset
+        cogr_ds = gdal_open_vector(path.encode("utf-8"), 0, None, {})
+        if cogr_ds == NULL:
+            raise IOError("Failed to remove data source {}".format(path))
```
sgillies: How about using `exc_wrap_pointer` here?

fiona/ogrext.pyx
```diff
@@ -1287,6 +1295,37 @@ def _remove(path, driver=None):
...
+    
+    if layer_index < 0:
+        layer_names = _listlayers(path)
+        layer_index = len(layer_names) + layer_index
+    
+    cogr_ds = gdal_open_vector(path.encode("utf-8"), 1, None, {})
+    if cogr_ds == NULL:
+        raise IOError("Failed to open data source: {}".format(path))
+
+    result = OGR_DS_DeleteLayer(cogr_ds, layer_index)
```
sgillies: Can we use `exc_wrap_int` here?

fiona/ogrext.pyx
```diff
@@ -1287,6 +1295,37 @@ def _remove(path, driver=None):
         raise RuntimeError("Failed to remove data source {}".format(path))
```
sgillies: Instead of `RuntimeError` let's raise a very specific error about data source deletion. We'll need a new exception class derived from `fiona.error.FionaError`. I'm not sure what the best name will be.

fiona/ogrext.pyx
snorfalorpagus: I tried this but weirdly it didn't work as expected. In the `test_remove_nonexistent` the call to open returns `NULL`, but `CPLGetLastErrorType()` returns 0. This results in a message `SystemError: <built-in function _remove> returned NULL without setting an error`. Not sure what's going on...

fiona/ogrext.pyx
snorfalorpagus: There are a few places we could apply this (anywhere `OGRERR_*` are used). Perhaps better to do this systematically as a separate PR?

snorfalorpagus:
I've made the requested changes, except for `exc_wrap_int` which I think is better left to a separate PR.

sgillies:
@snorfalorpagus fine with me to defer on `exc_wrap_int`, that doesn't affect the API.

Issue #465, opened 2017-07-28
3D geometry schema lost when writing and reading shapefile


ChrisVolz:
Hello,

I am wondering if something changed lately dealing with 3D geometries?  Writing 3D geometries with Fiona and reading the results leads to degeneration to 2D geometry? In earlier versions (1.6.x) I haven't encountered this issue.
  

```
import fiona
from shapely.geometry import Point, mapping

schema = {'geometry':'3D Point', 'properties':{'name':'str'}}
with fiona.open('test.shp', 'w', driver='ESRI Shapefile', schema=schema) as shp:
    pt = Point(1,1,1)
    shp.write({'geometry':mapping(pt), 'properties':{'name':'test'}})
with fiona.open('test.shp') as shp:
     print shp.schema
```

I expect to be written:
`{'geometry': '3D Point', 'properties': OrderedDict([(u'name', 'str:80')])}`

but I get:
`{'geometry': 'Point', 'properties': OrderedDict([(u'name', 'str:80')])}`

(Fiona version 1.7.8, shapely version 1.5.12, tested on Linux and Windows)

PS: ArcGIS seems to reckognize the written file as 3D Shapefile. Hence, it may be rather a problem with reading than writing.

EDIT: No one has a solution to this problem? Or can't you reproduce the error? This seems quite a big thing to me...

sgillies:
@ChrisVolz I'm sorry about the confusion. We have eliminated the distinction between 2D and 3D geometry types. If you open a 'Point 3D' shapefile with Fiona, the geometry type in the schema is 'Point'. 'Point 3D' is accepted when you create a shapefile, but it's now only an alias for 'Point'.

ChrisVolz:
Thanx for your answer.

But I must say, I am not totally happy with the new behaviour. The following use case makes me a lot of problems in my projects:

- I open a shapefile with 3D Geometries, store the meta information and read all features
- I write a new shapefile with the stored meta information and write all features

Then I have **lost** all 3D information; only 2D geometries are written. Apparently I have to explicitely set the geometry to '3D Polygon' (or '3D Point') to preserve the 3D information. That is quite strange. However, I can live with it ;)

jorisvandenbossche:
> 'Point 3D' is accepted when you create a shapefile, but it's now only an alias for 'Point'.

[...]

PR #539, opened 2018-01-20
Geometry type schema is now more flexible

### LINKED ISSUE #446: mandatory schema validation prevents writing mixed geometries to formats where the underlying driver supports it (e.g. MapInfo TAB, MIF)
## Expected behaviour:

It should be possible for Fiona to write features with different geometry types to a single layer, if the underlying driver supports it.

Further, it should ideally be easy for a user to do this in the special simple case where they are trying to write out a copy of a layer they have just read in using Fiona.

I ran into this problem for MapInfo files, but this expected behaviour seems reasonable for any file format / data source that can store mixed geometry types. MapInfo TAB and MIF layers can store mixed geometry types in a single layer and the formats do not appear to have any concept of a per-layer expected geometry type. Reading and writing mixed geometry MapInfo layers is supported by the underlying driver used by Fiona.

## Actual behaviour:

Fiona can read in features from a MapInfo TAB / MIF file containing mixed geometry types, but then by default refuses to write them out due due to failed schema validation of the geometry types.

Fiona does not appear to provide a documented way to disable schema validation of geometry types or define a schema that can hold mixed geometry types (e.g. Points and LineStrings and Polygons). There doesn't appear to be an undocumented way to disable it either, short of patches/hacks.

Please see attached script and example MapInfo MIF data file containing a Point, LineString and Polygon:

[mapinfo_multigeometry_issue.tar.gz](https://github.com/Toblerity/Fiona/files/1022205/mapinfo_multigeometry_issue.tar.gz)

To reproduce

0.   assuming you are in a fiona dev environment
1.   extract the attached `mix_geometry.mif`, `mixed_geometry.mid` and `test_mapinfo.py` somewhere
2.   cd into directory that contains `test_mapinfo.py`
3.   `py.test .`

Observe `test_by_default_we_can_write_a_copy_of_a_mixed_geometry_mapinfo_layer_we_read_in` xfails.

Observe `test_using_ugly_hacks_we_can_write_a_copy_of_a_mixed_geometry_mapinfo_layer_we_read_in` passes, proving that writing the layer actually works if we bypass Fiona's schema validation.

## Operating system, Fiona version, provenance

Ubuntu 16.04 LTS
Fiona 1.7.5
Fiona cloned from this github repo, b9eb1b5545a7e10f33c2967ec15d2e7abb2978c3

Python 2.7.12
GDAL 1.11.3

Same behaviour on Windows 7 using Fiona version 1.5-ish of unknown provenance.

## Possible ways to support this.

I am happy to contribute a pull request to improve support for this, if this enhancement in functionality is desired and there's a clean way it can work.

Here's a sketch of one design to extend Fiona's schema validation to support this:

Define a new value of `schema['geometry']`, e.g. `'anything'`, that means any geometry type is permitted. Perhaps another way of encoding this would be to make the `geometry` item of a schema itself optional.

Modify Fiona's code for writing features to skip validation of feature geometry types if the schema indicates that any geometry type is permitted.

Modify Fiona's code for inferring the schema geometry type when reading a layer to assume that the layer may contain any geometry type if the layer doesn't explicitly specify a uniform geometry type and the user has not explicitly specified a schema geometry type.

* Easy case: when reading a MapInfo file that contains points and linestrings and polygons, infer the geometry type as "any geometry type is permitted"
* Trickier case: when reading a MapInfo file that contains only points, infer the geometry type as "any geometry type is permitted" because by default for MapInfo format one is allowed to put whatever geometry in a layer regardless of what is already in there (assuming no schema external to the MapInfo file has been explicitly defined and provided by the user).

The overall principle of this suggestion is to make the default schema validation behaviour maximally permissive in terms of letting the user perform operations that are potentially supported by the underlying driver if Fiona has not been explicitly given a schema.

snorfalorpagus:
This PR modifies the validation done when writing geometries to ensure they match the collection schema.

The existing behaviour of specifying a single geometry type is unchanged:

```
fiona.open(..., schema={"geometry": "Point", ...})
```

Alternatively you can specify a list of types:

```
fiona.open(..., schema={"geometry": ("Point", "LineString", "Polygon", ...})
```

Or allow any type:

```
fiona.open(..., schema={"geometry": "Unknown", ...})
```

This works for the edge cases I could think of (e.g. ESRI Shapefile doesn't care between LineString/MultiLineString, Polygon/MultiPolygon), but I may have missed something out.

A new exception is raises if a geometry is written that doesn't match the schema: `GeometryTypeValidationError`.

It's possible to specify a schema that isn't actually valid for the driver. For example, a schema of `("Point", "LineString")` for an ESRI Shapefile. The code doesn't check for this explicitly, but will raise a `RuntimeError` if you actually try to write something invalid.

Closes #446.

### DIFF
--- a/fiona/collection.py
+++ b/fiona/collection.py
@@ -10,7 +10,7 @@ from fiona.ogrext import Iterator, ItemsIterator, KeysIterator
 from fiona.ogrext import Session, WritingSession
 from fiona.ogrext import (
     calc_gdal_version_num, get_gdal_version_num, get_gdal_release_name)
-from fiona.ogrext import buffer_to_virtual_file, remove_virtual_file
+from fiona.ogrext import buffer_to_virtual_file, remove_virtual_file, GEOMETRY_TYPES
 from fiona.errors import DriverError, SchemaError, CRSError
 from fiona._drivers import driver_count, GDALEnv
 from fiona.drvsupport import supported_drivers, AWSGDALEnv
@@ -167,6 +167,9 @@ class Collection(object):
             self.guard_driver_mode()
             if not self.encoding:
                 self.encoding = self.session.get_fileencoding().lower()
+
+        if self.mode in ("a", "w"):
+            self._valid_geom_types = _get_valid_geom_types(self.schema, self.driver)
 
     def __repr__(self):
         return "<%s Collection '%s', mode '%s' at %s>" % (
@@ -429,6 +432,30 @@ class Collection(object):
         # Note: you can't count on this being called. Call close() explicitly
         # or use the context manager protocol ("with").
         self.close()
+
+
+def _get_valid_geom_types(schema, driver):
+    schema_geom_type = schema["geometry"]
+    if isinstance(schema_geom_type, string_types) or schema_geom_type is None:
+        schema_geom_type = (schema_geom_type,)
+    valid_types = set()
+    for geom_type in schema_geom_type:
+        valid_types.add(str(geom_type).lstrip("3D "))
+
+    if "Unknown" in valid_types:
+        ALL_GEOMETRY_TYPES = set([
+            geom_type for geom_type in GEOMETRY_TYPES.values()
+            if "3D " not in geom_type and geom_type != "None"])
+        valid_types.update(ALL_GEOMETRY_TYPES)
+        valid_types.remove("Unknown")
+
+    if driver == "ESRI Shapefile" and "Point" not in valid_types:
+        print("OMG")
+        for geom_type in list(valid_types):
+            if not geom_type.startswith("Multi"):
+                valid_types.add("Multi"+geom_type)
+
+    return valid_types
 
 
 def get_filetype(bytesbuf):
--- a/fiona/collection.py
+++ b/fiona/collection.py
@@ -11,7 +11,7 @@ from fiona.ogrext import Session, WritingSession
 from fiona.ogrext import (
     calc_gdal_version_num, get_gdal_version_num, get_gdal_release_name)
 from fiona.ogrext import buffer_to_virtual_file, remove_virtual_file, GEOMETRY_TYPES
-from fiona.errors import DriverError, SchemaError, CRSError
+from fiona.errors import DriverError, SchemaError, CRSError, UnsupportedGeometryTypeError
 from fiona._drivers import driver_count, GDALEnv
 from fiona.drvsupport import supported_drivers, AWSGDALEnv
 from six import string_types, binary_type
@@ -406,7 +406,7 @@ class Collection(object):
     def close(self):
         """In append or write mode, flushes data to disk, then ends
         access."""
-        if self.session is not None:
+        if self.session is not None and self.session.isactive():
             if self.mode in ('a', 'w'):
                 self.flush()
             log.debug("Flushed buffer")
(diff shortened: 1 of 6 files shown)


fiona/collection.py
```diff
@@ -427,6 +430,30 @@ def __del__(self):
...
+
+    if "Unknown" in valid_types:
+        ALL_GEOMETRY_TYPES = set([
+            geom_type for geom_type in GEOMETRY_TYPES.values()
+            if "3D " not in geom_type and geom_type != "None"])
+        valid_types.update(ALL_GEOMETRY_TYPES)
+        valid_types.remove("Unknown")
+
+    if driver == "ESRI Shapefile" and "Point" not in valid_types:
+        print("OMG")
```
sgillies: Funny, but I think we should remove this :)

fiona/collection.py
```diff
@@ -427,6 +430,30 @@ def __del__(self):
...
 
+def _get_valid_geom_types(schema, driver):
+    schema_geom_type = schema["geometry"]
+    if isinstance(schema_geom_type, string_types) or schema_geom_type is None:
+        schema_geom_type = (schema_geom_type,)
+    valid_types = set()
+    for geom_type in schema_geom_type:
+        valid_types.add(str(geom_type).lstrip("3D "))
+
+    if "Unknown" in valid_types:
```
sgillies: I think "Any" might be better than "Unknown".

fiona/errors.py
```diff
@@ -33,5 +33,9 @@ class UnsupportedGeometryTypeError(KeyError):
     """When a OGR geometry type isn't supported by Fiona."""
 
 
+class GeometryTypeValidationError(FionaValueError):
+    """Tried to write a geometry type not specified in the schema"""
```
sgillies: Great.

fiona/collection.py
snorfalorpagus: Ha! This wasn't supposed to be left in. It's my go-to statement for checking branches are being run. I'll remove it.

fiona/collection.py
snorfalorpagus: The reason for using "Unknown" is that this is what GDAL uses when it opens a multi-type layer. I think we need to support it, so that you can reuse the schema when reading from one layer directly into another - I'll add a test specifically for this. I don't mind adding an alias for this for writing layers. "Any" is my preference.

sgillies: Ah, okay, I see it does. I quibble with GDAL's semantics all the time. Consider the two statements:

1. Create a new layer to store features with an *unknown* geometry type
2. Create a new later to store features with *any* geometry type

I think *any* suits the writing situation better and *unknown* suits the reading situation better, if you see what I mean. I think we should choose one and follow GDAL here. "unknown" it is.

snorfalorpagus:
I've added "Any" as an alias for "Unknown" when creating layers. "Unknown" will still work, and needs to so that we can re-use the metadata from a source layer in the destination layer (see `test_unknown`).

I fixed a bug in `geometry_type_code` whereby the exception was being ignored by cython.

I also added a check to see if the session `isactive` before trying to close it (in `__del__`) which was triggering a warning when a layer failed to open due to an invalid geometry type in the schema.

You'll need to squash the commits when you do the merge to avoid the `print("OMG")` ending up in the history.

snorfalorpagus:
This may need some more work in relation to #465.

Currently if you open a layer with schema="3D Point" and try to write a geometry type="Point", it works but uses a 0 value for the Z coordinate. I'm not sure what the correct behaviour should be.

sgillies:
@snorfalorpagus the options would be to use 0 for z or refuse to convert and raise an exception, yes?

Elsewhere, I've written that users shouldn't depend on Fiona schemas as a crosswalk for their data. That you shouldnt' specify `float:7.2` and expect a string like `'123.45678'` to be converted perfectly.

That said, adding a z value to 2D points is the user-friendly thing to do, and `0` is the best of all possible values.

snorfalorpagus:
I'm OK with setting the z-value to 0 instead of raising an exception so long as it's documented.

I'm looking into the error for `test_schema_geom.py::test_unknown` now - weirdly it passes on my machine.

snorfalorpagus:
I can reproduce the issue now. It looks like the branch is OK and the issue is in the merge to master.

snorfalorpagus:
I've tracked the issue down to this:

https://github.com/Toblerity/Fiona/blame/master/fiona/ogrext.pyx#L257-L262

Previously if a feature had no geometry set the fiona feature would contain `{"geometry": None}`. The new behaviour (#488) removes the key completely if the geometry is missing. This works until you try to write the feature to another layer. I see two options here: either revert the change to handling of null geometry (i.e. use `"geometry": None`), or interpret the missing key as a null geometry when writing features. @sgillies?

sgillies:
@snorfalorpagus after line 262, let's add a `else: fiona_feature["geometry"] = None` to get back to the original behavior. Okay?

snorfalorpagus:
The failures seem to be GitHub related...

snorfalorpagus:
I think I confused Travis by accidentally pushing a branch with the same name to origin, rather than my own user. The PR tests are passing which is what matters. I think this is ready now.

Issue #469, opened 2017-08-16
Make columnar-based access possible in addition to record-based model?


jorisvandenbossche:
Related to https://github.com/geopandas/geopandas/issues/491 (exploration of ways to make data ingestion faster in geopandas)

Currently `fiona` exposes a model where you access access the data by records (eg by iterating over a collection, accessing one record at a time). When the goal is to load the full dataset (eg to put all records in a geopandas GeoDataFrame), this record-based access can give some performance overhead.

Therefore, I am wondering to what extent `fiona` would welcome additions to also make columnar-based access possible. 
With columnar-based access I mean that you could get the values of all records (so of the full collection) at once in an array per property and geometry.

snorfalorpagus:
This is certainly possible from a technical perspective. I'd guess the biggest speed gains would come from copying the data from GDAL directly into a NumPy array, avoiding any intermediate Python objects. Fiona doesn't currently depend on NumPy for anything, so I think this would be best in an optional module (similar to what we've done with Shapely). This module could provide methods that wrap `Session` and a variant of `FeatureBuilder.build` to return arrays. For the geometry column it's probably easiest to return an array of WKB/WKT data, which could then be quickly parsed with Shapely.

jorisvandenbossche:
> there's a requirement to iterate over the records in the data and build up a list of values.

Yes, but I think what @snorfalorpagus does is, while doing this iteration, directly filling in numpy arrays on the C level (so you don't need to build up a 'list' (in the python sense) of values).

> I suggest we look at 3 smaller optimizations.

To be really sure about the need, I think we should try to do some benchmarking of the proof of concept @snorfalorpagus now made compared the the options you outline above. 

> With these optimizations, GeoPandas could make numpy structured arrays from the new fiona collection iterator, yes?

Yes, that would be possible, and it would certainly also be faster as the current implementation, but how much faster is difficult to say.

I would somewhat assume that the overhead of creating intermediate python objects (whether the full feature dict or flatter tuple) is the main bottleneck, but it might also be that the actual conversion from wkt to shapely/geopandas geometries takes the most time of the current `geopandas.read_file` and in that case option 1 could already give a lot of speed-up. But as I said, I would need to do some timings to actually assess this.

brendan-ward:
I'm running an experiment with this idea over in [`pyogrio`](https://github.com/brendan-ward/pyogrio).  The idea is to use the bare essentials from `fiona` to create a numpy-oriented API to vector data sources.

[...]

PR #540, opened 2018-01-21
[WIP] Initial commit of bulk reading into numpy arrays

snorfalorpagus:
This is a first attempt at providing a bulk load of data into NumPy arrays, as suggested in #469.

It's a work in progress. I'm sure it's not perfectly optimized and there are some edge cases that aren't handle yet (e.g. NULL values in `int` columns), date fields.

Does this provide the right kind of interface for geopandas @jorisvandenbossche? Do you have any opinions @sgillies?


jorisvandenbossche:
@snorfalorpagus Thanks a lot for exploring this! 
To say whether this would output something useful for geopandas, I would first need to take a closer look (but in principle returning numpy arrays is fine). You now create one array per dtype? That would be perfect for geopandas, although from a more general user point of view, a 1D array per field might make more sense. 

With regard to the TODO "best way to return geometries for shapely?", I would say either WKB or WKT, depending on which is the fastest/cheapest to create and convert afterwards (which I assume will be WKB?)

snorfalorpagus:
- [ ] Support WKT or WKB
- [ ] Support for all data types available in Fiona already
- [ ] Support for ignoring fields
- [ ] Improved iterating over features

snorfalorpagus:
It looks like GDAL 1.x is having problem with the datetime field? It looks like we don't actually test this in the regular reader/writer.

fiona/_vectorized.pyx
```diff
@@ -0,0 +1,188 @@
...
+    session = collection.session
+    encoding = session._fileencoding
+
+    if session.cogr_layer == NULL:
+        raise ValueError("Null layer")
+
+    length = OGR_L_GetFeatureCount(session.cogr_layer, 0)
+
+    data_fids = np.empty([length], dtype=object)
+    data_properties = {}
```
snowman2: Thoughts on using a [structured array](https://docs.scipy.org/doc/numpy-1.15.0/user/basics.rec.html) for `data_properties`?

PR #541, opened 2018-01-29
Do not collapse Z and M types to 2D

sgillies:
See #465.


sgillies:
I think this fixes the bug introduced in 0a2bc7c.

`M` geometry types are still not supported.

sgillies:
Ping @snorfalorpagus for a review.

PR #544, opened 2018-02-08
Fix fio cat with bbox when using more than one file

### LINKED ISSUE #543: fio cat fails with bounding box and more than one file
When concatenating features from more than one file while using a bounding box at the same time, `fio cat` fails with the following error message:

```
  File "/Users/trtim/miniconda3/envs/resource-potentials/lib/python3.6/site-packages/fiona/fio/cat.py", line 74, in cat
    bbox = tuple(map(float, bbox.split(',')))
AttributeError: 'tuple' object has no attribute 'split'
Aborted!
```

The following always fails:

    fio cat --bbox -10,20,0,30 file1 file2

while these always work:
```
fio cat --bbox -10,20,0,30 file1
fio cat --bbox -10,20,0,30 file2
fio cat file1 file2
```

## Issue
The issue is likely to be [line 73 in cat.py](https://github.com/Toblerity/Fiona/blob/620bc475e09257bb1d947228f44e68628f8ae62e/fiona/fio/cat.py#L73) which is executed more than once when using more than one file, but really should be executed once only.

## Environment

Fiona 1.7.11, GDAL 2.1.3, macOS 10.12.6

timtroendle:
Fixes #543.

### DIFF
--- a/fiona/fio/cat.py
+++ b/fiona/fio/cat.py
@@ -65,14 +65,14 @@ def cat(ctx, files, precision, indent, compact, ignore_errors, dst_crs,
             layer[str(i)] = [0]
     try:
         with fiona.drivers(CPL_DEBUG=verbosity > 2):
+            if bbox:
+                try:
+                    bbox = tuple(map(float, bbox.split(',')))
+                except ValueError:
+                    bbox = json.loads(bbox)
             for i, path in enumerate(files, 1):
                 for lyr in layer[str(i)]:
                     with fiona.open(path, layer=lyr) as src:
-                        if bbox:
-                            try:
-                                bbox = tuple(map(float, bbox.split(',')))
-                            except ValueError:
-                                bbox = json.loads(bbox)
                         for i, feat in src.items(bbox=bbox):
                             if dst_crs or precision >= 0:
                                 g = transform_geom(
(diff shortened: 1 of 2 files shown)


tests/test_fio_cat.py
```diff
@@ -46,6 +46,16 @@ def test_bbox_yes():
...
 
 
+def test_bbox_yes_two_files():
+    runner = CliRunner()
+    result = runner.invoke(
+        cat.cat,
+        [WILDSHP, WILDSHP, '--bbox', '-109,37,-107,39'],
+        catch_exceptions=False)
+    assert result.exit_code == 0
+    assert result.output.count('"Feature"') == 38
```
sgillies: Great!

fiona/fio/cat.py
```diff
@@ -65,14 +65,14 @@ def cat(ctx, files, precision, indent, compact, ignore_errors, dst_crs,
             layer[str(i)] = [0]
     try:
         with fiona.drivers(CPL_DEBUG=verbosity > 2):
+            if bbox:
+                try:
+                    bbox = tuple(map(float, bbox.split(',')))
+                except ValueError:
+                    bbox = json.loads(bbox)
```
sgillies: Yes, perfect.

PR #532, opened 2018-01-11
Improved error handling when GDAL 2.x fails to open existing layer

snorfalorpagus:
This PR provides slightly more informative error messages when failing to open a dataset in read mode. Previous more detailed messages from GDAL could be hidden by a generic (and potentially incorrect) message from Fiona.

I came across this issue in the 2.3.x (svn-trunk) job on Travis, which was raising this error:

```
fiona.errors.FionaValueError: No dataset found at path '/vsimem/c513cc7b5ce848a6a3358d4c0854f9bd.json' using drivers: {'GeoJSON'}
```

This is now replaced by a more correct message:

```
fiona.errors.FionaValueError: `/vsimem/8c013c21ebeb42cabfb0bb93239bf4d4.json' not recognized as a supported file format. using drivers {'GeoJSON'}
```

I think the actual problem is something in GDAL (see upstream issue https://trac.osgeo.org/gdal/ticket/7198 ), but the message Fiona gives is more useful now.

PR #555, opened 2018-03-15
Some fixes for Python 2.7, and a fix for the binary field test

snorfalorpagus:
Another batch of commits extracted from #532.

https://github.com/Toblerity/Fiona/commit/6ebf990d7de43636a98796debb7aa932693ef546 fixes an issue whereby the binary field test ways always skipped when running on CI.


snorfalorpagus:
@sgillies Yes

PR #559, opened 2018-03-20
Expand range of Python and GDAL versions in Appveyor matrix

sgillies:
- Pythons: 2.7 and 3.6
- GDAL: 1.11.4, 2.2.3, and trunk (from GIS internals)

I'm still not using conda-forge for GDAL because of the value in testing against the GDAL trunk.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -121,4 +121,4 @@ test_script:
 
 matrix:
   allow_failures:
-    GDAL_VERSION : trunk
+    - GDAL_VERSION : trunk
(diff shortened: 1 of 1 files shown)


snorfalorpagus:
I agree it's worth sticking with gisinternals for the nightly builds.

This looks like it's working as expected. The failures for GDAL 2.3 are the same as on Linux (Travis).

The failures for travis-ci/pr look like a network issue?

sgillies:
@snorfalorpagus yes, I think so about the network failure, I restarted one and it passed, now I'm restarting the other two.

PR #563, opened 2018-03-25
Test support for date/datetime/time fields

snorfalorpagus:
#540 got me looking at the support for date/datetime/time fields. It's a little complicated, with different drivers having different levels of support and behaviours. I've added some tests to illustrate the current behaviour. I don't think anything needs to change, except that we should raise a more useful error when trying to create a time-type field in an ESRI Shapefile, which currently raises an unexpected KeyError.

### DIFF
--- a/fiona/collection.py
+++ b/fiona/collection.py
@@ -11,7 +11,7 @@ from fiona.ogrext import Session, WritingSession
 from fiona.ogrext import (
     calc_gdal_version_num, get_gdal_version_num, get_gdal_release_name)
 from fiona.ogrext import buffer_to_virtual_file, remove_virtual_file, GEOMETRY_TYPES
-from fiona.errors import DriverError, SchemaError, CRSError, UnsupportedGeometryTypeError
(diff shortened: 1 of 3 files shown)


snorfalorpagus:
Apparently GPKG support for datetime is in GDAL 2.x only. Need to update test further.

snorfalorpagus:
Need to update changelog if this is accepted.

sgillies:
@snorfalorpagus sorry for the review backlog! I'll try to get to them tomorrow. I see that we have an unresolved dependency on Windows.

snorfalorpagus:
> sorry for the review backlog! I'll try to get to them tomorrow. I see that we have an unresolved dependency on Windows.

No problem! I created #593 for the dependency issue.

sgillies:
Wow, I'm baffled by the failure on Appveyor. The click-plugins project hasn't changed in a long time and we used to pass the tests with pip 10.0.1. I'm stumped for now.

PR #564, opened 2018-03-28
Make way for int64, the best integer

### LINKED ISSUE #562: Fix for 32 vs 64-bit integer ambiguity
This issue was reported in https://github.com/geopandas/geopandas/issues/691 and probably also complicates PR #540 in some way.

@snorfalorpagus I think it's time to make 'int' in a Fiona schema mean 64-bit integer as in Python and Numpy.

```
>>> numpy.dtype('int')
dtype('int64')
```

and, instead of relying on integer field width, add a new 'int32' name here: https://github.com/Toblerity/Fiona/blob/master/fiona/ogrext.pyx#L48 for 32 bit-integers.

Fiona 1.8 will then fix that particular GeoPandas issue. We will not backport this to 1.7, where there is a field width workaround: `int:18` for big ints.

sgillies:
Collection schemas still report 'int:9' or 'int:18' for now (with GDAL 2) but in creating a collection, 'int' means 'int64'.

Resolves #562

### DIFF
--- a/fiona/_err.pyx
+++ b/fiona/_err.pyx
@@ -226,6 +226,18 @@ cdef inline object exc_check():
         return
 
 
+cdef int exc_wrap_int(int err) except -1:
+    """Wrap a GDAL/OGR function that returns CPLErr or OGRErr (int)
+
+    Raises a Rasterio exception if a non-fatal error has be set.
+    """
+    if err:
+        exc = exc_check()
+        if exc:
+            raise exc
+    return err
+
+
 cdef void *exc_wrap_pointer(void *ptr) except NULL:
     """Wrap a GDAL/OGR function that returns GDALDatasetH etc (void *)
     Raises a Rasterio exception if a non-fatal error has be set.
--- a/fiona/_err.pyx
+++ b/fiona/_err.pyx
@@ -136,10 +136,19 @@ class CPLE_AWSInvalidCredentialsError(CPLE_BaseError):
 class CPLE_AWSSignatureDoesNotMatchError(CPLE_BaseError):
     pass
 
+
 class FionaNullPointerError(CPLE_BaseError):
     """
     Returned from exc_wrap_pointer when a NULL pointer is passed, but no GDAL
     error was raised.
+    """
+    pass
+
+
+class FionaCPLError(CPLE_BaseError):
+    """
+    Returned from exc_wrap_int when a error code is returned, but no GDAL
+    error was set.
     """
     pass
 
@@ -235,6 +244,8 @@ cdef int exc_wrap_int(int err) except -1:
         exc = exc_check()
         if exc:
             raise exc
+        else:
+            raise FionaCPLError(-1, -1, "The wrapped function returned an error code, but no error message was set.")
     return err
 
 
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -29,6 +29,7 @@ from fiona.errors import (
 from fiona.compat import OrderedDict
 from fiona.rfc3339 import parse_date, parse_datetime, parse_time
 from fiona.rfc3339 import FionaDateType, FionaDateTimeType, FionaTimeType
+from fiona.schema import FIELD_TYPES, FIELD_TYPES_MAP, normalize_field_type
 
 from fiona._shim cimport is_field_null
 
@@ -37,40 +38,7 @@ from libc.string cimport strcmp
 from cpython cimport PyBytes_FromStringAndSize, PyBytes_AsString
 
 
-log = logging.getLogger("Fiona")
-
-# Mapping of OGR integer field types to Fiona field type names.
-#
-# Lists are currently unsupported in this version, but might be done as
-# arrays in a future version.
-
-FIELD_TYPES = [
-    'int',          # OFTInteger, Simple 32bit integer
-    None,           # OFTIntegerList, List of 32bit integers
-    'float',        # OFTReal, Double Precision floating point
-    None,           # OFTRealList, List of doubles
-    'str',          # OFTString, String of ASCII chars
-    None,           # OFTStringList, Array of strings
-    None,           # OFTWideString, deprecated
-    None,           # OFTWideStringList, deprecated
-    'bytes',        # OFTBinary, Raw Binary data
-    'date',         # OFTDate, Date
-    'time',         # OFTTime, Time
-    'datetime',     # OFTDateTime, Date and Time
-    'int',          # OFTInteger64, Single 64bit integer
-    None,           # OFTInteger64List, List of 64bit integers
-    ]
-
-# Mapping of Fiona field type names to Python types.
-FIELD_TYPES_MAP = {
-    'int':      int,
-    'float':    float,
-    'str':      text_type,
-    'date':     FionaDateType,
-    'time':     FionaTimeType,
-    'datetime': FionaDateTimeType,
-    'bytes':    bytes,
-   }
+log = logging.getLogger(__name__)
 
 DEFAULT_TRANSACTION_SIZE = 20000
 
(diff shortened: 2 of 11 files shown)


fiona/ogrext.pyx
```diff
@@ -29,6 +29,7 @@ from fiona.errors import (
 from fiona.compat import OrderedDict
 from fiona.rfc3339 import parse_date, parse_datetime, parse_time
 from fiona.rfc3339 import FionaDateType, FionaDateTimeType, FionaTimeType
+from fiona.schema import FIELD_TYPES, FIELD_TYPES_MAP, normalize_field_type
```
sgillies: A new module dedicated to schema and field stuff.

fiona/ogrext.pyx
```diff
@@ -37,40 +38,7 @@ from libc.string cimport strcmp
...
-FIELD_TYPES_MAP = {
-    'int':      int,
-    'float':    float,
-    'str':      text_type,
-    'date':     FionaDateType,
-    'time':     FionaTimeType,
-    'datetime': FionaDateTimeType,
-    'bytes':    bytes,
-   }
+log = logging.getLogger(__name__)
```
sgillies: This is way more correct and makes testing the logging more simple (just configure the root logger).

fiona/ogrext.pyx
```diff
@@ -154,8 +122,29 @@ cdef class FeatureBuilder:
...
+        ignore_fields : sequence
+            A sequence of field names that will be ignored and omitted
+            in the Fiona feature properties
+        ignore_geometry : bool
+            Flag for whether the OGR geometry field is to be ignored
+
+        Returns
+        -------
+        dict
+        """
```
sgillies: Added a missing docstring.

fiona/ogrext.pyx
```diff
@@ -985,12 +993,17 @@ cdef class WritingSession(Session):
...
 
                 field_type = FIELD_TYPES.index(value)
-                if GDAL_VERSION_NUM >= 2000000:
-                    # See https://trac.osgeo.org/gdal/wiki/rfc31_ogr_64
-                    if value == 'int' and (width is not None and width >= 10):
-                        field_type = 12
+                #if GDAL_VERSION_NUM >= 2000000:
+                #    # See https://trac.osgeo.org/gdal/wiki/rfc31_ogr_64
+                #    if value == 'int64':
+                #        field_type = 12
```
sgillies: TODO: remove commented code.

fiona/schema.py
```diff
@@ -0,0 +1,71 @@
...
+
+from fiona.errors import SchemaError
+from fiona.rfc3339 import FionaDateType, FionaDateTimeType, FionaTimeType
+
+
+# Mapping of OGR integer field types to Fiona field type names.
+# Lists are currently unsupported in this version, but might be done as
+# arrays in a future version.
+#
+# NB: in Fiona 2, we will switch 'int' -> 'int64'.
```
sgillies: TODO: remove this comment.

fiona/schema.py
```diff
@@ -0,0 +1,71 @@
...
+        if width == 0 or width >= 10:
+            return 'int64'
+        else:
+            return 'int32'
+    elif ftype.startswith('str'):
+        return 'str'
+    elif ftype.startswith('float'):
+        return 'float'
+    else:
+        raise SchemaError("Unknown field type: {}".format(ftype))
```
sgillies: This function normalizes legacy schema types to the new and improved set of types.

tests/test_transactions.py
```diff
@@ -26,7 +26,9 @@ def emit(self, record):
         if self.pattern in record.msg:
             self.history[record.msg] += 1
 
-log = logging.getLogger("Fiona")
+
+log = logging.getLogger()
```
sgillies: Configure the root logger, not an obscure "Fiona' logger. I can't believe I was so clueless about logging for so long.

tests/test_schema.py
```diff
@@ -99,31 +102,31 @@ def test_schema(self):
...
+                        ('TaxReturns', 'float:24.15'),
+                        ('State', 'str:80'),
+                        ('Location', 'str:80'),
+                        ('GSrchCnt', 'float:24.15'),
+                        ('TotalWages', 'float:24.15'),
+                        ('Lat', 'float:24.15'),
+                        ('MTFCC10', 'str:80'),
+                        ('INTPTLON10', 'str:80'),
+                        ('GEOID10', 'str:80'),
+                        ('Decommisio', 'str:80')]))
```
sgillies: Whole lotta formatting to satisfy flake8.

sgillies:
@snorfalorpagus would you be willing to review this?

The state of AppVeyor builds is still mixed up. The first 3 fail due to missing AWS credentials. The 4th fails because Python 2.7 and GDAL C++ are incompatible (TODO: skip this combination). The 5th passes.

snorfalorpagus:
Yes I will review this week.

sgillies:
Thanks, @snorfalorpagus ! The Rasterio push ate up all my time over the past 3 months, but I'm determined to see Fiona 1.8 come out later this summer. If only so we never have to backport anything to maint-1.7 again 😅

fiona/ogrext.pyx
```diff
@@ -539,25 +552,35 @@ cdef class Session:
         cogr_featuredefn = OGR_L_GetLayerDefn(self.cogr_layer)
         if cogr_featuredefn == NULL:
             raise ValueError("Null feature definition")
+
         n = OGR_FD_GetFieldCount(cogr_featuredefn)
+
         for i from 0 <= i < n:
+            cogr_fielddefn == NULL
```
snorfalorpagus: This doesn't do anything. I think it can be removed.

fiona/ogrext.pyx
```diff
@@ -973,52 +996,70 @@ cdef class WritingSession(Session):
...
+                log.debug("Begin creating field: %r value: %r", key, value)
+
                 field_subtype = OFSTNone
 
                 # Convert 'long' to 'int'. See
                 # https://github.com/Toblerity/Fiona/issues/101.
-                if value == 'long':
-                    value = 'int'
-                
+                if GDAL_VERSION_NUM >= 2000000 and value in ('int', 'long'):
```
snorfalorpagus: This could use the new `GDALVersion.major >= 2` syntax which is more readable.

fiona/ogrext.pyx
```diff
@@ -973,52 +996,70 @@ cdef class WritingSession(Session):
...
 
-                cogr_fielddefn = OGR_Fld_Create(
-                    key_bytes,
-                    field_type)
+                try:
+                    cogr_fielddefn = exc_wrap_pointer(
+                        OGR_Fld_Create(key_bytes, field_type))
+                    log.debug("Creating field %r (%r) with type %r (%r)", key, key_bytes, field_type, value)
+                except Exception:
+                    raise
```
snorfalorpagus: If all exceptions are raised is there a need for the try-except block?

fiona/_err.pyx
```diff
@@ -226,6 +226,18 @@ cdef inline object exc_check():
         return
 
 
+cdef int exc_wrap_int(int err) except -1:
```
snorfalorpagus: I tried writing something like this in an earlier PR. I think there might be some functions which return a non-zero error code, but don't set the GDAL exception. I can't remember the specifics though. I think it's worth adding an `else` to this function like `exc_wrap_pointer` raises `FionaNullPointerError`. It may never happen, but if it does and we skip over a non-zero error it can be difficult to debug.

fiona/schema.pyx
```diff
@@ -0,0 +1,83 @@
...
+
+from fiona.errors import SchemaError
+from fiona.rfc3339 import FionaDateType, FionaDateTimeType, FionaTimeType
+
+
+cdef extern from "gdal.h":
+    char * GDALVersionInfo (char *pszRequest)
+
+
+def get_gdal_version_num():
```
snorfalorpagus: I'm not keen on the duplication of this from `ogrext.pyx` although I can see why it's needed. Can we either `cdef` it or prefix an `_` so that users aren't tempted to do `from fiona.schema import get_gdal_version_num`?

tests/test_bigint.py
```diff
@@ -31,6 +34,8 @@ def setUp(self):
     def tearDown(self):
         shutil.rmtree(self.tempdir)
 
+    @pytest.mark.xfail(get_gdal_version_num() < calc_gdal_version_num(2, 0, 0),
```
snorfalorpagus: ```@pytest.mark.xfail(fiona.gdal_version.major < 2,```

tests/test_bigint.py
```diff
@@ -43,27 +48,39 @@ def testCreateBigIntSchema(self):
...
+def test_issue691(tmpdir, dtype):
+    """Type 'int' maps to 'int64'"""
+    schema = {'geometry': 'Any', 'properties': {'foo': dtype}}
+    with fiona.open(
+            str(tmpdir.join('test.shp')), 'w', driver='Shapefile',
+            schema=schema, crs='epsg:4326') as dst:
+        dst.write({
+            'type': 'Feature',
+            'geometry': {'type': 'Point',
+                         'coordinates': (-122.278015, 37.868995)},
```
snorfalorpagus: Any significance to these coordinates? :)

snorfalorpagus:
Mostly minor comments, overall looks good.

Issue #573, opened 2018-04-15
gdal 2.3 failing tests


rbuffat:
I was looking at why the travis builds against gdal 2.3 are failing.

When building gdal on travis timeouts after 10min when no compilation output is displayed with the silent option `make -s`. 

test_bytescollection.py -> test_grenada_bytes_geojson does not raise a FionaValueError anymore but on a first glance seems to detect that it json now. This could be related to https://trac.osgeo.org/gdal/wiki/rfc70_output_format_guess

The other failing tests seem to be related to the following behavior change of gdal: The tests fail with "Missing url parameter": https://travis-ci.org/Toblerity/Fiona/jobs/366895108#L2927
 
For http URLs, fiona currently creates the following path: `/vsicurl/svn.osgeo.org/gdal/trunk/autotest/ogr/data/poly.shp`

With gdal 2.2.2, this works:
`ogrinfo /vsicurl/svn.osgeo.org/gdal/trunk/autotest/ogr/data/poly.shp`

with  GDAL 2.3.0dev-9a563c5-dirty, released 2018/04/13:
```
/opt/gdal23/bin/ogrinfo /vsicurl/svn.osgeo.org/gdal/trunk/autotest/ogr/data/poly.shp
ERROR 5: Missing url parameter
ERROR 5: Missing url parameter
ERROR 5: Missing url parameter
ERROR 5: Missing url parameter
ERROR 5: Missing url parameter
ERROR 5: Missing url parameter                                                                                                                                                                                                                     
ERROR 5: Missing url parameter                                                                                                                                                                                                                     
ERROR 5: Missing url parameter                                                                                                                                                                                                                     
ERROR 5: Missing url parameter                                                                                                                                                                                                                     
ERROR 5: Missing url parameter                                                                                                                                                                                                                     
ERROR 1: Only read-only mode is supported for /vsicurl                                                                                                                                                                                             
FAILURE:                                                                                                                                                                                                                                           
Unable to open datasource `/vsicurl/svn.osgeo.org/gdal/trunk/autotest/ogr/data/poly.shp' with the following drivers.     
```

While `/opt/gdal23/bin/ogrinfo /vsicurl/http://svn.osgeo.org/gdal/trunk/autotest/ogr/data/poly.shp`
works. I did not test if we would modify the vsicurl paths to include the protocol would solve the issue, but I suspect so.

@sgillies  Was the protocol removed from the vsicurl uri on purpose or is this an upstream bug?

[...]

PR #576, opened 2018-04-16
Gdal23 fixes

rbuffat:
This PR implements https://github.com/Toblerity/Fiona/issues/573

It includes two bigger changes and a few smaller ones:

**Disable test_grenada_bytes_geojson**
In Gdal 2.3 no FionaValueError is thrown anymore. This is possible due to [RFC70](https://trac.osgeo.org/gdal/wiki/rfc70_output_format_guess): 

> Command line utilities, when neither -f nor -of are specified (note: since r39878 both switches can be indifferently used), will loop through the registered drivers and check if one or several drivers, with output capabilities, declare to recognize the extension of the output filename. 

**Include vsi remote schemes**
Changes are limited to parse_paths() in vfs.py.  This function is rather complex, I hope I thought about every cases.  There are a few tests in test_vfs.py, but I think it would be good to have a bit more documentation, either in the function itself or as test cases.

Minor changes include verbose make of gdal to avoid timeouts of travis as well as to move from svn.osgeo.org to github for the tests.

Sorry for the messy commit history, it was a bit debugging on travis as I first hesitated to install a local gdal 2.3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ install:
   - fio --gdal-version
 
 script:
-  - pytest --cov fiona --cov-report term-missing --strict
+  - pytest --cov fiona --cov-report term-missing
 
 after_success:
   - coveralls || echo "!! intermittent coveralls failure"
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ install:
   - fio --gdal-version
 
 script:
-  - pytest --cov fiona --cov-report term-missing
+  - pytest --cov fiona --cov-report term-missing --strict
 
 after_success:
   - coveralls || echo "!! intermittent coveralls failure"
--- a/fiona/vfs.py
+++ b/fiona/vfs.py
@@ -57,8 +57,11 @@ def parse_paths(uri, vfs=None):
         scheme = parts.scheme
         path = parts.path
         if parts.netloc and parts.netloc != 'localhost':
-            # We need to deal with cases such as zip+https://server.com/data.zip
-            path = "{}://{}{}".format(scheme.split("+")[-1], parts.netloc, path)
+            if scheme.split("+")[-1] in REMOTESCHEMES:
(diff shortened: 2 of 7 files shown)


sgillies:
@rbuffat thank you ! Some of the AppVeyor failures look spurious, so I'm rerunning builds for this PR.

rbuffat:
I'm not sure why it is failing on Windows. Maybe I find time to look at it in my Windows VM during the weekend.

snorfalorpagus:
I can reproduce the issue on Windows. It seems to be related to cURL and certificates.

https://github.com/OSGeo/gdal/commit/5f95e03297889f8ddaaddf155203723caef3f86e

I can get the tests to pass if I set the `SSL_CERT_FILE` environment variable.

```
set SSL_CERT_FILE=C:\gdal\bin\curl\curl-ca-bundle.crt
```

We can do this in Appveyor easily, but need a better solution for users.

rbuffat:
So this then worked before because gdal defaulted on http when no protocol was specified?

From [http://www.gdal.org/cpl__http_8h.html](http://www.gdal.org/cpl__http_8h.html#a578662ddc6557b028fecc36e302aed41) it looks like that the CAINFO support is only available for GDAL >= 2.1.3 

```
CAINFO=/path/to/bundle.crt. This is path to Certificate Authority (CA) bundle file. By default, it will be looked in a system location. If the CAINFO options is not defined, GDAL will also look if the CURL_CA_BUNDLE environment variable is defined to use it as the CAINFO value, and as a fallback to the SSL_CERT_FILE environment variable. (GDAL >= 2.1.3)
```

rbuffat:
I just noticed https://trac.osgeo.org/gdal/ticket/6732
@sgillies is there already something implemented in rasterio?

sgillies:
@rbuffat no, Rasterio doesn't handle cert locations for users. They have to use the environment variables.

Issue #311, opened 2015-12-22
Switch to cligj's features_in_arg


sgillies:
Which is ported from Fiona, and remove the now duplicated code in this project.

snorfalorpagus:
Closed by #584.

PR #584, opened 2018-05-06
Use features_in_arg from cligj in fiona/fio/load.py

sid-kap:
Addresses #311

### DIFF
--- a/fiona/fio/load.py
+++ b/fiona/fio/load.py
@@ -7,6 +7,7 @@ import json
 import logging
 
 import click
+import cligj
 
 import fiona
 from fiona.fio import options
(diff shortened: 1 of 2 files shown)


fiona/fio/load.py
```diff
@@ -23,6 +24,7 @@
 @options.src_crs_opt
 @click.option('--dst-crs', '--dst_crs',
               help="Destination CRS.  Defaults to --src-crs when not given.")
+@cligj.features_in_arg
 @click.option('--sequence / --no-sequence', default=False,
               help="Specify whether the input stream is a LF-delimited "
                    "sequence of GeoJSON features (the default) or a single "
```
sgillies: @sid-kap the sequence option and and argument for `load()` can be deleted. `cligj.features_in_arg` figures out whether the input is a sequence or not with no option needed.

Issue #409, opened 2017-01-13
Support file-like objects in fiona.open()


timtroendle:
If `fiona.open()` was accepting file-like objects instead of solely strings modelling filesystem paths, one could use many more objects as input, including but obviously not limited to `io.StringIO` or `tempfile.TemporaryFile`.

youngpm:
@timtroendle Does the `BytesCollection` help you here? You pass it in bytes, and it maps it to GDAL's VSI's filesystem so that fiona can open it:

https://github.com/Toblerity/Fiona/blob/2c8f049c861654003c0a2929e46501bae7d6b4ad/fiona/collection.py#L425-L427

sgillies:
What @youngpm points out is correct: this feature exists in Fiona, albeit at a lower level, and I've come up with something in Rasterio – [in-memory files](https://mapbox.github.io/rasterio/topics/memory-files.html) – that I'd like to port to Fiona. Because `BytesCollection` and `MemoryFile` are lower level abtractions, they're not inconsistent with `fiona.open()` accepting file-like objects, and the feature you've suggested here would necessarily flow through the existing classes.

timtroendle:
Thanks a lot @youngpm and @sgillies for pointing this out. I had a quick look at `BytesCollection` before but got the wrong impression that it was not what I was looking for. Got it to work now!

It still would ease usability to handle that transparently in the `open` function; plus one would have access to the parameters in `open` which by using `BytesCollection` one does not.

prcastro:
This is essential when trying to open files from Hadoop File System. When using [HDFS3](https://github.com/dask/hdfs3), we can access a file from e.g. Pandas using a file-like object and passing it to `pandas.read`.

Implementing this here would also enable accessing the HDFS files from geopandas.

sid-kap:
Thanks! So it seems like `fiona.open` currently accepts `str` as an argument, and interprets that as URI. However, we also want to accept `str` types that include an entire file's contents (if I'm understanding the issue correctly). How would we figure out whether the argument is a URI or an actual file? (One option is to guess based on whether the starts with a `protocol://` or based on the string length, but that might be brittle. Another option is to take another boolean parameter that says whether it's a URI or file contents?)

sgillies:
@sid-kap I don't feel that taking a file's contents as a string as the OGR GeoJSON driver does is a good idea for Python, it's better to wrap the contents up with BytesIO and pass that object:

```
[...]

PR #587, opened 2018-05-16
Accept pathlib.Path type in fiona.open

sid-kap:
Attempt to handle #409


sid-kap:
Okay, I've updated the test as you suggested.

PR #594, opened 2018-06-05
Added fiona.gdal_version named tuple

### LINKED ISSUE #575: Add GDAL version number as named tuple
It would be useful if we had access to the GDAL version number in a named tuple, e.g.

```python
import fiona
print(fiona.gdal_version.major)
print(fiona.gdal_version.minor)
print(fiona.gdal_version.release)

assert fiona.gdal_version.major >= 2
assert fiona.gdal_version >= (2, 2, 0)
```

snorfalorpagus:
Adds a named tuple with the runtime GDAL version number.

```python
import fiona
print(fiona.gdal_version.major)
print(fiona.gdal_version.minor)
print(fiona.gdal_version.release)

assert fiona.gdal_version.major >= 2
assert fiona.gdal_version >= (2, 2, 0)
```

Closes #575.

### DIFF
--- a/fiona/__init__.py
+++ b/fiona/__init__.py
@@ -86,7 +86,8 @@ from fiona.compat import OrderedDict
 from fiona.io import MemoryFile
 from fiona.ogrext import _bounds, _listlayers, FIELD_TYPES_MAP, _remove, _remove_layer
 from fiona.ogrext import (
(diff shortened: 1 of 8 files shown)


snorfalorpagus:
Note that rasterio takes a much heavier approach to this: https://github.com/mapbox/rasterio/blob/master/rasterio/env.py#L352-L526

PR #610, opened 2018-07-18
TST: Fix flaky listing test.

QuLogic:
Since this reads layer names from multiple files, there's no guarantee that they will be in order since the filesystem can iterate in any order.


sgillies:
Thanks @QuLogic ! Could you see if this PR would be better aimed at the master branch? We're very close to stopping development on the maint-1.7 branch.

QuLogic:
Well, I kind of assumed you might merge up. I did make an earlier PR #427 to master but then later 1.7.x releases were made with the same error, so I targeted it instead. This does seem to be mostly applicable to master as well.

QuLogic:
I changed this to target `master`.

PR #620, opened 2018-08-23
Fix log statement causing UnicodeDecodeError in certain circumstances

Ariki:
In Python 2, when writing bytes data to a binary field, `UnicodeDecodeError` is possible if the field name is provided as a Unicode string, due to string formatting in `log.debug()`.

The fix doesn't solve problems with non-ASCII field names in Python 2, just allows to pass field name consisting of ASCII symbols as a Unicode string (which is convenient in case when the calling code uses `from __future__ import unicode_literals`).

The changes in test code are to make sure that all possible byte values are written correctly.

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -388,7 +388,7 @@ cdef class OGRFeatureBuilder:
                 set_field_null(cogr_feature, i)
             else:
                 raise ValueError("Invalid field type %s" % type(value))
-            log.debug("Set field %s: %s" % (key, value))
+            log.debug("Set field %s: %r" % (key, value))
         return cogr_feature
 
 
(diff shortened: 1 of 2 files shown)


fiona/ogrext.pyx
```diff
@@ -388,7 +388,7 @@ cdef class OGRFeatureBuilder:
                 set_field_null(cogr_feature, i)
             else:
                 raise ValueError("Invalid field type %s" % type(value))
-            log.debug("Set field %s: %s" % (key, value))
+            log.debug("Set field %s: %r" % (key, value))
```
sgillies: 👍 `%r` is better than `%s` for sure.

sgillies:
Thanks @Ariki !

PR #629, opened 2018-09-23
Port rasterio's env and session

sgillies:
I've copied Rasterio env, session, and path modules straight over and adjusted the Fiona modules and CLI. There's no attempt to test or use Google Cloud Storage, but it is at the very least not far from working.

Resolves #507

PR #631, opened 2018-09-24
Update test data bucket, GDAL win sdks

sgillies:
Also use `GDAL_HTTP_UNSAFESSL=YES` (set in the AppVeyor project environment).


sgillies:
The AppVeyor PR failure can be ignored here. All that will be fixed by #629 .

PR #634, opened 2018-10-01
gdal version macro < gdal 1.10

rbuffat:
With gdal 1.10.0 (Released in April 2013) the current "GDAL Version Information macro" was introduced. Older versions used a different version of the macro. The current version of Fiona does only follow the new macro definition. However, the documentation says that GDAL/OGR 1.8+ is supported. 
This PR makes calc_gdal_version_num and get_gdal_version_tuple aware of the different GDAL Version Information macro for OGR versions prior to 1.10.0. 

I'm not sure if there are users that use a recent version of Fiona in combination with such an old version of gdal/ogr.  Thus we could also just bump the minimal supported version of gdal to 1.10. 

However,  we currently do not test in the CI for backwards compability to the oldest supported version of gdal/ogr, which might be something that we should do (thus, this PR will not be tested by the CI).

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -100,8 +100,15 @@ def _bounds(geometry):
         return None
 
 def calc_gdal_version_num(maj, min, rev):
-    """Calculates the internal gdal version number based on major, minor and revision"""
-    return int(maj * 1000000 + min * 10000 + rev*100)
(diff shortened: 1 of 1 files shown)


sgillies:
@rbuffat Great! I don't see any harm in merging this.

PR #639, opened 2018-10-09
Remove env cruft, add env tests

### LINKED ISSUE #637: Remove unused arguments from Env constructor
See https://github.com/Toblerity/Fiona/blob/master/fiona/env.py#L101. These were left in when we imported this code from Rasterio, but aren't used.

sgillies:
Resolves #637 and more

### DIFF
--- a/fiona/compat.py
+++ b/fiona/compat.py
@@ -1,7 +1,6 @@
 import sys
 import collections
 
-# from six.moves import UserDict
 try:
     from collections import OrderedDict
 except ImportError:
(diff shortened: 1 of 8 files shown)


fiona/compat.py
```diff
@@ -1,7 +1,7 @@
 import sys
 import collections
 
-from six.moves import UserDict
+# from six.moves import UserDict
```
sgillies: TODO: remove this.

PR #643, opened 2018-10-10
Add get_linear_geometry to _shim, use it to convert curve geoms

### LINKED ISSUE #617: Convert nonlinear geometries to linear approximations by default
Since GDAL 2.2, we have to explicitly convert them. See https://github.com/Toblerity/Fiona/issues/492#issuecomment-407214987.

sgillies:
It is called right before we build a Fiona geometry.

Resolves #617

This first take leaks memory and creates an unnecessary copy. Will be improved in subsequent commits.


sgillies:
I will merge this one as soon as the checks are completed.

PR #648, opened 2018-10-20
Cut down on unittest usage

QuLogic:
This leaves 4 files with unittest:
$ rg -l 'import unittest'
test_collection.py
test_geometry.py
test_multiconxn.py
test_bytescollection.py

Also fixes finding of test data in the tests that don't use the fixtures because they used to be `unittest` but now aren't, and can subsequently use the fixture.


QuLogic:
Sorry it's so large, but if you tell GitHub to not show whitespace changes, it's not nearly as bad.

PR #653, opened 2018-10-21
Remove packaging from test requirements.

QuLogic:
Its only use is an import that is never used for anything.


sgillies:
Thanks!

PR #655, opened 2018-10-23
Drop unittest from collection tests

QuLogic:
With #654, this means there's no more unittest.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -5,6 +5,6 @@ base class.  This file can probably be deleted if that condition is removed.
 For example:
 
     $ git grep 'from \.' | grep test
-    tests/test_layer.py:from .test_collection import ReadingTest
-    tests/test_vfs.py:from .test_collection import ReadingTest
+    tests/test_layer.py:from .test_collection import TestReading
+    tests/test_vfs.py:from .test_collection import TestReading
 """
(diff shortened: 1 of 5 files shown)


sgillies:
🥇 @QuLogic

sgillies:
@QuLogic no blame from me, but this change is tripping up the wheel builds for macOS and Python 2.7: https://travis-ci.com/sgillies/fiona-wheels/jobs/154518129. I've very recently switched from developing on a MacBook to developing on a Linux laptop and that's why the issue wasn't caught before the tagged commit went to the wheel builder.

I've restarted the build job to check whether it was a transient error and it has failed again. Since all the other jobs have passed, I'm asking myself whether there is a tarfile bug specific to OS X on this version of Python 2.7.

I may have to revert the changes to TestTarArchiveReading until after we ship 1.8.0.

QuLogic:
That link is to a job that passed?

sgillies:
Oops, I forgot that I re-ran it after force pushing to the tag (questionable, I know) and now the history is lost.

QuLogic:
Can you try the build again so I can see what the error was?

sgillies:
@QuLogic check here in ~15 minutes: https://travis-ci.com/sgillies/fiona-wheels/builds/89538021.

QuLogic:
OK, well, I don't quite understand the failure there. I don't have a mac either, so I can't really debug much. If you add macOS CI, then I can try and re-do it.

Issue #553, opened 2018-03-11
Crash with pypy


kannes:
I tried running things with pypy and got crashes. They are weird and this is the minimal I got:

```
import fiona
import json

with fiona.open("ne_10m_admin_0_countries.shp") as source:
    countries = list(source)

with open("randomfilenameiuiubiqubeiquequze", "w") as sink:
    for country in countries:
        sink.write(json.dumps(country)+"\n")
```
The shapefile is from the latest http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/cultural/ne_10m_admin_0_countries.zip

Running this with pypy3 I get three different tracebacks, seemingly on random.

> RPython traceback:
>   File "rpython_jit_metainterp_1.c", line 41233, in assembler_call_helper_65
>   File "rpython_jit_metainterp_8.c", line 21619, in AbstractResumeGuardDescr_handle_fail
>   File "rpython_jit_metainterp_8.c", line 30933, in AbstractResumeGuardDescr__trace_and_compile_from
>   File "rpython_jit_metainterp_8.c", line 38152, in MetaInterp_handle_guard_failure
>   File "rpython_jit_metainterp_8.c", line 41192, in MetaInterp__handle_guard_failure
>   File "rpython_jit_metainterp_1.c", line 5322, in MetaInterp__interpret
>   File "rpython_jit_metainterp_2.c", line 22084, in MIFrame_run_one_step
>   File "rpython_jit_metainterp_4.c", line 13161, in handler_ref_return_1
>   File "rpython_jit_metainterp_4.c", line 61659, in MetaInterp_finishframe
> *** Invalid usage of a dying CPython object ***
> 
> cpyext, the emulation layer, detected that while it is calling
> an object's tp_dealloc, the C code calls back a function that
> tries to recreate the PyPy version of the object.  Usually it
> means that tp_dealloc calls some general PyXxx() API.  It is
> a dangerous and potentially buggy thing to do: even in CPython
> the PyXxx() function could, in theory, cause a reference to the
> object to be taken and stored somewhere, for an amount of time
> exceeding tp_dealloc itself.  Afterwards, the object will be
> freed, making that reference point to garbage.
> >>> PyPy could contain some workaround to still work if
> you are lucky, but it is not done so far; better fix the bug in
> the CPython extension.
> Aborted (core dumped)
> 

or

> RPython traceback:
>   File "rpython_jit_metainterp_2.c", line 50548, in handle_jitexception_68
[...]

PR #659, opened 2018-10-26
Don't call Python methods from __dealloc__()

Ariki:
Python methods should not be called from `__dealloc__()` because this may cause object resurrection.

Using `cpdef` directive generates a C version of the method along with Python version, thus solving the problem.

This fixes PyPy compatibility (#553).


sgillies:
That's all it takes @Ariki ? Thank you very much! I'm not going to add pypy to the build matrix yet, but will do so after 1.8.0.

PR #669, opened 2018-11-05
Eliminate log messages about ENCODING

### LINKED ISSUE #668: Prevent warnings about unsupported ENCODING creation option
The ENCODING creation option is only for Shapefiles. I'm not going to just silence the warnings, but filter the option out for all other format drivers.

sgillies:
Resolves #668

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -520,7 +520,7 @@ cdef class Session:
             OGR_L_TestCapability(
                 self.cogr_layer, OLC_STRINGSASUTF8) and
             'utf-8') or (
-            self.get_driver() == "ESRI Shapefile" and
+            "Shapefile" in self.get_driver() and
             'ISO-8859-1') or locale.getpreferredencoding().upper()
 
         if collection.ignore_fields:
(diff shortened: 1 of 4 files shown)


fiona/ogrext.pyx
```diff
@@ -520,7 +520,7 @@ cdef class Session:
             OGR_L_TestCapability(
                 self.cogr_layer, OLC_STRINGSASUTF8) and
             'utf-8') or (
-            self.get_driver() == "ESRI Shapefile" and
+            "Shapefile" in self.get_driver() and
```
sgillies: It occurred to me that we had a potential bug with respect to the driver name. GDAL allows "Shapefile" or "ESRI Shapefile" and we should handle either.

fiona/ogrext.pyx
```diff
@@ -996,7 +994,14 @@ cdef class WritingSession(Session):
             name_c = name_b
 
             for k, v in kwargs.items():
+
+                # We need to remove encoding from the layer creation
+                # options if we're not creating a shapefile.
+                if k == 'encoding' and "Shapefile" not in collection.driver:
+                    continue
+
```
sgillies: Here's the heart of the PR.

jorisvandenbossche:
@sgillies Thanks a lot for fixing those warnings!

Issue #551, opened 2018-03-11
GDAL 2.3 build issues: CPLSetErrorHandler and OGR_G_CreateGeometry


schwehr:
Pardon the confusion with this bug report.  I'm building with a custom bazel env and not using setup.py.  But I think these will hit many people using a GDAL pre 2.3 release (and probably needs a stricter compiler config to run into trouble)

Sadly, I don't know enough about cython to figure out a patch to include with this report

## Expected behavior and actual behavior.

Build failures...

I think that CPLErrorHandler needs a ctypedef for CPLSetErrorHandler

```
third_party/py/fiona/_drivers.pyx_cython_generated.cc:2174:3: error: no matching function for call to 'CPLSetErrorHandler'
  CPLSetErrorHandler(((void *)__pyx_f_5fiona_8_drivers_errorHandler));
  ^~~~~~~~~~~~~~~~~~
third_party/gdal/port/cpl_error.h:161:37: note: candidate function not viable: cannot convert argument of incomplete type 'void *' to 'CPLErrorHandler' (aka 'void (*)(CPLErr, int, const char *)') for 1st argument
CPLErrorHandler CPL_DLL CPL_STDCALL CPLSetErrorHandler( CPLErrorHandler );
```

and OGR_G_CreateGeometry can't cast an int to an enum.  I copied _shim22.* to _shim.* in my build

```
third_party/py/fiona/_geometry.pyx_cython_generated.cc:2460:27: error: no matching function for call to 'OGR_G_CreateGeometry'
  __pyx_v_cogr_geometry = OGR_G_CreateGeometry(__pyx_v_wkbtype);
                          ^~~~~~~~~~~~~~~~~~~~
third_party/gdal/ogr/ogr_api.h:87:22: note: candidate function not viable: no known conversion from 'unsigned char' to 'OGRwkbGeometryType' for 1st argument
OGRGeometryH CPL_DLL OGR_G_CreateGeometry( OGRwkbGeometryType ) CPL_WARN_UNUSED_RESULT;
                     ^
third_party/py/fiona/_geometry.pyx_cython_generated.cc:4822:27: error: no matching function for call to 'OGR_G_CreateGeometry'
  __pyx_v_cogr_geometry = OGR_G_CreateGeometry(__pyx_v_geom_type);
                          ^~~~~~~~~~~~~~~~~~~~
[...]

PR #672, opened 2018-11-05
Some type improvements

snorfalorpagus:
Some type improvements, reducing the number of warnings at compile time.

The handling of the `OGRFieldSubType` enum isn't quite right but doesn't cause any errors.

### DIFF
--- a/fiona/_drivers.pyx
+++ b/fiona/_drivers.pyx
@@ -18,7 +18,8 @@ cdef extern from "cpl_conv.h":
 
 
 cdef extern from "cpl_error.h":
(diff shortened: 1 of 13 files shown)


snorfalorpagus:
I've had a go at resolving #551. You can now force Fiona to build entirely in C++ by passing `--cython-language c++` as an argument to `setup.py`.

sgillies:
@snorfalorpagus you're awesome! What would you think about basing this on the new maint-1.8 branch?

snorfalorpagus:
Thanks @sgillies. I've rebased against `maint-1.8`.

PR #596, opened 2018-06-08
Fixed issue when using field names not supported by layer encoding

snorfalorpagus:
This PR fixes a bug triggered when creating a field with a name not supported by the layer encoding. Previously this would go unnoticed until a feature was written, at which point a `KeyError` is raised because the field is missing in the schema mapping. This change detects the issue at the time of creation and raises a `ValueError` with some useful information. I'd be willing to raise a more specific error if we can think of a name - `FieldCreationError`?

Closes #595.

PR #675, opened 2018-11-08
Fix bug in unicode test introduced in #596

snorfalorpagus:
I had assumed this was libiconv related, but it wasn't. The problem was with this part of the test:

```
with pytest.raises(SchemaError):
    fiona.open(os.path.join(self.tempdir, "test2.shp"), "w", **meta)
```

This correctly raises a SchemaError on Windows, but doesn't seem to close the GDAL file handle. The teardown method then tries to remove the shapefile but errors because a process is still accessing it.

The "fix" here is to use the pytest `tmpdir` fixture - which we should be using anyway. This works because it doesn't teardown the temporary directory immediately.

This is probably pointing to a more serious issue that we aren't closing the GDAL file handle when errors like this happen. The reason it's only showing on Windows is that Linux/macOS doesn't have the same "file in use" lock that Windows does.

### DIFF
--- a/tests/test_unicode.py
+++ b/tests/test_unicode.py
@@ -13,8 +13,7 @@ import fiona
 from fiona.errors import SchemaError
 
 
-class TestUnicodePath(object):
-
+class TestUnicodePath:
     def setup(self):
         tempdir = tempfile.mkdtemp()
         self.dir = os.path.join(tempdir, u'français')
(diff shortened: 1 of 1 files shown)


sgillies:
@snorfalorpagus changing `os.path.join(tmpdir, "test-write-gb18030.shp")` etc to `str(tmpdir.join("test-write-gb18030.shp"))` should fix this. Funny that it works with Python 3.

sgillies:
@snorfalorpagus check out https://ci.appveyor.com/project/sgillies/fiona/builds/20312337. I think this is solved. I'll re-run this PR after that one passes to check, this PR is worth having for sure.

sgillies:
Ack, we still have a single failure in maint-1.8 for GDAL versions >= 2.0. I think this only really is because of a defect in our GDAL binary. I'm not sure how to xfail this test other than testing the OS. I don't see that GDAL exposes whether it was configured with libiconv or not.

sgillies:
I resolved this one by skipping that test.

Issue #632, opened 2018-09-24
Fiona fails to read datasets containing TIN Z


rbuffat:
## Expected behavior and actual behavior.

Fiona can't open a dataset containing TIN Z geometries:

`~/swisstopo/musterdatenswissbuildings3d20lv03/LV03/SHP$ fio cat Musterdaten_swissBUILDINGS3D20_LV03_1166_44.shp
ERROR:fio:Exception caught during processing
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/fiona/fio/cat.py", line 77, in cat
    for i, feat in src.items(bbox=bbox):
  File "fiona/ogrext.pyx", line 1257, in fiona.ogrext.ItemsIterator.__next__
  File "fiona/ogrext.pyx", line 235, in fiona.ogrext.FeatureBuilder.build
  File "fiona/_geometry.pyx", line 176, in fiona._geometry.GeomBuilder.build
fiona.errors.UnsupportedGeometryTypeError: 16
Aborted!
`

GeometryType 16 is not included in https://github.com/Toblerity/Fiona/blob/ed1874e5d4ebaaed726146d3b074094e36307b22/fiona/_geometry.pyx#L36

I'm no expert in the geojson specification, but after a quick search it seems that geojson does not contain a tin datatype. However, it seems as tin is a special case of multipolygon (where with TIN each polygon is a triangle and has 4 vertices).

## Steps to reproduce the problem.

The dataset can be obtained from [https://shop.swisstopo.admin.ch/en/products/landscape/build3D2](https://shop.swisstopo.admin.ch/en/products/landscape/build3D2) -> Read more --> Additional Information --> Sample data --> swissBUILDINGS3D_2.0_LV03

## Operating system

Arch, respectively ubuntu

## Fiona and GDAL version and provenance

fiona 1.7.13 installed using pip
gdal 2.2.4

From my current understanding support for tin would require the introduction of  a mapping of triangle to polygon and tin to multipolygon, as both of these datatypes are not supported by geojson.
Hopefully I will find time during the next few days to further look into this.

snorfalorpagus:
Closed by #679.

PR #679, opened 2018-11-12
Rfc 64: support for Triangle, Polyhedral surface and TIN geometry types

rbuffat:
This PR adds support for Triangle, Polyhedral surface and TIN geometry types and solves https://github.com/Toblerity/Fiona/issues/632. As these types are not supported by GeoJSON, they are converted to MultiPolygon, respectively Polygon using the OGR_G_ForceTo* function as suggested [RFC64](https://trac.osgeo.org/gdal/wiki/rfc64_triangle_polyhedralsurface_tin), which was implemented with gdal 2.2.

> The OGRGeometryFactory::forceTo() and forceToMultiPolygon() methods have been enhanced to support conversions between the new geometry types, and towards multipolygon. Note that converting a TIN or a PolyhedralSurface? into a MultiPolygon? is semantically incorrect since a MultiPolygon? is suppose to contain geometries in the same plane, but it might help when converting those new geometry types into a format that doesn't support them (and such conversion was for example implicitly done in the reading side of the shapefile driver previously)

To use the OGR_G_ForceTo* methods, it was necessary to first clone the feature as otherwise, a segfault in _deleteOgrFeature() occurs. This is not nice and it should be reviewed if there is a better way.

Furthermore, this PR modifies the behavior introduced with the support for curves. This PR replaces "if 7 < code < 100:" with  "if 7 < code % 100 < 15" (https://github.com/Toblerity/Fiona/blob/master/fiona/ogrext.pyx#L307). Code 15,16, 17 belong to the Triangle, Polyhedral surface and TIN geometry types and according to [RFC49](https://trac.osgeo.org/gdal/wiki/rfc49_curve_geometries#Geometrytypes) curves with a Z dimension have codes bigger 1000.


rbuffat:
I think it would be better to use a modulo of 1000 instead of 100. I will update this tomorrow.

rbuffat:
I refactored the base type detection code of the GeomBuilder in an inline function base_geometry_type_code which should give consistent base type detection now.

It remains the issue with the need for cloning a feature when using OGR_G_Force.

tests/test_rfc64_tin.py
```diff
@@ -0,0 +1,27 @@
...
+from .conftest import  requires_gdal22
+
+
+def test_tin_shp(path_test_tin_shp):
+    """Convert TIN to MultiPolygon"""
+    with fiona.open(path_test_tin_shp) as col:
+        assert col.schema['geometry'] == 'Unknown'
+        features = list(col)
+        assert len(features) == 1
+        assert features[0]['geometry']['type'] == 'MultiPolygon'
```
snorfalorpagus: Should we also check the coordinates to see if the conversion has been done as expected?

rbuffat:
@snorfalorpagus As far as I understand the transformation should not impact the coordinates as these data types are basically special cases of MultiPolygon respectively Polygon. But you are right, it is better to check.

rbuffat:
I think it's getting there.

In order to avoid an unnecessary clone, we have first to steal the geometry from the feature! The drawback is that we are now taking the ownership of the geometry and need to take care to destroy it.

I adapted get_linear_geometry to work the same way as the force*Polygon methods so that input geometries, that are no longer used, get destroyed. This gives, in my opinion, a bit nicer looking code.

tests/test_rfc64_tin.py
rbuffat: Tests are extended to check for coordinates

fiona/ogrext.pyx
```diff
@@ -255,18 +255,26 @@ cdef class FeatureBuilder:
...
+            """ We steal the geometry: the geometry of the in-memory feature is now null
+            and we are responsible for cogr_geometry.
+            """
+            cogr_geometry = OGR_F_StealGeometry(feature)
 
             if cogr_geometry is not NULL:
-                code = OGR_G_GetGeometryType(cogr_geometry)
 
-                if 7 < code < 100:  # Curves.
+                code = base_geometry_type_code(OGR_G_GetGeometryType(cogr_geometry))
```
rbuffat: Is this acceptable or would it be better style to create a separate variable for OGR_G_GetGeometryType(cogr_geometry)?

sgillies: This is fine @rbuffat .

sgillies:
@rbuffat I'd like to see this in an upcoming 1.8.x release. If you merge the recent changes in fiona/master into your rfc64 branch, I believe I will be able to painlessly switch the target branch using the Github UI.

fiona/_shim22.pyx
```diff
@@ -117,5 +117,8 @@ cdef void set_field_subtype(void *fielddefn, OGRFieldSubType subtype):
 cdef bint check_capability_create_layer(void *cogr_ds):
     return GDALDatasetTestCapability(cogr_ds, ODsCCreateLayer)
 
-cdef void *get_linear_geometry(void *geom):
-    return OGR_G_GetLinearGeometry(geom, 0.0, NULL)
+cdef inline void *get_linear_geometry(void *geom):
+    cdef void *linear_geom
+    linear_geom = OGR_G_GetLinearGeometry(geom, 0.0, NULL)
+    OGR_G_DestroyGeometry(geom)
+    return linear_geom
```
sgillies: I don't think we need to steal the geometry in FeatureBuilder, and so the changes above should be reversed.

fiona/_shim22.pyx
```diff
@@ -117,5 +117,8 @@ cdef void set_field_subtype(void *fielddefn, OGRFieldSubType subtype):
 cdef bint check_capability_create_layer(void *cogr_ds):
     return GDALDatasetTestCapability(cogr_ds, ODsCCreateLayer)
 
-cdef void *get_linear_geometry(void *geom):
-    return OGR_G_GetLinearGeometry(geom, 0.0, NULL)
+cdef inline void *get_linear_geometry(void *geom):
+    cdef void *linear_geom
```
sgillies:
```suggestion
    cdef void *linear_geom = NULL
```

fiona/_shim22.pyx
```diff
@@ -117,5 +117,8 @@ cdef void set_field_subtype(void *fielddefn, OGRFieldSubType subtype):
...
+    linear_geom = OGR_G_GetLinearGeometry(geom, 0.0, NULL)
+    OGR_G_DestroyGeometry(geom)
```
sgillies:
```suggestion

```

It would be better for this function to not delete the given geometry.

fiona/ogrext.pyx
```diff
@@ -255,18 +255,26 @@ cdef class FeatureBuilder:
         cdef void *cogr_geometry = NULL
 
         if not ignore_geometry:
-            cogr_geometry = OGR_F_GetGeometryRef(feature)
+            """ We steal the geometry: the geometry of the in-memory feature is now null
+            and we are responsible for cogr_geometry.
+            """
```
sgillies: @rbuffat I now see that stealing the geometry is the proper thing to do. Please convert the comment to use `#` instead of triple quotes.

fiona/_shim22.pyx
sgillies: Summary: I think the function is best as it was originally written. Let's handle the memory allocation deallocation in the functions which call get_linear_geometry.

fiona/ogrext.pyx
```diff
@@ -255,18 +255,26 @@ cdef class FeatureBuilder:
...
+            cogr_geometry = OGR_F_StealGeometry(feature)
 
             if cogr_geometry is not NULL:
-                code = OGR_G_GetGeometryType(cogr_geometry)
 
-                if 7 < code < 100:  # Curves.
+                code = base_geometry_type_code(OGR_G_GetGeometryType(cogr_geometry))
+
+                if 7 < code < 15:  # Curves.
                     cogr_geometry = get_linear_geometry(cogr_geometry)
```
sgillies: Now we have two geometries in memory because OGR_G_GetLinearGeometry makes a copy (the difference from the "ForceTo" functions is confusing!). For the curve situation, let's do something like

```
cdef OGRGeometryH original_geometry = cogr_geometry
cogr_geometry = get_linear_geometry(original_geometry)
OGR_G_DestroyGeometry(original_geometry)
```

fiona/ogrext.pyx
```diff
@@ -255,18 +255,26 @@ cdef class FeatureBuilder:
...
-                else:
-                    geom = GeomBuilder().build(cogr_geometry)
+                elif code in (15, 16):  # RFC 64: Polyhedral surface and TIN
+                    cogr_geometry = OGR_G_ForceToMultiPolygon(cogr_geometry)
+
+                elif code == 17:  # RFC 64: Triangle
+                    cogr_geometry = OGR_G_ForceToPolygon(cogr_geometry)
+
+                geom = GeomBuilder().build(cogr_geometry)
+                OGR_G_DestroyGeometry(cogr_geometry)
```
sgillies: The changes above look great!

rbuffat:
@sgillies I reverted the changes in get_linear_geometry. I originally did it like this because I think it leads to a bit nicer looking code in ogrext, as get_linear_geometry then behaves similarly to OGR_G_ForceTo*.

I think the first appveyor fail is due to a communication error and the last one is that gdal master now points to 2.5

sgillies:
Okay, I will try to aim this at maint-1.8 now.

sgillies:
@rbuffat excellent! Can't wait to try this out.

Issue #692, opened 2018-12-07
A property named "time" raises DriverSupportError for ESRI Shapefiles


mdklatt:
## Expected behavior and actual behavior.

From https://github.com/Toblerity/Fiona/issues/572, I see that properties of *type* `datetime.time` are not supported for ESRI Shapefiles, but surely a `str` property *named* "time" is acceptable?

Fiona raises `DriverSupportError: ESRI Shapefile does not support time fields` for schemas that have a property named "time".

## Steps to reproduce the problem.

Running this code:

``` python
import fiona

schema = {
    "geometry": "Point",
    "properties": {
        "time": "str",  # type is str, should be okay
    }
}

c = fiona.collection("time_bug.shp", "w", "ESRI Shapefile", schema)
```

Gives this result:
``` console
Traceback (most recent call last):
  File "/vagrant/fiona_time_bug.py", line 13, in <module>
    c = fiona.collection("time_bug.shp", "w", "ESRI Shapefile", schema)
  File "/opt/DataPortal/lib/python3.6/site-packages/fiona/env.py", line 405, in wrapper
    return f(*args, **kwargs)
  File "/opt/DataPortal/lib/python3.6/site-packages/fiona/__init__.py", line 263, in open
    **kwargs)
  File "/opt/DataPortal/lib/python3.6/site-packages/fiona/collection.py", line 145, in __init__
    self._check_schema_driver_support()
  File "/opt/DataPortal/lib/python3.6/site-packages/fiona/collection.py", line 419, in _check_schema_driver_support
    raise DriverSupportError("ESRI Shapefile does not support time fields")
fiona.errors.DriverSupportError: ESRI Shapefile does not support time fields

```

## Operating system

Linux-3.10.0-514.el7.x86_64-x86_64-with-centos-7.3.1611-Core
Python 3.6.5

## Fiona and GDAL version and provenance

Installed from PyPI via pip:

Fiona 1.8.3
GDAL 2.3.2

mdklatt:
The source of the exception is [collection.py](https://github.com/Toblerity/Fiona/blob/master/fiona/collection.py#L418):

``` python
def _check_schema_driver_support(self):
        """Check support for the schema against the driver

        See GH#572 for discussion.
        """
        gdal_version_major = get_gdal_version_num() // 1000000
        for field in self._schema["properties"]:
            field_type = field.split(":")[0]
            if self._driver == "ESRI Shapefile":
                if field_type == "datetime":
[...]

PR #694, opened 2018-12-09
Fix shp time

rbuffat:
@sgillies I hope you don't mind. I implemented the fix for https://github.com/Toblerity/Fiona/issues/692 proposed by @snorfalorpagus  and replaced "get_gdal_version_num() // 1000000" as this will fail for gdal versions older than 1.10.0.

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -148,10 +148,8 @@ def test_unsupported_geometry_type():
             'geometry': 'BOGUS',
             'properties': {}}}
 
-    try:
+    with pytest.raises(UnsupportedGeometryTypeError):
         fiona.open(tmpfile, 'w', **profile)
-    except UnsupportedGeometryTypeError:
-        assert True
 
 
 @pytest.mark.parametrize('x', list(range(1, 10)))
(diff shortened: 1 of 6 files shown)


tests/test_schema.py
```diff
@@ -197,3 +197,37 @@ def test_normalize_std(x):
...
+            items = [('field1',  field_type)]
+            with fiona.open(name, 'w',
+                            driver="ESRI Shapefile",
+                            schema={
+                                'geometry': 'LineString',
+                                'properties': items}) as c:
+                assert False
+            
+        except DriverSupportError:
+            assert True
```
sgillies: @rbuffat would you be willing to change the above in two ways? 1) the several field types could be done using pytest.mark.parametrize https://docs.pytest.org/en/latest/parametrize.html, which creates separate tests for each field type, and 2) use `with pytest.raises(DriverSupportError)` to assert that an exception is raised.

tests/test_schema.py
```diff
@@ -197,3 +197,37 @@ def test_normalize_std(x):
...
+        items = [('field1',  'time')]
+        with fiona.open(name, 'w',
+                        driver="GPKG",
+                        schema={
+                            'geometry': 'LineString',
+                            'properties': items}) as c:
+            assert False
+        
+    except DriverSupportError:
+        assert True
```
sgillies: Similarly, I'd like to see the geopackage case become its own proper test separate from the shapefile test.

tests/test_schema.py
rbuffat: I refactored the test using pytest.raises and parametrize

tests/test_schema.py
rbuffat: Refactored to a seperate test case

rbuffat:
@sgillies  Done. I also refactored test_schema_ordering_items to use pytest.raises.

Strange that the coverage decreases with the additional tests. Especially that the now missing line should not be covered by tests previously as we don't test for old gdal 1.x versions.

sgillies:
@rbuffat I don't have 100% trust in the math at coveralls.io. Their issue tracker has a number of similar reports.

Would you be willing to rebase your changes on the maint-1.8 branch? I wish I had spotted this before. Sorry!

rbuffat:
@sgillies could you give me some pointers on how to do the rebase correctly? "git rebase upstream/maint-1.8" did not really do what I intended

PR #696, opened 2018-12-11
[WIP] Compatibility with gdal 1.8

rbuffat:
We claim that we are compatible with GDAL/OGR 1.8+, however, we don't test against gdal 1.8.

I enabled gdal 1.8 on travis and there are quite a few tests that are failing. I suspect in many cases tests are not aware of newer gdal features not present in version 1.8.

My plan is to go through the test cases and adapt the tests if needed. This will take some time. I suspect not many people are using very old versions of gdal anymore, therefore it is also not really time critical.

I created this PR in order to track the failing tests visibly on travis. I could also just open an issue if this is preferred.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ env:
     - GDALINST=$HOME/gdalinstall
     - GDALBUILD=$HOME/gdalbuild
   matrix:
+    - GDALVERSION="1.8.0"
     - GDALVERSION="1.11.5"
     - GDALVERSION="2.0.3"
     - GDALVERSION="2.1.4"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: python
 
-dist: xenial
-
 cache:
   directories:
     - $GDALINST
@@ -15,27 +13,19 @@ env:
     - GDALBUILD=$HOME/gdalbuild
 
   matrix:
-    - GDALVERSION="1.8.0"
     - GDALVERSION="1.11.5"
-    - GDALVERSION="2.0.3"
+    - GDALVERSION="2.1.4"
     - GDALVERSION="2.3.2"
-    - GDALVERSION="trunk"
-
-matrix:
-  fast_finish: true
-  allow_failures:
-    - env: GDALVERSION="1.8.0"
-    - env: GDALVERSION="trunk"
     
-  exclude:
-    - env: GDALVERSION="1.8.0"
-      python: "3.6"
-    - env: GDALVERSION="trunk"
-      python: "2.7"
-
 python:
   - "2.7"
   - "3.6"
+
+matrix:
+  fast_finish: true
+  
+allow_failures:
+  - env: GDALVERSION="trunk"
 
 addons:
   apt:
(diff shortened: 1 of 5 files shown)


sgillies:
@rbuffat Adding 1.8 to the build matrix slows down our builds for very little gain. I would prefer that we update the documentation to say that 1.8 may work but has no official support any longer. @snorfalorpagus yes or no?

rbuffat:
@sgillies I agree with the concern regarding the build matrix. However, without testing, we just now for sure that fiona works against the oldest gdal 1.x we test for, which is I think the latest available. Maybe a compromise would be to have one gdal 1.8 / python 2 only test on appveyor. (If there are gisinternal binaries for gdal 1.8 available). One test on appveyor needs around 2.5 minutes.

rbuffat:
testing gdal 1.8 on appveyor seems not straightforward possible as there are is no *libs.zip available.

I looked at the travis documentation and found some interesting options, such as stages. I played a bit around with them. I don't mind if something of this is used at end or not!

The idea with stages would be to have two testing stages. The first stage tests the most important versions of gdal (like latest gdal version for gdal 1.x and 2.x). The second stage is only executed if all tests of the first stage completed successfully and could contain a more comprehensive coverage of gdal versions.  This would reduce the time until a build fails, however, is likely to increase the overall runtime when everything runs successfully, as no jobs of the second stage start before the first starts.

These stages can also be subject to conditions: https://docs.travis-ci.com/user/conditions-v1
Thus, the second stage could be configured to only be run towards a new release based on a commit message or a tag or run regularly using a cron job.

rbuffat:
After experimenting a bit I would propose the following travis setup:

- First stage: (default matrix): latest gdal version for each shim (current 1.11.5, 2.1.4 and 2.3.2)
- Stage 2: “Legacy” versions of gdal that are not tested in the first stage. This stage is only tested if the commit message contains “ci_legacy” or “ci_all”
- Stage 3: “trunk” tests the trunk version of gdal.  This stage is activated for the master branch or when the commit message contains “ci_gdaltrunk” or “ci_all”


Per default, the second test stage is not tested. It needs to be invoked manually via the commit message. I think this is a compromise between efficient testing while still having the possibility to test easily against a wide range of gdal versions.

Furthermore, I optimized a bit the travis setup in order to save some time:
- No ubuntu gdal binaries are installed.
- Pip install instructions were optimized to not reinstall previously installed libraries
- There were two pip install fiona commands, I think only one of them is necessary
- Only a depth of 3 is used to clone the fiona git repository
- Travis_gdal_install.sh delete the gdal source directory after installation, this should reduce a bit the travis cache size
- Travis_gdal_install.sh adding some more gdal build options to reduce build time

Gdal 2.3.x fails currently because the cached version of gdal requires libwebp, which is now deactivated (--with-webp=no) and the library not installed.  To solve this the travis cache need to be cleared: https://docs.travis-ci.com/user/caching/#clearing-caches

sgillies:
@rbuffat sorry for the delay on this PR, I'm tied up with work and fixing Rasterio's wheel building infrastructure. I like what I read about the streamlining of the builds!

rbuffat:
I'm closing this PR as most of the changes will be obsolete with Fiona 2.0.  I did some experiments with travis, which could also be beneficial for Fiona 2.0. I will open a new PR specific for travis changes at a later time.

PR #709, opened 2019-01-17
Add google cloud storage support

theduckylittle:
Adds the appropriate prefix to support Google Cloud Storage.

Requires support in GDAL (2.3 or newer).


theduckylittle:
Replaces https://github.com/geomoose/gm3/pull/345

PR #712, opened 2019-01-21
WIP: Added Python 3.7 to Appveyor

snorfalorpagus:
This PR adds a build using Python 3.7 to Appveyor. I've also cleaned the config up a bit.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,39 +11,52 @@ environment:
         # See: http://stackoverflow.com/a/13751649/163740
         CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\appveyor\\run_with_env.cmd"
         GDAL_HOME: "C:\\gdal"
+        PYTHON_ARCH: "64"
 
     matrix:
+        # The 4-digit number in the GISInternals archives is the MSVC version used to build
+        # the libraries. It does not need to match the version of MSVC used to build Python.
+        # https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#Internal_version_numbering
+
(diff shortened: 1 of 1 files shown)


sgillies:
@snorfalorpagus I'm not sure what to make of that first error. Should I re-run it?

snorfalorpagus:
@sgillies It looks like a corrupt download of the gisinternals binaries. If you could re-run it my guess is it'll work.

My aim with this is to get Python 3.6 + 3.7 wheels for Windows that could be uploaded to PyPI.

sgillies:
I re-built and it failed to extract the zip file again.

snorfalorpagus:
I see what has gone wrong. That combination of MSVC/GDAL does not exist. I'll fix it now.

snorfalorpagus:
One solution to the 2.7 issue is to use:

```
python -W ignore -m pip install ...
```

However it's used in several places and makes everything ugly for the sake of supporting Python 2.7 for a few more months.

This can also be done with an environment variable, but we'd want to unset it before running the tests as we still want to know about warnings in Fiona.

snorfalorpagus:
Another temporary solution is to pin `pip` at an earlier version, before the deprecation warning was introduced.

snorfalorpagus:
The issue seems to be that Appveyor is detecting the warning as an error because it's printed to stderr. Not sure how to fix this yet.

snorfalorpagus:
I think this is ready for another look @sgillies.

I've switched the pip commands to use `cmd` instead of `ps` which prevents the deprecation error from causing the build to fail.

Issue #713, opened 2019-01-24
[proposal] Add support for metadata read/write for GeoPackage


snowman2:
One feature that I think would be particularly useful is reading/writing metadata in GeoPackage files. I have seen that this is supported in [gdal python tests](https://trac.osgeo.org/gdal/browser/trunk/autotest/ogr/ogr_gpkg.py#L1653). However, I don't see it in the Fiona package. If it is already there and I missed it, feel free to point it out to me.

Is this a feature that you would be interested in? If so, I would be willing to create a prototype of a potential solution we could iterate on.

sgillies:
@snowman2 yes, this would be a fine addition. I think the tags and get_tag_item methods in rasterio._base.DatasetBase could be a useful example.

snowman2:
Great, I will try to make time for this.

PR #715, opened 2019-01-28
initial support for geopackage metatata

snowman2:
issue #713

@sgillies here is the first iteration.


snowman2:
Strange, it says [here](https://www.gdal.org/drv_geopackage_raster.html) that the metadata has been supported since GDAL 1.11, but that seems to be the only one failing.

snowman2:
Hmmm, any ideas on this error:
```
E   fiona.errors.DriverError: sqlite3_open(/vsimem//tmp/pytest-of-travis/pytest-0/test_set_tag_item__namespace_N0/test.gpkg) failed: out of memory
```

snowman2:
GDAL 1.11 is the only one that fails to get/set metadata correctly. How would you like that handled?

fiona/ogrext.pyx
```diff
@@ -1210,6 +1283,85 @@ cdef class WritingSession(Session):
...
+            obj = self.cogr_layer
+        else:
+            obj = self.cogr_ds
+
+        if ns:
+            ns = ns.encode('utf-8')
+            domain = ns
+
+        cdef char **metadata = NULL
+        cdef char *metatata_item_c = NULL
```
snorfalorpagus: This appears to be unused.

fiona/collection.py
```diff
@@ -221,6 +221,87 @@ def crs_wkt(self):
             self._crs_wkt = self.session.get_crs_wkt()
         return self._crs_wkt
 
+    def tags(self, ns=None):
```
snorfalorpagus: If rasterio had not already set a precedent I would have suggested using `namespace=` instead of `ns=`.

snorfalorpagus:
Despite the documentation it seems to me that GDAL 1.x doesn't support GPKG metadata:

```python
>>> from osgeo import ogr
>>> driver = ogr.GetDriverByName("GPKG")
>>> ds = driver.CreateDataSource("new.gpkg")
>>> ds.SetMetadataItem("foo", "bar")
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/osgeo/ogr.py", line 577, in <lambda>
    __getattr__ = lambda self, name: _swig_getattr(self, DataSource, name)
  File "/usr/lib/python3/dist-packages/osgeo/ogr.py", line 74, in _swig_getattr
    return _swig_getattr_nondynamic(self, class_type, name, 0)
  File "/usr/lib/python3/dist-packages/osgeo/ogr.py", line 69, in _swig_getattr_nondynamic
    return object.__getattr__(self, name)
AttributeError: type object 'object' has no attribute '__getattr__'
```

fiona/ogrext.pyx
snowman2: Nice catch! Leftovers from a previous iteration.

fiona/collection.py
snowman2: I had the same thought as I was going through this. Do you think it would be terrible to use `namespace=` instead?

snowman2:
>Despite the documentation it seems to me that GDAL 1.x doesn't support GPKG metadata:

That's unfortunate. Maybe adding some kind of require GDAL 2.x function would be the solution here?

snowman2:
I was looking at the metadata on the raster geopackage that pinned metadata to 1.11. The vector [geopackage](https://www.gdal.org/drv_geopackage.html) docs say GDAL >= 2.

fiona/collection.py
sgillies: "ns" is not uncommon shorthand for namespace in the XML domain. "xmlns" for example, or the "start-ns" and "end-ns" events in the ElementTree API. I don't think there's any harm in using "ns" here as in Rasterio.

fiona/env.py
```diff
@@ -538,7 +538,6 @@ def some_func(foo=None):
     reason = '\n{0}'.format(reason) if reason else reason
 
     def decorator(f):
-        @wraps(f)
```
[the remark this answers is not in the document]
sgillies: `wraps` wasn't breaking anything before. Can you explain what has changed @snowman2 ?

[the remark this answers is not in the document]
snowman2: The difference here is that it is being used on a class method instead of a function. It works fine for functions. However, when being applied to a class method, it breaks in Python 2 [see build logs](https://travis-ci.org/Toblerity/Fiona/builds/485700774) This issue was resolved in Python 3.2 [see here](https://bugs.python.org/issue3445). Definitely open to ideas for a better solution.

sgillies:
I've finally made some time to get to this! Here's a test that was failing before the `require_gdal_version` decorator was added. 

```
_________________________ test_set_tag_item[test-test] _________________________
layer = 'test', namespace = 'test'
tmpdir = local('/tmp/pytest-of-travis/pytest-0/test_set_tag_item_test_test_0')
    @pytest.mark.parametrize("layer, namespace", [
        (None, None),
        (None, "test"),
        ("test", None),
        ("test", "test"),
    ])
    def test_set_tag_item(layer, namespace, tmpdir):
        test_geopackage = str(tmpdir.join("test.gpkg"))
        schema = {'properties': {'CDATA1': 'str:254'}, 'geometry': 'Polygon'}
        with fiona.open(test_geopackage, "w", driver="GPKG", schema=schema, layer=layer) as gpkg:
            assert gpkg.get_tag_item("test_tag1", ns=namespace) is None
            gpkg.set_tag_item("test_tag1", "test_value1", ns=namespace)
    
        with fiona.open(test_geopackage, layer=layer) as gpkg:
            if namespace is not None:
                assert gpkg.get_tag_item("test_tag1") is None
>           assert gpkg.get_tag_item("test_tag1", ns=namespace) == "test_value1"
E           AssertionError: assert None == 'test_value1'
E            +  where None = <bound method Collection.get_tag_item of <open Collection '/tmp/pytest-of-travis/pytest-0/test_set_tag_item_test_test_0/test.gpkg:test', mode 'r' at 0x7fabc2bb87b8>>('test_tag1', ns='test')
E            +    where <bound method Collection.get_tag_item of <open Collection '/tmp/pytest-of-travis/pytest-0/test_set_tag_item_test_test_0/test.gpkg:test', mode 'r' at 0x7fabc2bb87b8>> = <open Collection '/tmp/pytest-of-travis/pytest-0/test_set_tag_item_test_test_0/test.gpkg:test', mode 'r' at 0x7fabc2bb87b8>.get_tag_item
tests/test_session.py:54: AssertionError
----------------------------- Captured stderr call -----------------------------
ERROR 1: sqlite3_prepare_v2(SELECT Count(*) FROM test WHERE test_tag1) failed: no such column: test_tag1
ERROR 1: failed to prepare SQL: SELECT fid, "geom", "CDATA1" FROM test WHERE test_tag1
```

The method returns `None`. If we had an `Env` block around the calls, we would get an exception from ogrext.pyx. Instead, the errors are printed to stderr.

I think it would be fine to remove the decorator and let this method raise an exception with GDAL 1.11. We'll document the situation and will be leaving 1.x behind sooner or later. It shouldn't block us from moving ahead.

@snorfalorpagus @snowman2 Yes? No?

snowman2:
Sounds like a good plan to me.

snowman2:
@sgillies I removed the commits with the decorators.

snowman2:
Should I add the env block in the tests?

sgillies:
@snowman2 yes, please. And, if you like: catch the exception coming from fiona.ogrext with GDAL 1.11 and xfail the test inside the except clause using pytest.xfail.

snowman2:
xfail coming after I the next tests fail so I have the specific exception it raises.

snowman2:
@sgillies, it does not seem to raise an exception. How would you like to proceed?

sgillies:
@snowman2 I'm surprised at the lack of an exception, but for now, let's test the dict and xfail if it is empty and if the GDAL version is less than 2.0.

sgillies:
Congratulations @snowman2 ! I'm going to give this a little more review on the weekend, focusing on what to return or raise when the collection isn't open (no session or whatever). Generally, this is great stuff.

snowman2:
Just squashed it. :+1:

joehuanguf:
Is there still interest in this PR? I can take a stab at continuing it.

sgillies:
@joehuanguf yes, this is just held up by a refactor and my involvement in GDAL 3 fallout. We're still going to do this.

fiona/collection.py
```diff
@@ -221,6 +221,87 @@ def crs_wkt(self):
...
+        ns: str, optional
+            Used to select a namespace other than the default.
+
+        Returns
+        -------
+        int
+        """
+        if isinstance(self.session, WritingSession):
+            return self.session.set_tags(tags, ns=ns)
+        raise RuntimeError("Unable to set tags as not in writing mode.")
```
sgillies: @snowman2 here I think we should begin with 

```
if not isinstance(self.session, WritingSession):
    raise ...
```

and then I think we should consider a custom error. If we open a Python (3) file to read and then call its write method we get an io.UnsupportedOperation. We should add something like that to fiona.

sgillies:
Finally getting back to this. It's still relevant and good and I think it should be able to merge fairly cleanly with maint-1.9.

snowman2:
Didn't see dropdown to change branches. Moved to #821

fiona/collection.py
snowman2: Have a preference for the exception name?

Issue #685, opened 2018-11-17
Writing performance


jorisvandenbossche:
There has been a previous issue about the slow writing of GeoPackage files: https://github.com/Toblerity/Fiona/issues/476. But triggered by https://gis.stackexchange.com/questions/302811/how-to-get-fast-writing-with-geopandas-fiona, I was further looking into it with the latest versions of GeoPandas and Fiona, and it still seems relatively slow. 

It already improved a lot compared with the previous versions of both GeoPandas and Fiona. And of the remaining time, GeoPandas takes the most time (which I will try to fix, cfr https://github.com/geopandas/geopandas/issues/863). But even then, writing a file with 100k rows and 5 attribute columns takes ca 10s with Fiona. 

Sample set-up:
```
import pandas as pd
import geopandas
import fiona
import shapely.geometry

import random
import string

N = 100000

df = geopandas.GeoDataFrame(
    {'a': np.random.randn(N), 'b': np.random.randn(N),
     'c': np.random.randn(N), 'd': np.random.randint(100, size=N),
     'e': [''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(5)) for _ in range(N)],
     'geometry': [shapely.geometry.Point(random.random(), random.random()) for _ in range(N)]})

records = list(df.iterfeatures())
schema = geopandas.io.file.infer_schema(df)

with fiona.Env():
    with fiona.open("test_geopackage.gpkg", 'w', driver="GPKG", schema=schema) as colxn:
        colxn.writerecords(records)
```

Timing only the fiona-writing part (using Fiona 1.8.1, GDAL 2.3 with Python 3.6 on Ubuntu 16.04, installed with conda-forge):
```
In [37]: %%timeit
    ...: with fiona.Env():
    ...:     with fiona.open("test_geopackage_profile2.gpkg", 'w', driver="GPKG", schema=schema) as colxn:
    ...:         colxn.writerecords(records)
11.4 s ± 284 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Full example notebook exploring the performance of writing that dataframe to GPKG: http://nbviewer.jupyter.org/gist/jorisvandenbossche/c8590a3617698befad527e66eefb7f5b

As comparison, writing the same dataset with QGIS takes only a couple of seconds, and reading / writing it with `ogr2ogr` also takes less time (around 5s for both reading and writing it).

So the main question I am wondering: would there still be ways to improve this in Fiona? 

[...]

PR #718, opened 2019-02-03
remove per-feature logging to improve performance issue#685

snowman2:
I saw the issue #685, and so I took a stab at it based on the discussion to see if I could help move things forward.

I commented out the debug print statements instead of deleting them so that you don't have to re-create them when you are debugging (and an official decision on what to do with them has not been reached as far as I know). However, that presents the danger of being re-introduced by accident when committing changes after debugging.

Using this example:
```python
from itertools import chain, repeat
import fiona
def test_to_file():
    with fiona.Env(),fiona.open("../tests/data/coutwildrnp.shp") as collection:
        features = chain.from_iterable(repeat(list(collection), 2000))
        with fiona.open("/tmp/out.gpkg", "w", schema=collection.schema, crs=collection.crs, driver="GPKG") as dst:
            dst.writerecords(features)
```
And timing it:
```python
%%timeit
test_to_file()
```

Before:
1min 3s ± 2.32 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
After:
15.3 s ± 1.13 s per loop (mean ± std. dev. of 7 runs, 1 loop each)

### DIFF
--- a/fiona/_geometry.pyx
+++ b/fiona/_geometry.pyx
@@ -283,7 +283,7 @@ cdef class OGRGeomBuilder:
             coordinates = geometry.get('geometries')
             return self._buildGeometryCollection(coordinates)
         else:
-            raise ValueError("Unsupported geometry type %s" % typename)
+            raise UnsupportedGeometryTypeError("Unsupported geometry type %s" % typename)
 
 
 def geometryRT(geometry):
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -258,7 +258,6 @@ cdef class FeatureBuilder:
                 props[key] = data[:l]
 
             else:
-                # log.debug("%s: None, fieldtype: %r, %r" % (key, fieldtype, fieldtype in string_types))
                 props[key] = None
 
         cdef void *cogr_geometry = NULL
@@ -318,13 +317,10 @@ cdef class OGRFeatureBuilder:
         encoding = session.get_internalencoding()
 
         for key, value in feature['properties'].items():
-            # log.debug(
-            #     "Looking up %s in %s", key, repr(session._schema_mapping))
             ogr_key = session._schema_mapping[key]
 
             schema_type = normalize_field_type(collection.schema['properties'][key])
 
-            # log.debug("Normalizing schema type for key %r in schema %r to %r", key, collection.schema['properties'], schema_type)
             key_bytes = strencode(ogr_key, encoding)
             key_c = key_bytes
             i = OGR_F_GetFieldIndex(cogr_feature, key_c)
@@ -337,9 +333,6 @@ cdef class OGRFeatureBuilder:
 
             # Continue over the standard OGR types.
             if isinstance(value, integer_types):
-
-                # log.debug("Setting field %r, type %r, to value %r", i, schema_type, value)
-
                 if schema_type == 'int32':
                     OGR_F_SetFieldInteger(cogr_feature, i, value)
                 else:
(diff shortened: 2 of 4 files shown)


snowman2:
Looks like the GDAL trunk is using PROJ.6 now: `configure: error: PROJ 6 symbols not found`

sgillies:
@snowman2 wow, great! I think I'll be able to have some review time tomorrow.

snowman2:
Another idea I had recently was change:
https://github.com/Toblerity/Fiona/blob/013dec95f3a468459a76d57b60f676b415bda738/fiona/ogrext.pyx#L320

To instead iterate over:
```python
collection.schema['properties']
```
And set the property to use:
```python
set_field_null(cogr_feature, i)
```
If the property does not exist in `feature['properties']`.

This would eliminate the need for this check:

https://github.com/Toblerity/Fiona/blob/013dec95f3a468459a76d57b60f676b415bda738/fiona/ogrext.pyx#L1167

But, I wanted to run the idea by you first.

fiona/_geometry.pyx
```diff
@@ -283,7 +283,7 @@ cdef class OGRGeomBuilder:
             coordinates = geometry.get('geometries')
             return self._buildGeometryCollection(coordinates)
         else:
-            raise ValueError("Unsupported geometry type %s" % typename)
+            raise UnsupportedGeometryTypeError("Unsupported geometry type %s" % typename)
```
sgillies: 👍

fiona/ogrext.pyx
```diff
@@ -318,19 +318,14 @@ cdef class OGRFeatureBuilder:
...
 
-            log.debug("Normalizing schema type for key %r in schema %r to %r", key, collection.schema['properties'], schema_type)
-
-            try:
-                key_bytes = ogr_key.encode(encoding)
-            except UnicodeDecodeError:
-                log.warning("Failed to encode %s using %s codec", key, encoding)
-                key_bytes = ogr_key
+            # log.debug("Normalizing schema type for key %r in schema %r to %r", key, collection.schema['properties'], schema_type)
+            key_bytes = strencode(ogr_key, encoding)
```
sgillies: 👍

fiona/ogrext.pyx
```diff
@@ -1162,23 +1144,22 @@ cdef class WritingSession(Session):
...
-
+                        record['geometry']['type'],
+                        collection.schema['geometry'] ))
+            # log.debug("Creating feature in layer: %s" % record)
             cogr_feature = OGRFeatureBuilder().build(record, collection)
             result = OGR_L_CreateFeature(cogr_layer, cogr_feature)
             if result != OGRERR_NONE:
+                # Validate against collection's schema to give useful message
+                if set(record['properties'].keys()) != schema_props_keys:
+                    raise ValueError(
```
sgillies: Maybe raise SchemaError here?

sgillies:
@snowman2 what would you think about holding off on the schema/property checks for now? I think it's separable from elimination of the debug statements, which is a big win by itself.

I'm inclined to remove the debug statements that you've commented. I put them in when building the original feature and I don't think they are necessarily needed anymore.

snowman2:
> I'm inclined to remove the debug statements that you've commented.

I removed those in a separate commit.

> what would you think about holding off on the schema/property checks for now?

I checked the difference between checking the properties and the alternative method and I was surprised that it didn't make a significant difference. So, I agree that it is an unnecessary change. It makes me wonder if it would be a good idea to do a check and remove all unnecessary log.debug/log.info statements in rasterio & fiona for performance reasons?

sgillies:
@snorfalorpagus I tagged you for review of this, too.

sgillies:
I edited the PR title to reflect our decision to take a one thing at a time approach to optimization.

PR #728, opened 2019-03-06
Issue726 1.8

### LINKED ISSUE #726: Attribute value encoding issue when reading a Shapefile

Loading a properly encoded (from what I see) Shapefile results in encoding garbage.

https://ec.europa.eu/eurostat/cache/GISCO/distribution/v2/nuts/download/ref-nuts-2016-01m.shp.zip includes a NUTS_RG_01M_2016_3035.shp.zip which includes NUTS_RG_01M_2016_3035.shp

There *is* a .cpg file with `UTF-8` inside.

The feature with FID RS1 or at Fiona's resulting index 123 is good for testing.

```
In [1]: import fiona                                                            

In [2]: with fiona.open("NUTS_RG_01M_2016_3035.shp") as source: 
   ...:     print(source[123]['properties']['NUTS_NAME']) 
   ...:                                                                         
Ð¡Ñ\x80Ð±Ð¸Ñ\x98Ð° - Ñ\x81ÐµÐ²ÐµÑ\x80
```

Testing with GDAL shows no issues and correct encodings:
```
$ ogrinfo -al NUTS_RG_01M_2016_3035.shp | grep -B 1 RS1 
...
  NUTS_NAME (String) = Србија - север
  FID (String) = RS1
...
```

It also works fine if I force Fiona to UTF-8:

```
In [1]: import fiona                                                                                         

In [2]: with fiona.open("NUTS_RG_01M_2016_3035.shp", encoding="UTF-8") as source: 
   ...:     print(source[123]['properties']['NUTS_NAME']) 
   ...:                                                                                                      
Србија - север

```

## Operating system

Archlinux

`pip install --user --upgrade fiona` gave me https://files.pythonhosted.org/packages/b8/55/fdcd9a45b39a4fe50d79041d1194c2ebefcc7ef44958d54ad2b815129f02/Fiona-1.8.4-cp37-cp37m-manylinux1_x86_64.whl and that's where I reproduced this bug.

GDAL 2.3.2, released 2018/09/21 installed from the Archlinux repositories.

sgillies:
We're making more use of the OLCStringsAsUTF8 capability marker and eliminating some duplication of logic. 

Resolves #726, which reported that a .cpg file was being overlooked.

In the process, I discovered that the coutwildrnp.shp lacked a valid code page. I've added a ISO-8859-1 .cpg file for it.


fiona/_shim2.pyx
```diff
@@ -46,6 +47,10 @@ cdef void* gdal_open_vector(const char* path_c, int mode, drivers, options) exce
                 drvs = CSLAddString(drvs, name_c)
 
     for k, v in options.items():
+
+        if v is None:
+            continue
+
```
sgillies: Skip invalid options

fiona/collection.py
```diff
@@ -150,7 +150,7 @@ def __init__(self, path, mode='r', driver=None, schema=None, crs=None,
                     raise CRSError("crs lacks init or proj parameter")
 
         self._driver = driver
-        kwargs.update(encoding=encoding or '')
```
sgillies: Empty string is not needed since we now skip invalid options.

fiona/collection.py
```diff
@@ -166,8 +166,6 @@ def __init__(self, path, mode='r', driver=None, schema=None, crs=None,
 
         if self.session is not None:
             self.guard_driver_mode()
-            if not self.encoding:
-                self.encoding = self.session.get_fileencoding().lower()
```
sgillies: No longer needed.

fiona/ogrext.pyx
```diff
@@ -525,16 +521,53 @@ cdef class Session:
...
+        capability marker, it is not possible to know exactly what the
+        internal encoding is and this method returns best guesses. That
+        means ISO-8859-1 for shapefiles and the locale's preferred
+        encoding for other formats such as CSV files.
+
+        """
+        if OGR_L_TestCapability(self.cogr_layer, OLC_STRINGSASUTF8):
+            return 'utf-8'
+        else:
+            return self._fileencoding or self._get_fallback_encoding()
```
sgillies: This function returns the name of the codec we now use when going between Python and OGR field names and values.

tests/test_collection.py
```diff
@@ -115,8 +115,8 @@ def test_name(self):
     def test_mode(self):
         assert self.c.mode == 'r'
 
-    def test_collection(self):
-        assert self.c.encoding == 'iso-8859-1'
+    def test_encoding(self):
+        assert self.c.encoding is None
```
sgillies: GDAL doesn't let us know what is the native encoding of a dataset, so the value we returned before wasn't always accurate. It's more like Python's file object to return None.

sgillies:
The AppVeyor errors are (the pip exception) annoying and (build failure with 2.4) baffling.

snorfalorpagus:
I'm away from home this week, but will look at this when I get back.

fiona/gdal.pxi
```diff
@@ -522,4 +522,6 @@ cdef extern from "gdal_alg.h" nogil:
 
 cdef extern from "ogr_core.h" nogil:
 
+    const char * OLCStringsAsUTF8
```
snorfalorpagus: Is this used anywhere?

fiona/ogrext.pyx
```diff
@@ -480,38 +477,37 @@ cdef class Session:
         else:
             drivers = None
 
+        if 'encoding' in kwargs:
+            if not kwargs['encoding']:
+                kwargs.pop('encoding')
+            else:
+                kwargs['encoding'] = kwargs['encoding'].upper()
+
```
snorfalorpagus: I think this can be simplified to only one dictionary lookup:

```python
encoding = kwargs.pop("encoding", None)
if encoding:
    kwargs["encoding"] = encoding.upper()
```

fiona/ogrext.pyx
```diff
@@ -525,16 +521,53 @@ cdef class Session:
         self.cogr_ds = NULL
 
     def get_fileencoding(self):
+        """DEPRECATED"""
```
snorfalorpagus: Should this raise a deprecation warning?

fiona/ogrext.pyx
```diff
@@ -525,16 +521,53 @@ cdef class Session:
...
+        Parameters
+        ----------
+        None
+
+        Returns
+        -------
+        str
+
+        """
+        if "Shapefile" in self.get_driver():
```
snorfalorpagus: Could this ever be "ESRI Shapefile"?

snorfalorpagus: For example:

```python
>>> import fiona
>>> src = fiona.open("tests/data/gre.shp", driver="ESRI Shapefile")
>>> src.session.get_driver()
'ESRI Shapefile'
```

fiona/ogrext.pyx
```diff
@@ -991,19 +1022,16 @@ cdef class WritingSession(Session):
...
-            # 'iso-8859-1', then the system's default encoding as last resort.
+            # Determine which encoding to use. The encoding parameter given to
+            # the collection constructor takes highest precedence, then
+            # 'iso-8859-1' (for shapefiles), then the system's default encoding
+            # as last resort.
             sysencoding = locale.getpreferredencoding()
-            self._fileencoding = (userencoding or (
-                "Shapefile" in collection.driver and
-                'ISO-8859-1') or sysencoding).upper()
+            self._fileencoding = userencoding or ("Shapefile" in collection.driver and 'iso-8859-1') or sysencoding
```
snorfalorpagus: As previous re: "ESRI Shapefile"

fiona/gdal.pxi
sgillies: In a TestCapability call.

fiona/ogrext.pyx
sgillies: Yes, this was my attempt to shortcut and be future proof.

snorfalorpagus: Ah OK. Of course `"Shapefile" in` will match `"ESRI Shapefile"`.

PR #739, opened 2019-03-28
Draft RFC template

sgillies:
This would be the template for an RFC I will write concerning Fiona 2.0. I floated the idea in https://fiona.groups.io/g/dev/message/1 and will make it more formal after this PR is merged.

### DIFF
--- a/docs/rfc/0000-template.md
+++ b/docs/rfc/0000-template.md
@@ -1,0 +1,45 @@
+RFC 0: Fiona RFC template
+=========================
+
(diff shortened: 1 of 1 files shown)


sgillies:
Also note to self: remember to use [skip ci] or find a strategy for branching such that we don't waste build time for RFC PRs.

sgillies:
I'm moving this to https://github.com/Toblerity/fiona-rfc/pull/1 and will follow up on our dev discussion group thread.

PR #741, opened 2019-04-04
WIP: Use WKB for geometry instead of GeoJSON

snorfalorpagus:
This is a **work in progress**.

This branch exposes geometries for features as WKB data stored in a `bytes` object instead of GeoJSON-like structures.

I 100% think that GeoJSON should remain the default - for most situations the performance is acceptable, and it is a much easier route for people to get into working with spatial data. It's also easier to debug with.

Making this an option could improve the performance of workflows that use Shapely, e.g.

```
for feature in source:
    geom = shape(feature["geometry"])
    ...
```

vs

```
for feature in source:
    geom = wkb.loads(feature["geometry"])
    ...
```

GeoPandas would benefit from this also as it works natively with Shapely geometries.

Some quick benchmarking shows an improvement in performance for some workflows, but this will depend very much on the data being read and how it's being used.

TODO:

- How should this mode be enabled? An argument to `fiona.open`?
- Improved method names (not `build2`...)
- Documentation

### DIFF
--- a/fiona/_geometry.pyx
+++ b/fiona/_geometry.pyx
@@ -2,6 +2,8 @@
 
 from __future__ import absolute_import
 
+from libc.stdlib cimport malloc, free
+
 import logging
 
 from fiona.errors import UnsupportedGeometryTypeError
@@ -190,6 +192,16 @@ cdef class GeomBuilder:
         result = self.build(cogr_geometry)
         _deleteOgrGeom(cogr_geometry)
         return result
+
+    cdef bytes build2(self, void *geom):
+        cdef int size = OGR_G_WkbSize(geom)
+        cdef unsigned char* buf = <unsigned char*>malloc(size * sizeof(char))
+        if not buf:
+            raise MemoryError()
+        OGR_G_ExportToWkb(geom, 1, buf)
+        cdef bytes wkb = buf[:size]
+        free(buf)
+        return wkb
 
 
 cdef class OGRGeomBuilder:
--- a/fiona/_geometry.pxd
+++ b/fiona/_geometry.pxd
@@ -93,7 +93,7 @@ cdef extern from "ogr_api.h":
     void *  OGR_G_CreateGeometry (OGRwkbGeometryType wkbtypecode)
     void    OGR_G_DestroyGeometry (void *geometry)
     unsigned char * OGR_G_ExportToJson (void *geometry)
-    void    OGR_G_ExportToWkb (void *geometry, int endianness, char *buffer)
+    void    OGR_G_ExportToWkb (void *geometry, int endianness, unsigned char *buffer)
     int     OGR_G_GetCoordinateDimension (void *geometry)
     int     OGR_G_GetGeometryCount (void *geometry)
     unsigned char *  OGR_G_GetGeometryName (void *geometry)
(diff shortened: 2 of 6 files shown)


fiona/_geometry.pyx
```diff
@@ -191,6 +193,16 @@ cdef class GeomBuilder:
         _deleteOgrGeom(cogr_geometry)
         return result
 
+    cdef bytes build2(self, void *geom):
```
sgillies: or subclass GeomBuilder and override build?

sgillies:
@snorfalorpagus Fiona itself could benefit from WKB geoms. Imagine reading from a collection, transforming, and then writing a new collection (using OGR_G_ImportFromWkb).

If `feature["geometry"]` were an instance of a class that derived from, say, `collections.abc.Mapping` we could add `__wkb__` and `__geo_interface__` attributes to it, the former being WKB bytes and the latter being a GeoJSON-like dict. On reading a feature from a collection, maybe we set the `__wkb__` attribute only, and compute the `__geo_interface__` only right before it is needed, as when a user does `feature["geometry"]["type"]`.

sgillies:
@snorfalorpagus my suggestion above was half-baked. My preferred way to get better performance is to make a `cdef class _Feature` and `cdef class _Geometry`. Something like

```
cdef class _Feature:
    cdef OGRFeatureH _featrh

cdef class _Geometry:
    cdef OGRGeometryH _geomh
```

with public

```
class Feature(collections.abc.Mapping):
    _feature : _Feature
    geometry = property()  # read-write

class Geometry(collections.abc.Mapping):
    _geom: _Geometry
    wkb = property()  # read-only
    __geo_interface__ = property()  # read-only
   from_mapping = classmethod()  # a factory method
   from_wkb = classmethod()  # another factory method
```

WKB and GeoJSON representations of the feature's geometry could be evaluated lazily and in some cases, not at all. For example, the following would not need to evaluate either of those, it could exclusively use the low-level cdef OGRFeatureH _featrh.

```python
with fiona.open("example.shp") as source, fiona.open("result.shp", "w", ...) as destination:
    destination.writerecords(source)
```

Use with Shapely could be something like the following.

```python
def buffer_feature(feature):
    geom = feature.geometry  # implementation detail: OGR_F_GetGeometryRef(feature._feature._featrh) is called, geom keeps a reference to feature so that it is not garbage collected.
    wkb = geom.wkb  # I.D.: OGR_G_ExportToWkb(geom._geom._geomh) is called.
    buffered_geom = shapely.geometry.wkb.loads(wkb).buffer(100.0)
    feature.geometry = buffered_geom  # I.D.: OGR_F_SetGeometry(feature._feature._featrh, Geometry.from_mapping(buffered_geom)._geom._geomh) is called
    return feature

with fiona.open("example.shp") as source, fiona.open("result.shp", "w", ...) as destination:
    destination.writerecords(map(buffer_feature, source))
```

I don't think we need to implement separate geometry classes like Point. Also, making Geometry instances immutable will be key to keeping the implementation simple.

What do you think of this?

snorfalorpagus:
This work is superseded by proposals for Fiona 2.0: https://github.com/Toblerity/fiona-rfc/pull/2

PR #743, opened 2019-04-14
Implement transform function that accepts WKB

caspervdw:
This would increase performance in workflows that use WKB to store geometries. For instance in geopandas, one would need to transform every geometry into dictionaries and back to do a transform using fiona.

The new function transform.transform_wkb just acts on the WKB representation of the geometry.

This builds onto #741

### DIFF
--- a/fiona/_geometry.pxd
+++ b/fiona/_geometry.pxd
@@ -124,7 +124,10 @@ cdef class GeomBuilder:
     cpdef _buildGeometryCollection(self)
     cdef build(self, void *geom)
     cpdef build_wkb(self, object wkb)
-    cdef bytes build2(self, void *geom)
+
+
+cdef class WKBGeomBuilder:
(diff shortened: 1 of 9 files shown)


sgillies:
Hi @caspervdw there are some plans being made that may make your work here obsolete. I recommend pausing and catching up on https://fiona.groups.io/g/dev/message/10 and https://github.com/Toblerity/fiona-rfc/pull/2. The gist of that proposal is: Fiona 2.0 will have a new geometry class with a wkb attribute and we won't implement separate GeoJSON and WKB flavors of functions.

caspervdw:
@sgillies  No worries. I agree that having a double of every function is not a good design.
But faster interoperability with shapely and geopandas would be really welcome, I caught up with the RFC and it looks great.

caspervdw:
Closing. After some initial work is done on the new geometry class, I am happy to give this another go.

Issue #745, opened 2019-04-21
Fails to build with GDAL 2.5.0-beta1 (ogrext.so: undefined symbol: OSRFixup)


sebastic:
## Expected behavior and actual behavior.

The fiona Debian package fails to build with GDAL 2.5.0-beta1:
```
...
fiona/_crs.c: In function ‘__pyx_pf_5fiona_4_crs_crs_to_wkt’:
fiona/_crs.c:2183:10: warning: implicit declaration of function ‘OSRFixup’; did you mean ‘OSRCleanup’? [-Wimplicit-function-declaration]
   (void)(OSRFixup(__pyx_v_cogr_srs));
          ^~~~~~~~
          OSRCleanup
...
fiona/ogrext.c: In function ‘__pyx_pf_5fiona_6ogrext_14WritingSession_start’:
fiona/ogrext.c:17239:18: warning: implicit declaration of function ‘OSRFixup’; did you mean ‘OSRCleanup’? [-Wimplicit-function-declaration]
           (void)(OSRFixup(__pyx_v_cogr_srs));
                  ^~~~~~~~
                  OSRCleanup
...
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/sphinx/config.py", line 368, in eval_config_file
    execfile_(filename, namespace)
  File "/usr/lib/python2.7/dist-packages/sphinx/util/pycompat.py", line 150, in execfile_
    exec_(code, _globals)
  File "/usr/lib/python2.7/dist-packages/six.py", line 709, in exec_
    exec("""exec _code_ in _globs_, _locs_""")
  File "<string>", line 1, in <module>
  File "/build/fiona-1.8.4/docs/conf.py", line 14, in <module>
    import fiona
  File "/build/fiona-1.8.4/.pybuild/cpython2_2.7_fiona/build/fiona/__init__.py", line 83, in <module>
    from fiona.collection import BytesCollection, Collection
  File "/build/fiona-1.8.4/.pybuild/cpython2_2.7_fiona/build/fiona/collection.py", line 9, in <module>
    from fiona.ogrext import Iterator, ItemsIterator, KeysIterator
ImportError: /build/fiona-1.8.4/.pybuild/cpython2_2.7_fiona/build/fiona/ogrext.so: undefined symbol: OSRFixup
```

[...]

PR #746, opened 2019-04-21
Add support for travis to build [beta|rc] versions of gdal and fiona rfc1

rbuffat:
Gdal 2.5.0beta1 is released and Fiona fails to build against it: https://github.com/Toblerity/Fiona/issues/745

This PR includes the work I did in https://github.com/Toblerity/Fiona/pull/696 to support the installation of beta and rc versions of GDAL on travis.

However, GDAL 2.5.0beta1 fails to build because it has now a dependency on proj6, which I did not yet find an ubuntu package for.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,15 +11,19 @@ env:
     - PIP_FIND_LINKS=file://$HOME/.cache/pip/wheels
     - GDALINST=$HOME/gdalinstall
     - GDALBUILD=$HOME/gdalbuild
+    - PROJBUILD=$HOME/projbuild
+    - PROJVERSION="6.0.0"
+    
+    
   matrix:
-    - GDALVERSION="1.11.5"
-    - GDALVERSION="2.0.3"
-    - GDALVERSION="2.1.4"
-    - GDALVERSION="2.2.4"
+#    - GDALVERSION="1.11.5"
+#    - GDALVERSION="2.0.3"
+#    - GDALVERSION="2.1.4"
+#    - GDALVERSION="2.2.4"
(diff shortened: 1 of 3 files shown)


snorfalorpagus:
I guess we could install PROJ 6 from source until a package is available?

rbuffat:
> I guess we could install PROJ 6 from source until a package is available?

I'm working on it, but this is not a good solution in my opinion.

rbuffat:
I now extended the install script to include building proj.4 version 6.0.0 from source. I think this needs to be only a temporary solution, as this increases the build time to over 20 minutes per matrix entry (without caching). The installation of proj gets cached along with gdal, thus consecutive builds should need only a bit more than before. 

Furthermore, I reduced the build matrix to correspond to [fiona rfc1](https://github.com/Toblerity/fiona-rfc/blob/master/rfc/0001-fiona-2-0-changes.md). This will help regarding the build times in the short term.
I also changed travis to xenial (ubuntu 16.04) and python 3.7 instead of 3.6.

sgillies:
@rbuffat thanks! As to the failure: I'm going to solve the missing OSRFixup problem in a temporary fashion so we can see how Fiona fares with GDAL 2.5, while working on restoring OSRFixup to GDAL.

sgillies:
It turns out that restoring OSRFixup barely is only the start of solving the 2.5 compatibility issues. I think we need to consider making 2.5 the baseline version of GDAL for Fiona 2.0 and then add shims for older GDAL versions (2.2-2.4). I'm going to merge this and we'll get started on the shims.

Issue #212, opened 2015-03-01
Incompatibility with Cython 0.22


youngpm:
Fiona is currently incompatible with the newest release of Cython (v0.22); this is only an issue for development using v0.22 and it compiles fine with v0.21.2.  Cython spits out a bunch of errors of the form
```
Cythonizing fiona/_geometry.pyx

Error compiling Cython file:
------------------------------------------------------------
...


cdef class OGRGeomBuilder:
    """Builds OGR geometries from Fiona geometries.
    """
    cdef void * _createOgrGeometry(self, int geom_type) except NULL:
        ^
------------------------------------------------------------

fiona/_geometry.pyx:138:9: Signature not compatible with previous declaration

Error compiling Cython file:
------------------------------------------------------------
...
    cdef build(self, void *geom)
    cpdef build_wkb(self, object wkb)


cdef class OGRGeomBuilder:
    cdef void * _createOgrGeometry(self, int geom_type)
             ^
------------------------------------------------------------
```

The problem is with non-Cython `cdef` declarations that end with `except NULL`, like [here](https://github.com/Toblerity/Fiona/blob/master/fiona/_geometry.pyx#L40).  Apparently the new version of Cython is stricter about such declarations, see [here](http://docs.cython.org/src/reference/language_basics.html#checking-return-values-for-non-cython-functions).

sgillies:
@youngpm I ran into this with Rasterio, too. My temporary solution there was to pin cython==0.21.2 in requirements-dev.txt. Let's do that for Fiona and I'll take on the migration to Cython 0.22 when we get closer to a 1.6 release.

PR #764, opened 2019-06-16
Test unpinned python dependencies using travis cron

rbuffat:
Since some time, requirements*.txt include pinned versions of python dependencies.  Thus, we do not check compatibility against updated dependencies and thus can miss potential issues.  Unpinning the versions is also not great, as dependencies update asynchronously to Fiona commits triggering the CI.

This PR updates the travis configuration so that when a build is triggered using a travis cron job, the pinned versions in the requirements*.txt are removed, thus installing the latest dependencies.  For cron jobs, only known working builds are included (currently excluding gdal 3.0 and trunk). 
In order to use these changes, a travis cron job needs to be configured: More options -> settings -> cron jobs: branch: master, interval: daily, options: always run

Potentials caveats are:
- a cron job can only be executed for roughly 1000 times for the same commit.
- a broken dependency can invalidate the proj/gdal cache and thus increase the build time for commits

Furthermore, this PR contains some cleanup in the travis configuration and removes enum34 and ordereddict as dependencies (drop of support for python < 3.5). Probably, six could also be removed, but this needs some refactoring.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ cache:
     - $GDALINST
     - ~/.cache/pip
 
+git:
+  depth: 3    
+
 env:
   global:
     - PIP_WHEEL_DIR=$HOME/.cache/pip/wheels
@@ -23,10 +26,18 @@ env:
     - GDALVERSION="2.2.4"
     - GDALVERSION="2.3.3"
     - GDALVERSION="2.4.1"
-    - GDALVERSION="3.0.0"
-    - GDALVERSION="trunk"
 
 matrix:
+
+  include:
+   - python: "3.7"
+     env: GDALVERSION="3.0.0"
+     if: type != cron
+
+   - python: "3.7"
+     env: GDALVERSION="trunk"
+     if: type != cron
+
   allow_failures:
     - env: GDALVERSION="trunk"
     - env: GDALVERSION="3.0.0"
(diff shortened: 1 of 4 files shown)


rbuffat:
Also, I noticed there are some inconsistencies between dependencies listed in README.rst and setup.py/requirements.txt. e.g. attrs (is this even used) and click-plugins.

```
grep -nr "attrs" .
./CHANGES.txt:144:- Dependency on attrs is made explicit.
./requirements.txt:2:attrs==18.2.0
./setup.py:269:    'attrs>=17',
```

sgillies:
@rbuffat thanks for looking into this. I'd like to have a little more discussion about this one. Until now, the Fiona builds have been focused on testing Fiona against different versions of Python and GDAL and we've pinned dependencies so that Python requirements don't change and continually break the build, obscuring issues with GDAL. I'm reluctant to change our strategy, because dependency incompatibilities haven't been common, but I suppose we should consider a change of strategy.

rbuffat:
@sgillies  I don't mind if you don't merge this RR. The idea was a bit spontaneous after I stumbled on [greenkeeper.io](https://greenkeeper.io) which does dependency checking for node js integrated into github and I wanted to try if something similar can be done with python. I think the situation between nodejs, where I assume there are a lot of fast changing libraries, and fiona is quite different.  As you mentioned before, all the dependencies fiona uses are quite major and stable. I suspect the dependency with the highest potential for issues is cython (e.g https://github.com/Toblerity/Fiona/issues/212).

I completely agree that the best strategy for regular commits / PR is to use a known working set of pinned dependencies. However, this strategy leaves two questions open:
- Do the minimal specified requirements in setup.py still work? (not yet addressed)
- Does fiona works with the latest version of its dependencies? (addressed in this PR)

If we should test these cases depends on the likelihood of something to break versus the work needed to create, maintain and run a system that checks them.

After having a daily cron job enabled on travis for this branch for a few days I noticed that for two consecutive days there were some network errors (HTTP code 503) for random builds.  Thus, if we want to go forward with testing on a daily basis probably the unit tests should be adopted that they just emit a warning and do not fail if the network is not available.

```
E   fiona._err.CPLE_HttpResponseError: HTTP response code: 503
E   fiona.errors.DriverError: '/vsizip/vsicurl/http://raw.githubusercontent.com/OSGeo/gdal/master/autotest/ogr/data/poly.zip' does not exist in the file system, and is not recognized as a supported dataset name.
```

sgillies:
@rbuffat thanks. I'm going to close this one if you still don't mind.

PR #770, opened 2019-07-16
Establish support for Python 3.6, 3.7 and 3.8

mwtoews:
This PR establishes the current supported Python versions in both `setup.py` and Travis CI tests.

A few other notes:
* Python version check is also done with `setup.py`
* Update a few aspects with `.travis.yml`, such as removing old sudo part (now ignored), and re-arranged matrix layout to do a few more tests with other Python versions and a single version of GDAL
* `argparse` and `OrderedDict` are now in standard modules
* `enum34` was only needed for older not-supported versions
* <s>Minor fix to `tests/test_listing.py` for Python 2.7 testing</s>


.travis.yml
```diff
@@ -17,19 +14,50 @@ env:
...
+      env:
+      - GDALVERSION="2.2.4"
+
+    - python: "3.7"
+      env:
+      - GDALVERSION="2.2.4"
+
+    - python: "3.7"
+      env:
+      - GDALVERSION="2.3.3"
```
sgillies: @mwtoews I like this approach of making the matrix more sparse. I think we can take it a little further. Python 3.7 hasn't made it out into distros very far yet, and so I think that any users are likely to be using nearly the cutting edge of GDAL. What would you think about dropping GDAL versions < 2.4 for Python 3.7?

README.rst
```diff
@@ -19,7 +19,7 @@ and protocols such as files, dictionaries, mappings, and iterators instead of
 classes specific to OGR. Fiona can read and write real-world data using
 multi-layered GIS formats and zipped virtual file systems and integrates
 readily with other Python GIS packages such as pyproj_, Rtree_, and Shapely_.
-Fiona is supported only on CPython versions 2.7 and 3.4+.
+Fiona is supported only on CPython versions 2.7 and 3.5+.
```
sgillies: 👍

rbuffat:
Would it make sense to also include python 3.8-dev (should be released later this year if I remember correctly) and gdal 2.4.x? 

I think for python 3.8 we also need to bump the cython dependency to a newer version.

mwtoews:
I was considering 3.8-dev (yes 3.8 is coming up soon), except that this would be best paired with GDAL 3.0.1 or trunk. Except these are currently failing. Perhaps this could be enabled when GDAL 3.0 support is established? Otherwise, it's not a big deal to add more to the test matrix (they're cheap!)

rbuffat:
@mwtoews yes adding them is cheap, but for each entry that is not cached gdal is compiled from source which requires roughly 10 minutes on travis.

.travis.yml
```diff
@@ -17,19 +14,46 @@ env:
...
 
 matrix:
+  include:
+    - python: "2.7"
+      env:
+      - GDALVERSION="2.2.4"
+
+    - python: "3.5"
+      env:
+      - GDALVERSION="2.2.4"
```
sgillies: @mwtoews we can eliminate these combinations. 1.9.0 will announce deprecation of Python 2.7 and 3.5 support.

rbuffat:
@mwtoews in case fiona fails to build with python 3.8 you can bunp the cython dependency to a version >= 0.29

mwtoews:
@sgillies see a recent commit on this branch, which drops 2.7 and 3.5, adds 3.8, and also upgrades the GDAL versions to their latest versions.

@rbuffat thanks for the tip, as I see the error "ValueError: 'fiona/_shim.pyx' doesn't match any files"; trying cython==0.29.12 right now to see if that helps

I'll assume that the "AttributeError: 'UnparsedPath' object has no attribute 'scheme'" error is unrelated, as I see this in master now too.

rbuffat:
@mwtoews the shim.pyx is due to missing gdal 3.0 supporr. Support was added recently in maint-1.8 and probabely maint-1.9 branch but not yet in master. Sorry, I'm currently only on my mobile phone online, otherwise I might have spotted this earlier.

mwtoews:
@rbuffat thanks again, I now see #804 for GDAL 3 support

@sgillies I'll await any suggestions to get this PR in shape

mwtoews:
I've merged any conflicts with master, so ready for more review

setup.py
```diff
@@ -10,6 +10,10 @@
 from setuptools.extension import Extension
 
 
+# Ensure minimum version of Python is running
+if sys.version_info[0:2] < (3, 6):
+    raise RuntimeError('Fiona requires Python>=3.6')
```
sgillies: 👍

sgillies:
There's something odd about this PR @mwtoews. A flaky test that was fixed in master remains in your branch: https://ci.appveyor.com/project/sgillies/fiona/builds/28983820/job/j6s8py7codbewui7#L1412.

Sadly we can't ever pass "zip://file.zip" URLs to pathlib.Path, it doesn't parse them correctly.

sgillies:
Nevermind my previous comment, seems like AppVeyor and GitHub are in sync.

sgillies:
Excellent, thanks @mwtoews !

PR #775, opened 2019-07-20
Register pytest markers and upgrade pytest

snorfalorpagus:
Registering the markers used by pytest helps to prevent typos, as unknown markers in newer versions raise a warning.

https://docs.pytest.org/en/latest/mark.html#registering-marks

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,8 +1,8 @@
 -r requirements.txt
 coverage==4.5.1
 cython==0.28.3
-pytest==3.4.2
-pytest-cov==2.5.1
+pytest==5.0.1
+pytest-cov==2.7.1
 setuptools==39.0.1
 boto3==1.9.19
 wheel==0.31.1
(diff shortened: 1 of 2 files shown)


sgillies:
Fantastic. Thanks, @snorfalorpagus !

snorfalorpagus:
This MR broke the appveyor build. Something to do with an older version of argparse but I haven't managed to reproduce it locally yet.

rbuffat:
@snorfalorpagus  There is a similar bug report in the pytest issue tracker: https://github.com/pytest-dev/pytest/issues/5594

Maybe something is odd with our Python setup on appveyor.

rbuffat:
I think the reason is that we install the argparse library with requirements.txt. Dropping this requirement fixes this issue. I think this should not be a problem as we won't support Python < 3.5 with Fiona 2.0 and argparse is in the standard library after Python 3.2.  To be safe we could add 
``` argparse==1.4.0 ; python_version <= '3.4' ```
(From https://pypi.org/project/argparse/ I assume the separate argparse library provides bugfixes up to 3.4)

It's strange that the error is not present on travis.

Issue #774, opened 2019-07-18
M Values in Shapefile aren't shown


alexmahrou:
## Expected behavior and actual behavior.

I have an M-Enabled Linestring Shapefile, and I've verified the x,y,m coordinates in a GIS. However upon bringing this into fiona and querying it, I only see the x,y. I plan to write this to postgis, but I'm concerned that the m values will not be preserved. Is there a way to ensure recognition of M values within fiona?

## Steps to reproduce the problem.

fiona.open(shp)
pprint.pprint(next(shp)['geometry']) yields only x and y.
![image](https://user-images.githubusercontent.com/10987392/61474643-4ca1fc80-a946-11e9-9ea2-a5591e8bd9d9.png)

## Operating system

Windows 10 conda virtual env.

## Fiona and GDAL version and provenance

gdal_version = 2.4.1
fiona = 1.8.6

sgillies:
@alexmahrou Fiona 1.x doesn't support M geometry types. It's on the radar for 2.0. Meanwhile, I recommend shp2pgsql or some other classic tool.

alexmahrou:
Thanks @sgillies  I appreciate your response! I'll give that a whirl for sure.

PR #778, opened 2019-07-29
Polyline m support

marouaneloutfi:
Following the issue #[774](https://github.com/Toblerity/Fiona/issues/774), this PR establishes support for Measurement values in shapefiles for both single point and LineString geometries.
it would also be the first step in many towards extending M-values support  to modules using Fiona for read/write purposes such as Geopandas.


sgillies:
@marouaneloutfi are you a subscriber at https://fiona.groups.io/g/dev? I'd like to have a discussion there about this PR as it will be a fairly big change.

marouaneloutfi:
@sgillies I've just subscribed to the dev group, I am looking forward to hearing your thoughts on this.

sgillies:
@marouaneloutfi I'm back from vacation and would like to get this moving again. I may not have been clear before, but I'm looking to you to take the lead on explaining the need for M support in Fiona, the design, interoperability considerations and things like that. I bet that with a good introductory email, other developers will indicate interest and/or have suggestions.

sgillies:
This is too old now. We should start fresh after 1.9.0 comes out.

Issue #747, opened 2019-04-22
Thoughts about improving travis


rbuffat:
Fiona should work for a range of different gdal versions, which we want to ensure by testing fiona against them on CI such as travis and appveyor.  While on appveyor we can use prebuilt binaries of gdal, we compile gdal from source on travis. The reason for that is that the standard ubuntu repository, respectively PPA such as UbuntuGIS only contain specific versions of gdal, which we can not control. Compiling GDAL, again and again, is a waste of time and resources. Although the impact of this is mitigated by caching the compiled binaries. However, for certain scenarios, such as new branches, new PR or previously failed jobs, this is not possible.

A solution would be to create prebuilt binaries of gdal and related libraries, such as proj.4 on our own. The question is, what would be the best way to do it? I can think of a few alternatives, but I'm not knowledgable in any of them:
- Creating a specific PPA with packages prebuilt binaries for the travis image we currently use. I'm not sure if we can have multiple versions of the same package in the same PPA and install a specific version. Furthermore, for me at least, building packages for Launchpad does not look straight forward. It looks as @sebastic has some experience with debian packages, maybe he can provide some insights.  
- Building binaries with travis in a separate github project and deploy the gdal binaries as release artifacts. This would probably be kind of a misuse of releases.
- Instead of deploying to github deploying to S3.

sebastic:
If you don't need packages for other architectures than amd64 (and i386), I would setup a custom package repository with reprepro for the various branches on GDAL you want to support, and add this repo to Travis.

If you need other architectures like arm*, Launchpad is a better option.

rbuffat:
@sebastic Thanks! I will look into if I can manage to build something with reprepro, travis and github pages.

PR #779, opened 2019-07-29
[WIP] Use precompiled deb packages for proj and gdal on travis

rbuffat:
This PR is currently a work in progress and aims at addressing the following issue https://github.com/Toblerity/Fiona/issues/747.

Travis needs roughly 15 minutes to compile on travis.  For most of the travis test matrix, cached versions of gdal are used. However, for all failing builds, respectively, for gdal master, this is not possible.  

I created a github project (https://github.com/rbuffat/gdal_builder) that builds gdal / proj deb packages on travis and uploads them to github pages. The deb packages have the same file structure as we currently use on travis. This allows using the existing gdal install scripts with small modifications. In case no deb package is found, gdal is compiled from source as before. The project can also build packages for gdal master, which are currently updated with a travis cron job on weekly basis (daily would also be possible).  

This reduces the travis run time from roughly 20-25 minutes to under 10 minutes. 

The current state is a working prototype. It still needs thorough testing and some modification. Also, it is probabely a good idea to have a discussion about to pros and cons of this approach. One downside is for sure the added complexity of the travis setup.

(I added gdal 2.5.0rc1 to the matrix only to test the script for beta/rc versions.)


rbuffat:
During the last months, I worked on and off on this PR.  To be honest it is more fragile than I anticipated. Currently, the travis tests are green on [my repository](https://travis-ci.org/rbuffat/Fiona/builds/599326234), but not here. The reason why this is like this I need yet to find out. 

In the past one of the headaches was, when I tried this method with rasterio, two proj related tests [failed](https://travis-ci.org/rbuffat/rasterio/jobs/576990421). I found out that the gdal binaries between the precompiled and the locally on travis compiled were not identical. Switching the travis image from "language: minimal" to "language: python" let the tests pass. I'm not yet sure what caused this issue, as both the python and minimal image should be quite similar.

In theory, the precompiled binaries should be more or less the same as the travis caching. The builder project currently builds proj, respectively gdal, and uploads a tar.gz archive to github pages.

I'm still of the opinion that having precompiled binaries would be quite an advantage, as the build time per row of the build matrix reduces to 3-4 minutes, including the git master version. Thus, even if everything is cached, travis would finish still roughly ~30min faster, as the last time I checked it only finishes after the master the git master job finishes.

rbuffat:
I'm closing this for now. I never got it to work properly.

PR #787, opened 2019-09-19
Add Feature, Geometry and ObjectEncoder classes to model

### LINKED ISSUE #760: Add feature, geometry, properties classes to model
These classes will extend model.Object.

sgillies:
Resolves #760

### DIFF
--- a/fiona/_geometry.pyx
+++ b/fiona/_geometry.pyx
@@ -5,7 +5,7 @@ from __future__ import absolute_import
 import logging
 
 from fiona.errors import UnsupportedGeometryTypeError
-
+from fiona.model import Geometry
 
 class NullHandler(logging.Handler):
     def emit(self, record):
@@ -181,7 +181,8 @@ cdef class GeomBuilder:
         self.geomtypename = GEOMETRY_TYPES[self.code]
         self.ndims = OGR_G_GetCoordinateDimension(geom)
         self.geom = geom
-        return getattr(self, '_build' + self.geomtypename)()
+        built = getattr(self, '_build' + self.geomtypename)()
+        return Geometry(**built)
 
     cpdef build_wkb(self, object wkb):
         # The only other method anyone needs to call
--- a/fiona/_transform.pyx
+++ b/fiona/_transform.pyx
@@ -10,6 +10,7 @@ from fiona cimport _cpl, _crs, _csl, _geometry
 from fiona._crs cimport OGRSpatialReferenceH
 
 from fiona.compat import UserDict
+from fiona.model import Geometry
 
 
 cdef extern from "ogr_geometry.h" nogil:
@@ -126,6 +127,9 @@ def _transform_geom(
     cdef void *src_ogr_geom = NULL
     cdef void *dst_ogr_geom = NULL
     cdef int i
+
+    if not isinstance(geom, Geometry):
+        geom = Geometry.from_dict(**geom)
 
     if src_crs and dst_crs:
         src = _crs_from_crs(src_crs)
--- a/fiona/fio/cat.py
+++ b/fiona/fio/cat.py
@@ -1,4 +1,4 @@
-"""$ fio cat"""
+"""fio-cat"""
 
 
 import json
@@ -10,6 +10,7 @@ import cligj
 
 import fiona
 from fiona.transform import transform_geom
+from fiona.model import ObjectEncoder
 from fiona.fio import options, with_context_env
 
 
@@ -82,7 +83,7 @@ def cat(ctx, files, precision, indent, compact, ignore_errors, dst_crs,
                             feat['bbox'] = fiona.bounds(g)
                         if use_rs:
                             click.echo(u'\u001e', nl=False)
-                        click.echo(json.dumps(feat, **dump_kwds))
+                        click.echo(json.dumps(feat, cls=ObjectEncoder, **dump_kwds))
 
     except Exception:
         logger.exception("Exception caught during processing")
--- a/fiona/model.py
+++ b/fiona/model.py
@@ -2,6 +2,7 @@
 
 from collections.abc import MutableMapping
 import itertools
+from json import JSONEncoder
 from warnings import warn
 
 from fiona.errors import FionaDeprecationWarning
@@ -89,6 +90,15 @@ class Geometry(Object):
         self._delegate = _Geometry(coordinates=coordinates, type=type)
         super(Geometry, self).__init__(**data)
 
+    @classmethod
+    def from_dict(cls, **data):
+        data = data.copy()
+        return Geometry(
+            coordinates=data.pop("coordinates", None),
+            type=data.pop("type", None),
+            **data
+        )
+
     @property
     def coordinates(self):
         """The geometry's coordinates
@@ -145,7 +155,7 @@ class Feature(Object):
                 type=geom_data.pop("type", None),
                 **geom_data
             )
-            if geom_data
+            if geom_data is not None
             else None
         )
         props_data = data.pop("properties", None)
@@ -196,3 +206,19 @@ class Feature(Object):
 
         """
         return "Feature"
+
+
+class ObjectEncoder(JSONEncoder):
+    """Encodes Geometry and Feature"""
+
+    def default(self, o):
+        if isinstance(o, Geometry):
+            return dict(**o)
+        elif isinstance(o, Feature):
+            o_dict = dict(**o)
+            o_dict["type"] = "Feature"
+            if o.geometry is not None:
+                o_dict["geometry"] = ObjectEncoder().default(o.geometry)
+            return o_dict
+
+        return JSONEncoder().default(o)
--- a/fiona/model.py
+++ b/fiona/model.py
@@ -1,6 +1,7 @@
 """Fiona data model"""
 
 from collections.abc import MutableMapping
+import itertools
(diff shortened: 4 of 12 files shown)


fiona/_geometry.pyx
```diff
@@ -5,7 +5,7 @@ from __future__ import absolute_import
 import logging
 
 from fiona.errors import UnsupportedGeometryTypeError
-
+from fiona.model import GEOMETRY_TYPES, Geometry
```
sgillies: To avoid circular import.

fiona/_transform.pyx
```diff
@@ -127,6 +128,9 @@ def _transform_geom(
     cdef void *dst_ogr_geom = NULL
     cdef int i
 
+    if not isinstance(geom, Geometry):
+        geom = Geometry.from_dict(**geom)
+
```
sgillies: This is perhaps better done in transform.py. We could strictly require Geometry in the extension module since it isn't in the public API.

fiona/fio/bounds.py
```diff
@@ -66,7 +67,7 @@ def bounds(ctx, precision, explode, with_id, with_obj, use_rs):
                         rec = (w, s, e, n)
                     if use_rs:
                         click.echo(u'\u001e', nl=False)
-                    click.echo(json.dumps(rec))
+                    click.echo(json.dumps(rec, cls=ObjectEncoder))
```
sgillies: ObjectEncoder can serialize Feature, Geometry, and Properties.

fiona/fio/cat.py
```diff
@@ -63,27 +61,22 @@ def cat(ctx, files, precision, indent, compact, ignore_errors, dst_crs,
...
-                                precision=precision)
-                            feat['geometry'] = g
-                            feat['bbox'] = fiona.bounds(g)
-                        if use_rs:
-                            click.echo(u'\u001e', nl=False)
-                        click.echo(json.dumps(feat, **dump_kwds))
-
-    except Exception:
-        logger.exception("Exception caught during processing")
-        raise click.Abort()
```
sgillies: The big try/except block isn't needed with click now. I'm not sure it ever was.

fiona/fio/helpers.py
```diff
@@ -8,11 +8,13 @@
...
 from munch import munchify
 
+from fiona.model import to_dict
+
 
 warnings.simplefilter('default')
 
 
-def obj_gen(lines):
+def obj_gen(lines, object_hook=None):
```
sgillies: Use decode_object() as object_hook to get Features or Geometry.

fiona/model.py
```diff
@@ -17,31 +53,268 @@ class Object(MutableMapping):
     about future deprecation of features.
     """
 
-    def __init__(self, **data):
-        self._data = {}
-        self._data.update(data)
+    _delegated_properties = []
+
+    def __init__(self, **kwds):
+        self._data = OrderedDict(**kwds)
```
sgillies: OrderedDict to support ordered Properties.

fiona/ogrext.pyx
```diff
@@ -291,13 +289,7 @@ cdef class FeatureBuilder:
...
                     geom = GeomBuilder().build(cogr_geometry)
 
-                fiona_feature["geometry"] = geom
-
-            else:
-
-                fiona_feature["geometry"] = None
-
-        return fiona_feature
+        return Feature(id=str(fid), properties=Properties(**props), geometry=geom)
```
sgillies: Here we go! Features, Properties, and Geometry.

sgillies:
@jorisvandenbossche you might want to have a look at this too. Fiona 1.9 will still accept data as dicts but will start returning dict-like Feature objects that are not instances of dict.

I'm eager to get this merged in so I can move on to supporting GDAL 3.0.

jorisvandenbossche:
@sgillies this looks good to me. I quickly tried out this branch with geopandas, and as expected all tests pass without problems.

fiona/model.py
```diff
@@ -17,31 +53,268 @@ class Object(MutableMapping):
...
+    """
+
+    _delegated_properties = ["coordinates", "type"]
+
+    def __init__(self, coordinates=None, type=None, **data):
+        self._delegate = _Geometry(coordinates=coordinates, type=type)
+        super(Geometry, self).__init__(**data)
+
+    @classmethod
+    def from_dict(cls, **kwds):
```
jorisvandenbossche: This is actually not "from a dict", in the sense that with the current implementation you cannot pass a dictionary:

```
geom = {'type': 'Point', 'coordinates': [...]}
Geometry.from_dict(geom)
```

will not work. `Geometry.from_dict(**geom)` of course does, but the name might imply otherwise. 

(just something I noticed while going through the code, not that you need to change anything ;) I also don't know how public those classes will be?)

sgillies: It would not be hard to take dict or **dict. Thanks for the feedback and testing, @jorisvandenbossche !

fiona/model.py
```diff
@@ -17,31 +53,269 @@ class Object(MutableMapping):
...
+        self.type = type
+
+
+class Geometry(Object):
+    """A GeoJSON-like geometry
+
+    Notes
+    -----
+    Delegates coordinates and type properties to an instance of
+    _Geometry, which will become an extension class in Fiona 2.0.
```
snorfalorpagus: Could you explain the purpose of these delegation classes? Why not expose the extension class as `Geometry` in the future?

tests/test_multiconxn.py
```diff
@@ -57,16 +55,14 @@ def test_meta(self):
     def test_read(self):
         c2 = fiona.open(self.shapefile_path, "r")
         f2 = next(iter(c2))
-        del f2['id']
-        assert self.f == f2
+        assert self.f.properties == f2.properties
```
snorfalorpagus: Is this an API breaking change? Does the comparison of features no longer work?

tests/test_collection.py
```diff
@@ -313,10 +314,10 @@ def test_ignore_geometry(self):
...
 
             feature = next(iter(collection))
-            assert(feature["properties"]["AREA"] is not None)
-            assert(feature["properties"]["STATE"] is not None)
-            assert(feature["properties"]["NAME"] is not None)
-            assert("geometry" not in feature.keys())
+            assert(feature.properties["AREA"] is not None)
+            assert(feature.properties["STATE"] is not None)
+            assert(feature.properties["NAME"] is not None)
+            assert(feature.geometry is None)
```
snorfalorpagus: Why does this need to change?

tests/test_model.py
```diff
@@ -27,9 +35,260 @@ def test_setitem_warning():
...
+def test_feature_no_geometry():
+    """Feature has no attribute"""
+    feat = Feature()
+    assert feat.geometry is None
+
+
+def test_feature_geometry():
+    """Feature has a geometry attribute"""
+    feat = Feature(geometry=Geometry(type="Point"))
+    assert feat.geometry is not None
```
snorfalorpagus:
```python
geom = Geometry(type="Point")
feat = Feature(geometry=geom)
assert feat.geometry is geom
```

tests/test_model.py
```diff
@@ -27,9 +35,260 @@ def test_setitem_warning():
...
+
+def test_feature_no_properties():
+    """Feature has no properties"""
+    feat = Feature()
+    assert feat.properties is None
+
+
+def test_feature_properties():
+    """Feature has properties"""
+    feat = Feature(properties=Object(foo=1))
```
snorfalorpagus: Should this be `properties=Properties(foo=1)`? Is it valid to pass `Object` like this?

tests/test_model.py
```diff
@@ -27,9 +35,260 @@ def test_setitem_warning():
...
+def test_feature_complete_bis():
+    """Feature can be created from GeoJSON"""
+    data = {
+        "id": "foo",
+        "type": "Feature",
+        "geometry": {"type": "Point", "coordinates": (0, 0)},
+        "properties": {"a": 0, "b": "bar"},
+        "extras": {"this": 1},
+    }
+    feat = Feature.from_dict(data)
```
snorfalorpagus: It took me a moment to spot the difference between these tests. Can you expand on the docstring?

fiona/model.py
sgillies: @snorfalorpagus it's a compromise. On the one hand we need an extension class to get the performance boost we're looking for. On the other hand, using only an extension class limits our implementation options. A Cython cdef class can't derive from a Python class, so we can't inherit MutableMapping and gain its methods.

The delegation is working out well in Rasterio's CRS case, and I think we could benefit from it here.

tests/test_collection.py
sgillies: I'd like to start paving the way for preferring attribute access to item access in the future.

snorfalorpagus: That's OK (and sounds good to me). I'm just checking this isn't a breaking change now.

tests/test_multiconxn.py
sgillies: Feature comparison restored.

sgillies:
@snorfalorpagus to your question, very belatedly :laughing: , the Geometry/_Geometry design was is anticipation of rewriting _Geometry as a Cython extension class, but now I think it was misguided. I'll probably undo it in a future PR.

PR #800, opened 2019-10-09
return None and raise warning when transform_geom dst_ogr_geom is NULL

snowman2:
Related to: https://github.com/mapbox/rasterio/issues/1170

PR #802, opened 2019-10-11
return None and raise warning when transform_geom dst_ogr_geom is NULL

snowman2:
Forward from #800

### DIFF
--- a/fiona/_transform.pyx
+++ b/fiona/_transform.pyx
@@ -5,6 +5,7 @@
 from __future__ import absolute_import
 
 import logging
+import warnings
 
 from fiona cimport _cpl, _crs, _csl, _geometry
 from fiona._crs cimport OGRSpatialReferenceH
@@ -126,12 +127,10 @@ def _transform_geom(
     cdef void *src_ogr_geom = NULL
     cdef void *dst_ogr_geom = NULL
     cdef int i
-
     if src_crs and dst_crs:
         src = _crs_from_crs(src_crs)
         dst = _crs_from_crs(dst_crs)
         transform = _crs.OCTNewCoordinateTransformation(src, dst)
-
         # Transform options.
         options = _csl.CSLSetNameValue(
                     options, "DATELINEOFFSET", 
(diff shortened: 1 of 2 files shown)


fiona/_transform.pyx
```diff
@@ -145,10 +144,17 @@ def _transform_geom(
...
                         options)
-        g = _geometry.GeomBuilder().build(dst_ogr_geom)
-
-        _geometry.OGR_G_DestroyGeometry(dst_ogr_geom)
-        _geometry.OGR_G_DestroyGeometry(src_ogr_geom)
+        if dst_ogr_geom == NULL:
+            g = None
+            warnings.warn(
+                "Destination geometry is None. If desired, consider wrapping "
+                " with fiona.Env(OGR_ENABLE_PARTIAL_REPROJECTION='"'YES'"')."
```
sgillies:
```suggestion
                " the transform_geom call in a Fiona.Env context like `with fiona.Env(OGR_ENABLE_PARTIAL_REPROJECTION=True): transform_geom(...)`."
```

fiona/_transform.pyx
```diff
@@ -145,10 +144,17 @@ def _transform_geom(
...
                         <OGRCoordinateTransformation *>transform,
                         options)
-        g = _geometry.GeomBuilder().build(dst_ogr_geom)
-
-        _geometry.OGR_G_DestroyGeometry(dst_ogr_geom)
-        _geometry.OGR_G_DestroyGeometry(src_ogr_geom)
+        if dst_ogr_geom == NULL:
+            g = None
+            warnings.warn(
+                "Destination geometry is None. If desired, consider wrapping "
```
sgillies:
```suggestion
                "Transformation of the geometry has failed because partial reprojection has not been enabled. To enable partial reprojection wrap "
```

snowman2:
@sgillies, I updated the error message with a mix of your suggestion and the GDAL error message:
```
ERROR 1: Full reprojection failed, but partial is possible if you define OGR_ENABLE_PARTIAL_REPROJECTION configuration option to TRUE
```

The current iteration looks like:
```
 UserWarning: Full reprojection failed, but partial is possible. To enable partial reprojection wrap the transform_geom call like so:
with fiona.Env(OGR_ENABLE_PARTIAL_REPROJECTION=True):
    transform_geom(...)
```

PR #806, opened 2019-10-19
WIP to fix appveyor tests

rbuffat:
This PR:
* Ignores deprecation warning of pip for python 2.7
* Adds gdal 3.0.0 from gisinternals to build matrix. Fiona fails to build with gdal 3.0.0, which should be fixed in gdal 3.0.1)
* Adds "not wheel" marker to pytest
* Bumps dependencies of coverage, pytest, and pytest-cov to the latest versions that support python 2.7

Also a bit an unrelated change in bumping cython dependency to a 0.29.x version, which will be required for python 3.8 (or at least a few months ago python 3.8-dev on travis failed with cython 0.28.3 but succeded with cython 0.29)


rbuffat:
appveyor fails with gdal 3.0.0, but it looks like this should be fixed with gdal 3.0.1: https://github.com/OSGeo/gdal/issues/1639

But currently, gisinternals does not provide gdal 3.0.1 binaries.

rbuffat:
@sgillies  I noticed that gisinternals provides daily builds of the latest stable branch (which is currently 3.0.x). I switched to these and Fiona could be built. However, some proj related tests failed and proj.db could not be found for this build. Explicitly setting the path to the correct location fixed the tests.  With the gdal3 binaries, the proj directory was renamed to proj6. If this is a pattern that is seen often we could think about adding proj6 to PROJDataFinder.

Thus, all builds on appveyor succeed now, although I set the gdal3 build to allow failures as does not use a released version of gdal. 

I noticed that the PYTHON_VERSION variables defined in the build matrix did not match the actual Python versions on the images (E.g. PYTHON_VERSION was set to "3.6.4" but the Python version on the image was 3.6.6). PYTHON_VERSION is used in run_with_env.cmd, but it uses only the major and minor parts. Thus to avoid confusion I removed the revision part.

sgillies:
@rbuffat thank you! I'm going to see about merging this into the maint-1.9 branch early next week.

PR #807, opened 2019-10-22
use _shim3.c for gdal >3.0

rbuffat:
Use _shim3.c for gdal >= 3.0


setup.py
```diff
@@ -261,7 +261,7 @@ def run(self):
         log.info("Building Fiona for gdal 1.x: {0}".format(gdalversion))
         ext_modules.append(
             Extension('fiona._shim', ['fiona/_shim1.c'], **ext_options))
-    else:
+    elif gdal_minor_version == 2:
```
jorisvandenbossche: This needs to be `gdal_major_version` instead of `gdal_minor_version` I think

rbuffat: oh no! yes, that was my mistake!

PR #811, opened 2019-10-24
modified transform geom to accept arrays of geoms

### LINKED ISSUE #799: [discussion] optimizing `transform_geom` for repeated transformations
https://github.com/Toblerity/Fiona/blob/d6ec15f1c2e7a860bc57d35a51e36aeb48377a1a/fiona/_transform.pyx#L133

Currently, the transformer is recreated each time the geometry transform occurs. I am wondering if creating a `Transformer` class similar to the one in `pyproj` would make sense to prevent repeated re-creation of the transformer.

https://pyproj4.github.io/pyproj/stable/advanced_examples.html#repeated-transformations

Thoughts?

snowman2:
Closes #799

### DIFF
--- a/fiona/_transform.pyx
+++ b/fiona/_transform.pyx
@@ -11,7 +11,7 @@ from fiona cimport _cpl, _crs, _csl, _geometry
 from fiona._crs cimport OGRSpatialReferenceH
 from fiona._shim cimport osr_set_traditional_axis_mapping_strategy
 
-from fiona.compat import UserDict
+from fiona.compat import UserDict, DICT_TYPES
 
 
 cdef extern from "ogr_geometry.h" nogil:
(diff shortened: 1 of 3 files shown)


fiona/_transform.pyx
```diff
@@ -224,6 +199,47 @@ def _transform_geom(
...
+    options = _csl.CSLSetNameValue(
+                options, "DATELINEOFFSET", 
+                str(antimeridian_offset).encode('utf-8'))
+    if antimeridian_cutting:
+        options = _csl.CSLSetNameValue(options, "WRAPDATELINE", "YES")
+
+    factory = new OGRGeometryFactory()
 
-    return g
+    if isinstance(geom, DICT_TYPES):
```
sgillies: @snowman2 this won't work in 1.9 because we're introducing a Geometry class that is not a dict. Instead, let's see if geom is iterable using either `isinstance(geom, Iterable)` like in https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable or `try: geoms = iter(geom); except TypeError: geoms = [geom]`.

snowman2: Hmmm, I looked at your PR adding the Geometry class and I think that the Geometry class would be considered an iterable. Maybe this can be updated to check for the Geometry instance specifically?

sgillies: Right. How about `if isinstance(geom, collections.abc.Mapping)` then? The OGRGeomBuilder class gets the type and coordinates of `geom` using only its mapping interface.

snowman2: Actually, that should be accounted for with `DICT_TYPES`

https://github.com/Toblerity/Fiona/blob/d6ec15f1c2e7a860bc57d35a51e36aeb48377a1a/fiona/compat.py#L26

In that case, is it okay to leave it as is?

sgillies: Can't believe I overlooked that. Yes, this is fine.

snowman2: Sounds good 👍. I will work on adding tests and updating the docstrings later.

snowman2:
Not sure what is going on, but I see errors in the build like:
```
NOT using gdal 2.4.2 as expected; 
```

sgillies:
@snowman2 oh dear, I think the maint-1.9 build is busted. I'm looking into it now.

sgillies:
Got the Travis build sorted out :tada:

snowman2:
Nice!

PR #819, opened 2019-11-05
use crs_from_wkt for case where non-standard CRS string such as URN is used

snowman2:


tests/test_collection_crs.py
```diff
@@ -27,10 +27,21 @@ def test_collection_no_crs_wkt(tmpdir, path_coutwildrnp_shp):
 @requires_gdal2
 def test_collection_create_crs_wkt(tmpdir):
     """A collection can be created using crs_wkt"""
-    filename = str(tmpdir.join("test.shp"))
+    filename = str(tmpdir.join("test.geojson"))
```
sgillies: @snowman2 is this intended?

fiona/collection.py
```diff
@@ -147,7 +148,10 @@ def __init__(self, path, mode='r', driver=None, schema=None, crs=None,
                 if 'init' in crs or 'proj' in crs or 'epsg' in crs.lower():
                     self._crs = crs
                 else:
-                    raise CRSError("crs lacks init or proj parameter")
+                    crs_wkt = crs_to_wkt(crs)
+                    if not crs_wkt:
+                        raise CRSError("Invalid input for CRS: {}".format(crs))
+                    self._crs_wkt = crs_wkt
```
sgillies: 👍

tests/test_collection_crs.py
snowman2: Yeah. It is to match the driver.

snowman2:
It is for creating a new file using a SRID as shown in the tests. It should succeed in GDAL 3+, but fails in the current version as it expects the CRS to be a PROJ string.

fiona/collection.py
```diff
@@ -147,7 +148,10 @@ def __init__(self, path, mode='r', driver=None, schema=None, crs=None,
                 if 'init' in crs or 'proj' in crs or 'epsg' in crs.lower():
                     self._crs = crs
                 else:
-                    raise CRSError("crs lacks init or proj parameter")
+                    crs_wkt = crs_to_wkt(crs)
+                    if not crs_wkt:
+                        raise CRSError("Invalid input for CRS: {}".format(crs))
```
jorisvandenbossche:
```suggestion
                        raise CRSError("Invalid input for crs: {}".format(crs))
```

(the keyword argument is lower case?)

snowman2: Oh ha, I guess it depends on what you are referring to. I was thinking that the input was invalid for creating a CRS, but I see that can be confused with the kwarg. I was actually thinking of moving this logic inside the `crs_to_wkt` function. Thoughts?

snowman2: Updated

snowman2:
@sgillies, any reason that `self._crs_wkt = crs_to_wkt(crs or crs_wkt)` shouldn't be used in place of all of the other code setting the CRS in this PR? I believe that would address #823 as well.

snowman2:
> @sgillies, any reason that `self._crs_wkt = crs_to_wkt(crs or crs_wkt)` shouldn't be used in place of all of the other code setting the CRS in this PR? I believe that would address #823 as well.

I added a separate commit with this implemented. Let me know what you think.

fiona/collection.py
```diff
@@ -140,14 +141,8 @@ def __init__(self, path, mode='r', driver=None, schema=None, crs=None,
...
-
-            if crs_wkt:
-                self._crs_wkt = crs_wkt
-            elif crs:
-                if 'init' in crs or 'proj' in crs or 'epsg' in crs.lower():
-                    self._crs = crs
-                else:
-                    raise CRSError("crs lacks init or proj parameter")
+            if crs_wkt or crs:
+                self._crs_wkt = crs_to_wkt(crs_wkt or crs)
```
sgillies: 👍

tests/test_collection_crs.py
```diff
@@ -27,10 +29,35 @@ def test_collection_no_crs_wkt(tmpdir, path_coutwildrnp_shp):
...
     wkt = 'GEOGCS["GCS_WGS_1984",DATUM["WGS_1984",SPHEROID["WGS_84",6378137,298.257223563]],PRIMEM["Greenwich",0],UNIT["Degree",0.017453292519943295],AUTHORITY["EPSG","4326"]]'
     with fiona.open(filename, 'w', schema={'geometry': 'Point', 'properties': {'foo': 'int'}}, crs_wkt=wkt, driver='GeoJSON') as dst:
-        assert dst.crs_wkt == wkt
+        assert dst.crs_wkt.startswith('GEOGCS["WGS 84')
+
+    with fiona.open(filename) as col:
+        assert col.crs_wkt.startswith('GEOGCS["WGS 84')
```
sgillies:
```suggestion
        assert col.crs_wkt.startswith('GEOGCS["GCS_WGS_1984')
```

See https://travis-ci.org/Toblerity/Fiona/jobs/608201443#L1157.

snowman2: Ha, passes on GDAL 1 & 3, but not 2. 

Thoughts on:
```python
assert col.crs_wkt.startswith('GEOGCS["WGS 84') or col.crs_wkt.startswith('GEOGCS["GCS_WGS_1984')
```

sgillies: No, the test is skipped with GDAL 1.11 and GDAL 3.0.

snowman2: Oh ha, it skips GDAL 1.1. But, I don't think it skips 3.0:
```python
requires_gdal2 = pytest.mark.skipif(
    not gdal_version.major >= 2,
    reason="Requires GDAL 2.x")
```
> Since those WKT are equivalent, I think that's a good approach.

Sounds like a good plan to me :+1:

fiona/collection.py
jorisvandenbossche: > I was thinking that the input was invalid for creating a CRS, but I see that can be confused with the kwarg.

The reason that I was confused was because my first reaction "but fiona has no CRS (*capital letters indicating an object*) class". But of course, it still represents the notion of a CRS, so my confusion was only of short duration, so I think your original one is also fine.

fiona/collection.py
jorisvandenbossche: There are still a few places where `self._crs` is used no? (and with the above, this is no longer set?)

snowman2: That's okay. It is a property that defaults to None:
```python
    @property
    def crs(self):
        """Returns a Proj4 string."""
        if self._crs is None and self.session:
            self._crs = self.session.get_crs()
        return self._crs
```

snowman2: Ah. found what you are talking about.

snowman2: Removed redundant code.

snowman2:
:crossed_fingers:

tests/test_collection_crs.py
snowman2: Fix pushed.

fiona/ogrext.pyx
```diff
@@ -970,51 +970,18 @@ cdef class WritingSession(Session):
...
-                            log.debug("PROJ.4 to be imported: %r", proj)
-                            proj_b = proj.encode('utf-8')
-                            proj_c = proj_b
-                            OSRImportFromProj4(cogr_srs, proj_c)
-                            osr_set_traditional_axis_mapping_strategy(cogr_srs)
-
-                    else:
-                        raise ValueError("Invalid CRS")
-
-            except (ValueError, CPLE_BaseError) as exc:
```
sgillies: Wow! buh-bye old code :wave:

snowman2:
Thanks for the quick release @sgillies!

PR #821, opened 2019-11-06
ENH: Added support for geopackage metatata

### LINKED ISSUE #713: [proposal] Add support for metadata read/write for GeoPackage

snowman2:
Closes #713

### DIFF
--- a/fiona/collection.py
+++ b/fiona/collection.py
@@ -14,7 +14,7 @@ from fiona.logutils import FieldSkipLogFilter
 from fiona._crs import crs_to_wkt
 from fiona._env import get_gdal_release_name, get_gdal_version_tuple
 from fiona.env import env_ctx_if_needed
-from fiona.errors import FionaDeprecationWarning
+from fiona.errors import FionaDeprecationWarning, DataIOError
 from fiona.drvsupport import supported_drivers
 from fiona.path import Path, vsi_path, parse_path
 from six import string_types, binary_type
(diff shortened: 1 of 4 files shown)


snowman2:
If you have a different exception name in mind, just let me know. I just picked one that seemed somewhat reasonable.

snowman2:
Everything but GDAL 1.11 passes. GDAL 1.11 core dumps.

snowman2:
@sgillies Since this does not work in `GDAL` 1 and `Fiona` 2+ drops support for `GDAL` 1, thoughts on making this a `Fiona` 2.0 feature?

rbuffat:
@snowman2 
Would it be possible to check the gdal version before accessing the code that core dumps and raising a DriverError when gdal 1.x is detected?
E.g. similar to this https://github.com/Toblerity/Fiona/blob/maint-1.8/fiona/collection.py#L81-L89

snowman2:
> Would it be possible to check the gdal version before accessing the code that core dumps and raising a DriverError when gdal 1.x is detected?
> E.g. similar to this https://github.com/Toblerity/Fiona/blob/maint-1.8/fiona/collection.py#L81-L89

Sounds like a good idea - definitely worth a shot. :+1:

fiona/ogrext.pyx
```diff
@@ -1207,6 +1278,80 @@ cdef class WritingSession(Session):
         gdal_flush_cache(cogr_ds)
         log.debug("Flushed data source cache")
 
+    def set_tags(self, tags, ns=None):
```
sgillies, line 1281: @snowman2 what what you think about renaming this to update_tags to align with https://rasterio.readthedocs.io/en/latest/api/rasterio._io.html?highlight=set_tags#rasterio._io.DatasetWriterBase.update_tags?

snowman2: Sounds like a good idea to me - just pushed some changes.

fiona/collection.py
```diff
@@ -284,12 +284,12 @@ def set_tags(self, tags, ns=None):
...
         if _GDAL_VERSION_TUPLE.major < 2:
             raise GDALVersionError(
-                "set_tags requires GDAL 2+, fiona was compiled "
+                "update_tags requires GDAL 2+, fiona was compiled "
                 "against: {}".format(_GDAL_RELEASE_NAME)
             )
         if not isinstance(self.session, WritingSession):
-            raise DataIOError("Unable to set tags as not in writing mode.")
-        return self.session.set_tags(tags, ns=ns)
+            raise DataIOError("Unable to update tags as not in writing mode.")
```
sgillies, line 291: @snowman2 I feel like I'm getting dangerously close to the "while you're at it..." review anti-practice, forgive me, but can you create a new fiona.errors.UnsupportedOperation exception to use here (like rasterio's https://github.com/mapbox/rasterio/blob/master/rasterio/errors.py#L103 and like the one in Python's io module)?

snowman2: Sounds like a logical addition. I pushed the change.

snowman2:
Thanks @snorfalorpagus, @rbuffat, and @sgillies  for reviewing this PR!

PR #822, opened 2019-11-06
Use OSRSetPROJSearchPaths with GDAL 3.0

sgillies:
This is better than relying on setting PROJ_LIB, which may come too late.

Our existing tests don't check this, I need to do a pre-release wheel build check.


fiona/_shim22.pyx
```diff
@@ -133,8 +134,13 @@ cdef bint check_capability_create_layer(void *cogr_ds):
...
 
+
 cdef const char* osr_get_name(OGRSpatialReferenceH hSrs):
     return ''
 
+
 cdef void osr_set_traditional_axis_mapping_strategy(OGRSpatialReferenceH hSrs):
     OSRFixup(hSrs)
+
+cdef void set_proj_search_path(object path)
```
jorisvandenbossche:
```suggestion
cdef void set_proj_search_path(object path):
```

missing colon is causing the failures

sgillies:
Disregard the AppVeyor failure. I've done a test build of wheels for cd753c9 and those pass, so I'm confident with this PR.

fiona/_env.pyx
```diff
@@ -410,8 +413,8 @@ cdef class GDALEnv(ConfigEnv):
                         path = PROJDataFinder().search()
 
                         if path:
-                            os.environ['PROJ_LIB'] = path
-                            log.debug("PROJ data not found in environment, set to %r.", path)
+                            log.debug("PROJ data not found in environment, setting to %r.", path)
+                            set_proj_data_search_path(path)
```
sgillies: Note: for GDAL versions < 3.0, this function sets PROJ_LIB as before. For GDAL >= 3.0, it calls a new OSRSetPROJSearchPaths function that overrides PROJ_LIB and is more safe in other ways. PROJ_LIB continues to be respected by PROJ 6, but there are issues on Windows with modifying the environment within a program.

PR #826, opened 2019-11-07
Fix access violation on Windows

cgohlke:
Fix crash when importing Fiona-1.8.10 + with GDAL 3.0.2 on Windows.

### DIFF
--- a/fiona/_shim3.pyx
+++ b/fiona/_shim3.pyx
@@ -153,7 +153,7 @@ cdef void osr_set_traditional_axis_mapping_strategy(OGRSpatialReferenceH hSrs):
 
 
 cdef void set_proj_search_path(object path):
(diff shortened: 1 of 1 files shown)


sgillies:
I was unfamiliar with this style of declaration, but copied it over from
GDAL. Why doesn't it work for Windows here? Do the Windows GDAL builds
rewrite it?

cgohlke:
> copied it over from GDAL. Why doesn't it work for Windows here?

The [function declaration for CSLAddString](https://gdal.org/doxygen/cpl__string_8h.html#ae84fb70e547d924b925556b2105714d5) is `char** CSLAddString (char ** papszStrList, const char * pszNewString)`.

I think it would be a bug if GDAL uses the function with a `const char *const *` variable.

PR #835, opened 2019-11-12
Some minor modifications to .travis.yml

rbuffat:
Just some minor modifications to travis.yml:
- A  rename from "trunk" to "master" was missing
- libgdal-dev is not needed
- I assume the gdal version check supporting beta and rc versions went missing somewhere

The current matrix produces three builds for master, which takes quite some time to build. I would suggest using an explicit matrix as proposed in https://github.com/Toblerity/Fiona/pull/770 and only test master for one Python version.  Alternatives would be to not test master on travis and only on appveyor, or to try to cache the master gdal builds somehow.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ python:
   - "3.8"
 
 before_install:
-  - pip install -U pip
+  - python -m pip install -U pip
   - export PATH=$GDALINST/gdal-$GDALVERSION/bin:$GDALINST/proj-$PROJVERSION/bin:$PATH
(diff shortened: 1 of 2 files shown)


rbuffat:
@sgillies  could you empty the travis cache and rerun the tests? the cached gdal binary has some (not necessary) dependencies (e.g. to webp), that are not installed without libgdal-dev. Thus the gdal binaries need to be rebuilt.

.travis.yml
```diff
@@ -30,7 +30,6 @@ matrix:
 addons:
   apt:
     packages:
-    - libgdal-dev
```
sgillies: :+1:

sgillies:
@rbuffat thank you! I will get these changes back to maint-1.9. 

I like the suggestion to build the master branch only once. I'll do that.

Issue #851, opened 2020-01-05
DLL resolution in Python 3.8 on Windows has changed


snorfalorpagus:
The resolution of DLLs has changed in Python 3.8 for Windows.

> New in version 3.8: Previous versions of CPython would resolve DLLs using the default behavior for the current process. This led to inconsistencies, such as only sometimes searching PATH or the current working directory, and OS functions such as AddDllDirectory having no effect.

> DLL dependencies for extension modules and DLLs loaded with ctypes on Windows are now resolved more securely. Only the system paths, the directory containing the DLL or PYD file, and directories added with add_dll_directory() are searched for load-time dependencies. Specifically, PATH and the current working directory are no longer used, and modifications to these will no longer have any effect on normal DLL resolution. If your application relies on these mechanisms, you should check for add_dll_directory() and if it exists, use it to add your DLLs directory while loading your library. Note that Windows 7 users will need to ensure that Windows Update KB2533623 has been installed (this is also verified by the installer). (Contributed by Steve Dower in bpo-36085.)

* https://bugs.python.org/issue36085
* https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew
* https://docs.python.org/3/library/os.html#os.add_dll_directory

As a result the Appveyor job for 3.8 is failing:

https://ci.appveyor.com/project/sgillies/fiona/branch/master/job/5e9p53tga9ono4j3

```
python -c "import fiona"
python : Traceback (most recent call last):
At line:1 char:1
+ python -c "import fiona"
+ ~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (Traceback (most recent call last)::String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
  File "<string>", line 1, in <module>
  File "C:\Python38-x64\Lib\site-packages\fiona\__init__.py", line 85, in <module>
    from fiona.collection import BytesCollection, Collection
  File "C:\Python38-x64\Lib\site-packages\fiona\collection.py", line 9, in <module>
    from fiona.ogrext import Iterator, ItemsIterator, KeysIterator
ImportError: DLL load failed while importing ogrext: The specified module could not be found.
Command executed with exception:   File "<string>", line 1, in <module>
  File "C:\Python38-x64\Lib\site-packages\fiona\__init__.py", line 85, in <module>
    from fiona.collection import BytesCollection, Collection
[...]

PR #852, opened 2020-01-10
Python38/Windows dll load directory

rbuffat:
A start towards fixing https://github.com/Toblerity/Fiona/issues/851

This PR loads the GDAL_HOME/bin directory with add_dll_directory when fiona is imported.

We should probably think about adding other directories. Maybe somebody with experience providing wheels for Windows has some advice here @cgohlke @jorisvandenbossche


fiona/__init__.py
```diff
@@ -77,6 +77,17 @@
...
 
+
+# Add gdal dll directory on Windows and Python >= 3.8, see https://github.com/Toblerity/Fiona/issues/851
+if platform.system() == 'Windows' and (3, 8) <= sys.version_info:
+
+    # Add GDAL_HOME/bin, if present
+    gdal_home = os.getenv('GDAL_HOME', None)
+
+    if gdal_home is not None and os.path.exists(gdal_home):
+        os.add_dll_directory(os.path.join(gdal_home, "bin"))
```
rbuffat: If GDAL_HOME is set, but the directory does not exist, we should emit a warning

rbuffat:
@cgohlke I tested this issue with Fiona‑1.8.13‑cp38‑cp38‑win32.whl from https://www.lfd.uci.edu/~gohlke/pythonlibs/#fiona.

Changing fiona/__init__.py from

```
# Unofficial Windows Binaries:
# Use gdal DLLs and data files from the osgeo package
try:
    import numpy
    import osgeo
except ImportError:
    pass
```

to

```
# Unofficial Windows Binaries:
# Use gdal DLLs and data files from the osgeo package
try:
    import osgeo
    for p in osgeo.__path__:
        os.add_dll_directory(p)
except ImportError:
    pass
```

fixed the issue for me. However, I think this is not a "standard" gdal location to look for directly in Fiona.

I removed numpy, because the code silently fails to import osgeo if it is not present.

cgohlke:
>  issue with Fiona‑1.8.13‑cp38‑cp38‑win32.whl from https://www.lfd.uci.edu/~gohlke/pythonlibs/#fiona

Those binaries are compiled against GDAL from https://www.lfd.uci.edu/~gohlke/pythonlibs/#gdal. They are not meant to work with any other GDAL distribution. `os.add_dll_directory()` should not be required because `import osgeo` pre-loads the DLLs required by Fiona into the process.

rbuffat:
@cgohlke yes, you are right. Sorry for that. My problem was, that I had numpy not installed.

I installed a fresh python 3.8, with gdal and fiona from your website. With this installation I got the following message: 

```
Python 3.8.1 (tags/v3.8.1:1b293b6, Dec 18 2019, 22:39:24) [MSC v.1916 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import fiona
  File "<stdin>", line 1, in <module>
  File "C:\Python38-32\lib\site-packages\fiona\__init__.py", line 87, in <module>
    from fiona.collection import BytesCollection, Collection
  File "C:\Python38-32\lib\site-packages\fiona\collection.py", line 9, in <module>
    from fiona.ogrext import Iterator, ItemsIterator, KeysIterator
ImportError: DLL load failed while importing ogrext: The specified module could not be found.
```

I assumed it is the same reason as in https://github.com/Toblerity/Fiona/issues/851 due to the changes in Python 3.8 on Windows and the similar ImportError message. However, as I did not have numpy installed, the import of osgeo never happened.

fiona/__init__.py
```diff
@@ -77,6 +77,17 @@
     class Path:
         pass
 
+
+# Add gdal dll directory on Windows and Python >= 3.8, see https://github.com/Toblerity/Fiona/issues/851
+if platform.system() == 'Windows' and (3, 8) <= sys.version_info:
```
rbuffat: is add_dll_directory harmful when gdal dlls are already loaded? E.g. when multiple gdal binaries are present.

cgohlke:
> My problem was, that I had numpy not installed.

The issue is that the GDAL Python package does not list numpy as an install requirement when the `array_module` is enabled. https://github.com/OSGeo/gdal/blob/master/gdal/swig/python/setup.py

fiona/__init__.py
```diff
@@ -67,21 +67,62 @@
...
 
 # TODO: remove this? Or at least move it, flake8 complains.
 if sys.platform == "win32":
     libdir = os.path.join(os.path.dirname(__file__), ".libs")
     os.environ["PATH"] = os.environ["PATH"] + ";" + libdir
 
+try:
+
+    from fiona.ogrext import _bounds, _listlayers, FIELD_TYPES_MAP, _remove, \
+        _remove_layer
```
snorfalorpagus: If this line is duplicated we need to remember to change both lines if we want to amend the imports in the future. Instead you could import `fiona.ogrext` only as a test, then have the real imports just one after the import succeeds:
```python
try:
    import fiona.ogrext
except ImportError as e:
    if ...:
        import fiona.ogrext
    else:

from fiona.ogr import _bounds, _listlayers, ...
```

fiona/__init__.py
```diff
@@ -67,21 +67,62 @@
...
+        dll_directory = None
+
+        # Parse PATH for gdal/bin
+        for path in os.getenv('PATH', '').split(';'):
+            p = Path(path.lower())
+
+            if p.parts[-2:] == ('gdal', 'bin') and os.path.exists(path):
+                dll_directory = path
+                break
+            del path, p
```
snorfalorpagus: If this Windows-specific section was moved into it's own function it would have it's own scope and this type of cleanup wouldn't be required.

fiona/__init__.py
```diff
@@ -67,21 +67,62 @@
...
+
+        # Use GDAL_HOME if present
+        if dll_directory is not None:
+            gdal_home = os.getenv('GDAL_HOME', None)
```
snorfalorpagus: This will need to be added to the documentation. I don't think it's a standard runtime environment variable for GDAL, although it does seem to be used in the build guide for Windows.

snorfalorpagus:
Should this be targetted at `master` or at a `maint-*` branch?

fiona/__init__.py
```diff
@@ -90,7 +90,7 @@
 
     see https://github.com/Toblerity/Fiona/issues/851
     """
-    if platform.system() == 'Windows' and (3, 8) <= sys.version_info:
+    if platform.system() == 'Windows':
```
rbuffat: I think it should not hurt to add the same directories for all versions of Python if the import of ogrext fails

fiona/__init__.py
```diff
@@ -67,21 +67,67 @@
...
+            # Parse PATH for gdal/bin
+            for _path in os.getenv('PATH', '').split(';'):
+                p = Path(_path.lower())
+
+                if p.parts[-2:] == ('gdal', 'bin') and os.path.exists(_path):
+                    dll_directory = _path
+                    break
+
+            # Use GDAL_HOME if present
+            if dll_directory is not None:
```
snorfalorpagus: Does this mean that `GDAL_HOME` is only checked if a directory has already been found in the `PATH`? That doesn't seem right.

README.rst
```diff
@@ -300,8 +300,11 @@ the ``GDAL_VERSION`` environment variable (e.g. ``set GDAL_VERSION=2.1``).
 
     $ python setup.py build_ext -I<path to gdal include files> -lgdal_i -L<path to gdal library> install --gdalversion 2.1
 
-Note: The GDAL DLL (``gdal111.dll`` or similar) and gdal-data directory need to
-be in your Windows PATH otherwise Fiona will fail to work.
+
+Note: The directory containing the GDAL DLL (``gdal111.dll`` or similar) needs to be in your Windows ``PATH``
+(e.g. ``C:\gdal\bin``). Alternatively, ``GDAL_HOME`` needs to be set (e.g. ``C:\gdal``). A ``*\gdal\bin`` directory in ``PATH`` 
+has priority over ``GDAL_HOME``. Furthermore, the gdal-data directory needs to be in your Windows ``PATH`` or the environment variable
```
snorfalorpagus: > A ``*\gdal\bin`` directory in ``PATH`` has priority over ``GDAL_HOME``.
This isn't what the code does currently.

fiona/__init__.py
```diff
@@ -67,21 +67,67 @@
...
 # TODO: remove this? Or at least move it, flake8 complains.
 if sys.platform == "win32":
     libdir = os.path.join(os.path.dirname(__file__), ".libs")
     os.environ["PATH"] = os.environ["PATH"] + ";" + libdir
 
+
+try:
+
+    import fiona.ogrext
+
```
snorfalorpagus: I don't think the line breaks around this are necessary.

fiona/__init__.py
rbuffat: fixed

s7726:
Currently hitting this. The [GDAL installers](http://www.gisinternals.com/query.html?content=filelist&file=release-1911-x64-gdal-3-0-4-mapserver-7-4-3.zip) put the executables and libraries in the GDAL directory, not under bin. So the `add_dll_directory_win` function is a little too strict.
Also once I manually add `C:\Program Files\GDAL` using `os.add_dll_directory` I run into another issue.

```
ImportError: DLL load failed while importing ogrext: The specified procedure could not be found.
```
Took awhile to hunt down this issue to solve the specified _module_ problem. I could use some direction in how to hunt down the _procedure_ aspect.

rbuffat:
@s7726 I'm not sure what caused this error message. Maybe it helps to recompile Fiona and check that it compiles against the same gdal dlls.

@sgillies @snorfalorpagus @cgohlke  Numpy has a separate init file for distributions: https://github.com/numpy/numpy/blob/master/numpy/_distributor_init.py
Should we add the same to Fiona?  Especially on Windows, there is no standardized way to distribute GDAL binaries and we can't check every possible location. This could be a place to locate the code for distribution specific loading of GDAL.

sgillies:
@rbuffat thanks for taking the lead on this. I would prefer a more explicit approach.

Effectively, what we need to do on Windows/Python 3.8 is

```
with os.add_dll_directory(DIR):
    import extension_module
```

in every Python module that imports Fiona extension modules (_env, _geometry, etc). We could write a generic context manager that used the Windows function when required and did nothing in other cases.

```
with fiona._loading.add_dll_directories():  # or something like.
    import extension_module
```

What would you think about doing this for a 1.8 bug fix release?

rbuffat:
@sgillies is that the direction you had in mind? 

```
try:

    import fiona.ogrext
    import fiona._env

except ImportError as e:

    import fiona._loading

    with fiona._loading.add_gdal_dll_directories():

            import fiona.ogrext
            import fiona._env
```

could be simplified to

```
with fiona._loading.add_gdal_dll_directories():

    import fiona.ogrext
    import fiona._env

```

However, my concern here is that other projects could implement other ways to load the required dlls independent on environment variables. Adding then additional directories could lead to side effects.  As far as I currently see the only way to know if the dlls are not available is when the import fails.

rbuffat:
closed in favor of https://github.com/Toblerity/Fiona/pull/875

PR #854, opened 2020-01-14
Adapt travis tests to gdal master

rbuffat:
There are currently two issues with the tests and gdal master. 

The mock debian tmpdir is missing a path for gdal 3.1. I changed the creation of the mockdir to use gdal_version. However, I'm not sure if there was a reason that the paths were hardcoded and I did not notice.

The second issue is, that gdal now something returns for the  COMPD_CS projection. As far as I understand the test_compound_crs test, it is only important that a crs (empty or not) is returned and no crash happens. Thus I changed it to just test if a dict is returned. 

Finally, the PR disables that the gdal master installation is always deleted and not cached.

### DIFF
--- a/scripts/travis_gdal_install.sh
+++ b/scripts/travis_gdal_install.sh
@@ -78,7 +78,7 @@ if [ "$GDALVERSION" = "master" ]; then
         cp newrev.txt $GDALINST/gdal-$GDALVERSION/rev.txt
         cp newproj.txt $GDALINST/gdal-$GDALVERSION/newproj.txt
         ./configure --prefix=$GDALINST/gdal-$GDALVERSION $GDALOPTS $PROJOPT
-        make -j 4
+        make
         make install
     fi
 
@@ -98,7 +98,7 @@ else
         tar -xzf gdal-$GDALVERSION.tar.gz
         cd gdal-$gdalver
         ./configure --prefix=$GDALINST/gdal-$GDALVERSION $GDALOPTS $PROJOPT
-        make -j 4
+        make
         make install
     fi
 fi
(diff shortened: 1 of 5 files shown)


rbuffat:
There was a bug in the caching of gdal master that prevented a new build of gdal master when the proj version is changed until the gdal repository has a new commit. This case is probably extremely rare.

Also, I was wondering about the make -j 4 in travis_gdal_install.sh, because if I remember correctly the travis workers only have two cores. 

The following command returns the unix timestamp of the last commit. 
`TZ=UTC git show --quiet --date='unix' --format="%cd"`
We could use this command in a similar manner as the current caching mechanism to cache gdal master e.g. for one or two weeks.

sgillies:
@rbuffat let's do `make -j2` then.

rbuffat:
@sgillies I switched to MAKEFLAGS="-j 2" as suggested by https://docs.travis-ci.com/user/speeding-up-the-build/#makefile-optimization 

It looks like the build time stayed roughly the same.

scripts/travis_gdal_install.sh
```diff
@@ -64,46 +64,32 @@ if [ "$GDALVERSION" = "master" ]; then
...
-            PROJOPT="--with-static-proj4=$GDALINST/gdal-$GDALVERSION"
-            ;;
-    esac
+
+    if $(dpkg --compare-versions "$GDALVERSION" "lt" "2.3"); then
+        PROJOPT="--with-static-proj4=$PROJINST/proj-$PROJVERSION";
+    else
+        PROJOPT="--with-proj=$PROJINST/proj-$PROJVERSION";
+
+    fi
```
sgillies: @rbuffat I apologize for not commenting on these lines before, but I think we should avoid using dpkg here to make it easier to build on a macosx server in the future.

rbuffat: That is a very good point. I reverted the PR to the case and added a default case so that it should not break if a gdal version > 3 is released.

sgillies:
Thanks @rbuffat !

PR #855, opened 2020-01-19
[Low priority] Pin python dependencies to oldest working version

rbuffat:
I noticed that click >= 5 is required instead of the >= 4 requirement of setup.py. In our current CI setup we will not notice when the requirements of setup.py get outdated. Thus, I had the idea that we could pin the Python dependencies to the oldest working versions. This makes it easy to see if a change in the code depends on a newer dependency and an update of setup.py is required.

This PR does two things:
- I searched for the oldest working combinations of the Python dependencies and updated requirements.txt and setup.py accordingly.
- I added a test to the build matrices of travis and appveyor that test the newest version of the dependencies in requirements.txt. These tests are marked as allowed to fail. Testing the newest version should help to identify quickly if an old version of a dependency is the root of an error. 

I also see some counter-arguments for why this is not a good idea. E.g. it adds complexity, it increases the CI build time, Fiona's dependencies are quite stable/slowly changing (except maybe for boto3) and testing with the oldest working versions makes it the likeliest that something breaks. Thus, @sgillies, if you are not in favor I have no problems to close the PR.   

The changes in appveyor.yml look a bit messy because I changed the order of the env variables so that the relevant variables are shown first in the build matrix.


sgillies:
This is outdated. Will close.

Issue #840, opened 2019-12-03
[Feature Request] Append mode for GeoPackage and GeoJSON


arredond:
Similar to #839 

Currently, Fiona allows read and write operations for both [GeoJSON](https://github.com/Toblerity/Fiona/blob/master/fiona/drvsupport.py#L52) and [GPKG](https://github.com/Toblerity/Fiona/blob/master/fiona/drvsupport.py#L60) (two of the most common spatial formats), but not appending to an existing file of these formats.

This seems to be supported by GDAL/OGR so it'd definitely be nice to have this in Fiona too.

I'd be happy to contribute with a PR if someone could point me to some guidelines for contributing!

## Appending with `ogr2ogr`

Sample data (`square.geojson`):

```geojson
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [ [0,0], [0,10], [10,10], [10,0], [0,0] ]
        ]
      }
    }
  ]
}
```

#### GeoJSON

Creating a copy, then appending to it
```bash
cp square.geojson square_double.geojson
ogr2ogr -f geojson -append -update square_double.geojson square.geojson
```

Checking contents:
```bash
fio info square.geojson | jq .counts # 1 feature
fio info square_double.geojson | jq .counts # 2 features
```

#### GeoPackage

Same procedure with GeoPackage. Converting original data:
```bash
ogr2ogr -f gpkg square.gpkg square.geojson
```

Creating a copy, then appending to it
```bash
cp square.gpkg square_double.gpkg
ogr2ogr -f gpkg -append -update square_double.gpkg square.gpkg
```

Checking contents:
```bash
fio info square.gpkg | jq .counts # 1 feature
fio info square_double.gpkg | jq .counts # 2 features
```

## Operating system

Mac OS Catalina 10.15.1

## Fiona and GDAL version and provenance

Fiona: 1.8.11 installed from PyPI using pip version 19.3.1.
GDAL: 2.4.2 installed via Homebrew

snorfalorpagus:
Closed by #858

Issue #839, opened 2019-12-01
GeoJSONSeq append to file


kylebarron:
## Expected behavior and actual behavior.

Appending to a GeoJSON newline-delimited file is not currently supported.

https://github.com/Toblerity/Fiona/blob/2ec38d087fea72c8bd0e7696d8ac1a6203df8851/fiona/drvsupport.py#L54

I can't find any other issues mentioning appending using GeoJSONSeq, and given the nature of the format, it seems like that should be possible. Where would a PR to enable that take place?

## Steps to reproduce the problem.

```py
import fiona
import geopandas as gpd
from geopandas.io.file import infer_schema
from shapely.geometry import Point

gdf = gpd.GeoDataFrame(geometry=[Point([0, 1])])
schema = infer_schema(gdf)
with fiona.open('test.geojson', "w", driver='GeoJSONSeq', schema=schema) as colxn:
    colxn.writerecords(gdf.iterfeatures())

gdf = gpd.GeoDataFrame(geometry=[Point([2, 2])])
schema = infer_schema(gdf)
with fiona.open('test.geojson', "a", driver='GeoJSONSeq', schema=schema) as colxn:
    colxn.writerecords(gdf.iterfeatures())
```

```
---------------------------------------------------------------------------
DriverError                               Traceback (most recent call last)
<ipython-input-26-01e338412202> in <module>
----> 1 with fiona.open('test.geojson', "a", driver='GeoJSONSeq', schema=schema) as colxn:
      2 
      3     colxn.writerecords(gdf.iterfeatures())
      4 

~/local/anaconda3/envs/create-database/lib/python3.7/site-packages/fiona/env.py in wrapper(*args, **kwargs)
    405             with Env.from_defaults(session=session):
    406                 log.debug("Credentialized: {!r}".format(getenv()))
--> 407                 return f(*args, **kwargs)
    408     return wrapper
    409 

~/local/anaconda3/envs/create-database/lib/python3.7/site-packages/fiona/__init__.py in open(fp, mode, driver, schema, crs, encoding, layer, vfs, enabled_drivers, crs_wkt, **kwargs)
    252         if mode in ('a', 'r'):
    253             c = Collection(path, mode, driver=driver, encoding=encoding,
--> 254                            layer=layer, enabled_drivers=enabled_drivers, **kwargs)
    255         elif mode == 'w':
    256             if schema:

[...]

PR #858, opened 2020-02-10
Added the append mode to GPKG and GeoJSON

gHerzmann:
I was trying to append to a GeoPackage file but I was receiving a DriverError: unsupported mode: 'a'. I found it strange since the append operations seems to be supported by GDAL/OGR. Searching the issues page I found issues #839 and #840 and decided to figure out a way to make it work.

I decided to start by simply trying to add the append mode to the GPKG and GeoJSON lines in fiona/drvsupport.py and then I started doing some tests. It seems it is now able to append correctly to those two formats.

This is my first PR so feel free to guide me in the right direction if I'm doing something wrong.


DavidMStraub:
Just ran into this issue today as well, great someone is addressing this. Would be great to see this merged.

sgillies:
Great, thanks @gHerzmann !

PR #860, opened 2020-02-16
Add drvsupport tests

rbuffat:
Currently, the selection of drivers and modes in drvsupport.py is not tested. Thus, it is hard to keep track of changes in gdal.

This PR adds the following:
* add fiona.open kwarg `fiona_force_driver` to overwrite driver and mode selection of drvsupport.py
* Check if gdal version is recent enough for write or append mode, raising of a DriverError Exception if not 
* a new test to check if append mode works (currently "special" driver such as CSV or GPX are excluded from this test) 
* a new test to check if "rw" drivers do not support append
* a new test to check if "r" mode driver cannot write

Due to these tests, the following changes of drvsupport.py were necessary for the 1.8-maint branch:
* BNA "raw" -> "rw"
* GML "raw" -> "rw"
* GeoJSON "rw" -> "raw"
~* GeoJSONSeq "rw" -> "raw"~
* GPKG "rw" -> "raw"
* PCIDSK "r" -> "raw"

BNA: `On export all layers are written to a single BNA file. Update of existing files is not currently supported.`
GML: `OGR has limited support for GML reading and writing. Update of existing files is not supported.`

What is missing:
* Documentation of fiona_force_driver
* Refactoring / choice of better names
* Current testing is quite rudimentary. The append test could fail despite a working append mode because of a wrong extension, or some special requirements of a driver.


rbuffat:
If the file extension is not `.geojsons`, the driver defaults to the json driver. This is probably related to https://github.com/Toblerity/Fiona/issues/839

docs/manual.rst
```diff
@@ -719,7 +719,7 @@ Features Access) installed, you can see this in by verifying the following.
    clean dirty data, so make sure you're getting yours from a clean source.
 
 Writing Vector Data
-===================
+===================with fiona.open("file.kmz", layer="QFESCurrentIncidents") as collection:
```
sgillies:
```suggestion
===================
```

docs/manual.rst
```diff
@@ -1318,6 +1318,34 @@ The single shapefile may also be accessed like so:
...
+It should, however, first be verified, if the local installation of GDAL/OGR 
+includes the required driver:
+
+.. sourcecode:: python
+
+    from fiona.env import Env
+
+    with Env() as gdalenv:
+        print(gdalenv.drivers().keys())
+
```
sgillies: This is great 🙏

fiona/collection.py
```diff
@@ -75,11 +75,18 @@ def __init__(self, path, mode='r', driver=None, schema=None, crs=None,
...
+            min_gdal_version = ".".join(list(map(str, driver_mode_mingdal[mode][driver])))
+
             raise DriverError(
-                "GPKG driver requires GDAL 1.11.0, fiona was compiled "
-                "against: {}".format(get_gdal_release_name()))
+                "{driver} driver requires at least GDAL {min_gdal_version} for mode '{mode}', "
+                "Fiona was compiled against: {gdal}".format(driver=driver,
+                                                            mode=mode,
+                                                            min_gdal_version=min_gdal_version,
+                                                            gdal=get_gdal_release_name()))
```
sgillies: 👍

fiona/drvsupport.py
```diff
@@ -141,6 +141,23 @@
...
+
+    'r': {'GPKG': (1, 11, 0)},
+
+    'w': {'GPKG': (1, 11, 0),
+          'PCIDSK': (2, 0, 0)},
+
+    'a': {'GMT': (2, 0, 0),
+          'GPKG': (1, 11, 0),
+          'GeoJSON': (2, 1, 0),
+          'GeoJSONSeq': (2, 0, 0),
```
sgillies: @rbuffat I am not sure that the GeoJSONSeq driver actually supports appending. There was a brief exchange about this on the gdal-dev list a few weeks ago. Also I think the min version might be 2.3.

rbuffat: yes, I forget to remove this. Before I changed GeoJSONSeq to test for the original driver it missed that it defaulted to the GeoJSON driver. I added a test to ensure the modes of driver_mode_mingdal are also present in supported_drivers.

rbuffat:
I went again over the tests to structure them a bit clearer.  Currently, no driver needs to be blacklisted. But CSV, DGN and DXF are probably the most likely to cause troubles, as they don't really follow the data model of OGR.  Thus, I think it is good to keep the blacklists in place so that these drivers can easily be blacklisted when they cause trouble in the future.

DXF and DGN seem not to support append.  The tests could fail here due to wrong schemas/records, but as far as I see the documentation of these drivers neither state that update/append is supported nor that it is not supported.

tests/test_drvsupport.py
```diff
@@ -2,13 +2,363 @@
...
+
+    path = str(tmpdir.join(get_temp_filename(driver)))
+
+    with pytest.raises(Exception):
+        with fiona.open(path, 'w',
+                        driver=driver,
+                        schema=get_schema(driver)) as c:
+            c.write(get_record1(driver))
+
+    supported_drivers[driver] = backup_mode
```
rbuffat: This line is not executed if pytest.raises(Exception) fails. Is there a simple way to setup and teardown the state of supported_drivers?

sgillies: @rbuffat how about using pytest's [monkeypatch fixture](https://docs.pytest.org/en/latest/monkeypatch.html)? I think that will let you do 

`monkeypatch.setitem(fiona.drvsupport.supported_drivers, driver, 'rw')`

once at the very start of the function. I think it would require one import change in the test module.

rbuffat: I didn't know about monkeypatch. This looks really handy!

rbuffat:
One last thing that could be done is to limit the tests for certain branches. For old versions of GDAL, the drivers should not really change, thus the master branch would the best candidate.  On the other hand, the setup of the tests is longer than running all tests, thus this would not make a big impact.

sgillies:
@rbuffat let's do that in a different PR? I'd love to get this merged.

rbuffat:
@sgillies  I think it is good to go.

Issue #861, opened 2020-02-21
Bring OGR error messages to Python exceptions


sgillies:
At places like https://github.com/Toblerity/Fiona/blob/master/fiona/ogrext.pyx#L1166-L1168 we are testing for success or failure of OGR API functions but are not using messages that OGR has put on the error stack and which might be helpful to users. Let's change this in 1.9.

sgillies:
Resolved by #866.

PR #866, opened 2020-02-25
Expose OGR error messages

rbuffat:
This PR exposes OGR error messages by wrapping ogr api functions returning OGRErr with exc_wrap_int. (See https://github.com/Toblerity/Fiona/issues/861)

Currently, this works for all functions except, OGR_L_GetExtent, OGR_DS_DeleteLayer, and OGR_L_SetIgnoredFields.

~It seems as wrapping OGR_L_GetExtent reveals a bug in the flush() method of Collection:
https://travis-ci.org/rbuffat/Fiona/jobs/654828442#L4458~

It looks like OGR_L_GetExtent returns OGRERR_FAILURE more often that is currently assumed.

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -775,7 +775,11 @@ cdef class Session:
             raise ValueError("Null layer")
 
         result = OGR_L_GetExtent(self.cogr_layer, &extent, 1)
-        return (extent.MinX, extent.MinY, extent.MaxX, extent.MaxY)
+        
+        if result == OGRERR_NONE:
+            return (extent.MinX, extent.MinY, extent.MaxX, extent.MaxY)
+        else:
+            return None
 
     def has_feature(self, fid):
         """Provides access to feature data by FID.
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -488,7 +488,7 @@ cdef class Session:
                     except AttributeError:
                         raise TypeError("Ignored field \"{}\" has type \"{}\", expected string".format(name, name.__class__.__name__))
                     ignore_fields = CSLAddString(ignore_fields, <const char *>name_b)
-                exc_wrap_int(OGR_L_SetIgnoredFields(self.cogr_layer, <const char**>ignore_fields))
+                OGR_L_SetIgnoredFields(self.cogr_layer, <const char**>ignore_fields)
             finally:
                 CSLDestroy(ignore_fields)
 
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -1454,7 +1454,7 @@ def _remove_layer(path, layer, driver=None):
     except (DriverError, FionaNullPointerError):
         raise DatasetDeleteError("Failed to remove data source {}".format(path))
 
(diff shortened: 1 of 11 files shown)


rbuffat:
Regarding the behavior of OGR_L_GetExtent, the following code:

```
import fiona

_records_with_float_property1 = [
    {
        'geometry': {'type': 'Point', 'coordinates': (0.0, 0.0)},
        'properties': {'property1': 12.22}
    },
    {
        'geometry': {'type': 'Point', 'coordinates': (0.0, 10.0)},
        'properties': {'property1': 12.88}
    }
]

with fiona.open(
            "test.geojson",
            "w",
            driver="GeoJSON",
            schema={'geometry': 'Point', 'properties': [('property1', 'int')]},
            crs='epsg:4326',
            encoding='utf-8'
    ) as c: 
    c.writerecords(_records_with_float_property1)
    print(c.bounds)

with fiona.open(
            "test.geojson",
            "r",
) as c2:
    print(c2.bounds)
```

gives with Fiona 1.8.13 the following output:

```
(0.0, 0.0, 0.0, 0.0)
(0.0, 0.0, 0.0, 10.0)
```

It could be a driver-specific issue.

rbuffat:
Regarding the behavior of OGR_L_GetExtent, the following code:

```
import fiona

_records_with_float_property1 = [
    {
        'geometry': {'type': 'Point', 'coordinates': (0.0, 0.0)},
        'properties': {'property1': 12.22}
    },
    {
        'geometry': {'type': 'Point', 'coordinates': (0.0, 10.0)},
        'properties': {'property1': 12.88}
    }
]

with fiona.open(
            "test.geojson",
            "w",
            driver="GeoJSON",
            schema={'geometry': 'Point', 'properties': [('property1', 'int')]},
            crs='epsg:4326',
            encoding='utf-8'
    ) as c: 
    c.writerecords(_records_with_float_property1)
    print(c.bounds)

with fiona.open(
            "test.geojson",
            "r",
) as c2:
    print(c2.bounds)
```

gives with Fiona 1.8.13 the following output:

```
(0.0, 0.0, 0.0, 0.0)
(0.0, 0.0, 0.0, 10.0)
```

It could be a driver-specific issue.

rbuffat:
Output for GPKG:

```
(0.0, 0.0, 0.0, 10.0)
(0.0, 0.0, 0.0, 10.0)
```

Output for Shapefile:

```
(0.0, 10.0, 0.0, 10.0)
(0.0, 0.0, 0.0, 10.0)
```

fiona/ogrext.pyx
```diff
@@ -775,7 +775,11 @@ cdef class Session:
             raise ValueError("Null layer")
 
         result = OGR_L_GetExtent(self.cogr_layer, &extent, 1)
-        return (extent.MinX, extent.MinY, extent.MaxX, extent.MaxY)
+        
+        if result == OGRERR_NONE:
+            return (extent.MinX, extent.MinY, extent.MaxX, extent.MaxY)
+        else:
+            return None
```
rbuffat: I'm not sure if this is the right thing to do, as for a user a None bound is probably not expected. Raising an exception would require the modifications of some tests.

rbuffat:
The issue with OGR_L_GetExtent seems to be more complex. I will create a separate PR.

PR #867, opened 2020-02-26
Bounds can not always be calculated

rbuffat:
Follow up to https://github.com/Toblerity/Fiona/pull/866 regarding OGR_L_GetExtent after writing of features. 

Currently, there are two issues:
1. Calculation of extents can fail
2. Correct calculation of extents after writing a record is driver dependant: https://github.com/OSGeo/gdal/issues/2269

Regarding 1:
- This PR disables the computations of bounds after records are written. They are lazy evaluated.
- When a bound cannot be calculated, a DriverError is raised. This leads to failed tests.

Regarding 2:
- We probably should detect cases when the calculation of bounds is not possible. ~Unfortunately, it seems as GDAL does not raise an error in this case. We could blacklist drivers.~

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -777,7 +777,8 @@ cdef class Session:
         result = OGR_L_GetExtent(self.cogr_layer, &extent, 1)
         
         if result != OGRERR_NONE:
-            raise DriverError("Driver was not able to calculate extent")
+            raise DriverError("Driver was not able to calculate bounds")
+
         return (extent.MinX, extent.MinY, extent.MaxX, extent.MaxY)
 
     def has_feature(self, fid):
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -775,7 +775,11 @@ cdef class Session:
             raise ValueError("Null layer")
 
         result = OGR_L_GetExtent(self.cogr_layer, &extent, 1)
-        return (extent.MinX, extent.MinY, extent.MaxX, extent.MaxY)
+        
+        if result == OGRERR_NONE:
+            return (extent.MinX, extent.MinY, extent.MaxX, extent.MaxY)
+        else:
+            raise DriverError("Driver was not able to calculate extent")
 
     def has_feature(self, fid):
         """Provides access to feature data by FID.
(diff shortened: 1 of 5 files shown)


rbuffat:
It looks as if when OGR_L_GetExtent returns a failure reliably detects when the extent cannot be computed, except for the Mapinfo driver. But for this driver, its special behavior regarding extents is documented: https://gdal.org/drivers/vector/mitab.html

tests/test_bounds.py
```diff
@@ -1,5 +1,9 @@
+import pytest
 import fiona
-
+from fiona.drvsupport import supported_drivers
+from fiona.errors import DriverError
+from .conftest import driver_extensions, requires_gdal2
```
snorfalorpagus: `requires_gdal2` is not used

fiona/ogrext.pyx
```diff
@@ -775,7 +775,11 @@ cdef class Session:
             raise ValueError("Null layer")
 
         result = OGR_L_GetExtent(self.cogr_layer, &extent, 1)
-        return (extent.MinX, extent.MinY, extent.MaxX, extent.MaxY)
+        
+        if result == OGRERR_NONE:
+            return (extent.MinX, extent.MinY, extent.MaxX, extent.MaxY)
+        else:
+            raise DriverError("Driver was not able to calculate extent")
```
snorfalorpagus: It's only a matter of style, but my preference is to indent the errors and keep the happy path unindented (as far as possible). This also means the expected return is at the end of the function. e.g.:
```suggestion
        if result != OGRERR_NONE:
            raise DriverError("Driver was not able to calculate extent")
        return (extent.MinX, extent.MinY, extent.MaxX, extent.MaxY)
```

tests/test_bounds.py
rbuffat: merci

fiona/ogrext.pyx
```diff
@@ -775,6 +775,9 @@ cdef class Session:
             raise ValueError("Null layer")
 
         result = OGR_L_GetExtent(self.cogr_layer, &extent, 1)
+        
+        if result != OGRERR_NONE:
+            raise DriverError("Driver was not able to calculate extent")
```
rbuffat: maybe "Driver was not able to calculate bounds" would be better, as Collection has a bounds property and not an extent property?

snorfalorpagus: I agree that "bounds" would be more useful as that is the terminology we expose to the user.

PR #873, opened 2020-04-01
cmd line example for Windows pip install

s7726:
Adding an example for how to get fiona installed via pip.


sgillies:
Thanks @s7726 !

PR #880, opened 2020-04-15
[WIP] Add function to print debug information

rbuffat:
This PR adds a function to print information that is typically useful when creating a new issue, such as the version of fiona, gdal and proj, but also the operating system and Python.

```
>>> fiona.print_debug_information()
Fiona version: 1.8.13.post1
GDAL version: 3.0.4
PROJ version: 6.3.1

OS: Linux 5.6.3-arch1-1
Python: 3.8.2
```

Since gdal 3.0.1 it is possible to query the proj version using OSRGetPROJVersion.  I did not create a separate shim for gdal 3.0.1, thus this breaks compatibility with gdal 3.0.0.  I was not sure if creating a shim for gdal 3.0.1 is worth it.

As OSRGetPROJVersion is not available for older versions of gdal, I created dummy functions in the gdal 1.x and 2.x shims for get_proj_version. I'm not sure if this is the correct way.

I'm not creative with function names, there is probably a better name as print_debug_information. Maybe print_environment()?

This function could also be extended to include information about GDAL_DATA and PROJ_LIB.


jorisvandenbossche:
This would certainly be useful!

> I'm not creative with function names, there is probably a better name as print_debug_information. Maybe print_environment()?

For reference, there are similar methods in some other packages. Eg both pyproj (https://github.com/pyproj4/pyproj/blob/master/pyproj/_show_versions.py) and geopandas (https://github.com/geopandas/geopandas/blob/master/geopandas/tools/_show_versions.py) call this `show_versions()`. So that's another option.

And to illustrate, they currently give this output:

```
>>> import pyproj    
>>> pyproj.show_versions()   
pyproj info:
    pyproj: 2.6.0
      PROJ: 6.3.0
  data dir: /home/joris/miniconda3/envs/geo-dev/share/proj

System:
    python: 3.8.1 | packaged by conda-forge | (default, Jan 29 2020, 14:55:04)  [GCC 7.3.0]
executable: /home/joris/miniconda3/envs/geo-dev/bin/python
   machine: Linux-4.15.0-1079-oem-x86_64-with-glibc2.10

Python deps:
       pip: 20.0.2
setuptools: 45.1.0.post20200119
    Cython: 0.29.14

>>> import geopandas     
>>> geopandas.show_versions()   

SYSTEM INFO
-----------
python     : 3.8.1 | packaged by conda-forge | (default, Jan 29 2020, 14:55:04)  [GCC 7.3.0]
executable : /home/joris/miniconda3/envs/geo-dev/bin/python
machine    : Linux-4.15.0-1079-oem-x86_64-with-glibc2.10

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.8.0
GEOS lib   : /home/joris/miniconda3/envs/geo-dev/lib/libgeos_c.so
GDAL       : 3.0.4
GDAL data dir: /home/joris/miniconda3/envs/geo-dev/share/gdal
PROJ       : 6.3.0
PROJ data dir: /home/joris/miniconda3/envs/geo-dev/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.7.0+19.gedf6e96
pandas     : 1.0.2
fiona      : 1.8.13
numpy      : 1.18.1
shapely    : 1.7.0
rtree      : 0.9.3
pyproj     : 2.6.0
matplotlib : 3.1.2
mapclassify: None
pysal      : None
geopy      : 1.21.0
psycopg2   : None
```

rbuffat:
I think it would be useful to include GDAL_DATA and PROJ_LIB paths, but I'm unsure how to detect them reliably. Also, I think they are not always exposed: https://github.com/Toblerity/Fiona/blob/4555731b4fed6237dfbb3d0b3d03a62e40d6c60e/fiona/env.py#L608

requirements-dev.txt
```diff
@@ -1,6 +1,6 @@
 -r requirements.txt
 coverage==4.5.4
-cython==0.29.7
+cython==0.29.16
```
sgillies, line 3: 👍

PR #882, opened 2020-04-15
Sync travis of maint-1.8 with master

rbuffat:
This PR backports changes in travis.yml applied to master to maint-1.8. 

As the build matrix with 8 gdal versions and 3 python version would be immense, I would suggest using a cross pattern:
* Testing all supported gdal versions with a single version of Python
* Testing the latest stable gdal release with all supported python versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,6 @@
-dist: trusty
-
 language: python
-
-python:
-  - "2.7"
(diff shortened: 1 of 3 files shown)


sgillies:
Sure! For a long time I'd been holding on to those ancient combinations of old Python x old GDAL, but let's move on.

Issue #813, opened 2019-10-28
Bug when using filter + start:stop out of range with GPKG


culebron:
I wonder if this is solvable in Fiona, or it's a Geos bug. If so, sorry for inconvenience.

Here's a test:

1. make a geojson file in QGIS with 10 features
2. also save it as geopackage
3. make a filter that overlaps some or all of them
4. open those files in fiona: `jsn = fiona.open('features.geojson')`, `pkg = fiona.open('features.gpkg')`.
4. try a combination of filter+slice for both formats, e.g. `jsn.filter(2, 20, mask=mask_object)` . and iter them.

This prints records starting from the 3rd one. If `start` is out of range, it works correctly with GeoJSON, but not with GeoPackage.

Example files:

![Selection_070](https://user-images.githubusercontent.com/119675/67723251-07459e80-fa0e-11e9-86f4-347197b93b09.jpg)

Features GeoJSON:

	{
	"type": "FeatureCollection",
	"name": "features",
	"crs": { "type": "name", "properties": { "name": "urn:ogc:def:crs:OGC:1.3:CRS84" } },
	"features": [
	{ "type": "Feature", "properties": { "fid": 1 }, "geometry": { "type": "Point", "coordinates": [ 82.840822, 55.076574 ] } },
	{ "type": "Feature", "properties": { "fid": 2 }, "geometry": { "type": "Point", "coordinates": [ 85.065732, 56.535160 ] } },
	{ "type": "Feature", "properties": { "fid": 3 }, "geometry": { "type": "Point", "coordinates": [ 86.066942, 55.319960 ] } },
	{ "type": "Feature", "properties": { "fid": 4 }, "geometry": { "type": "Point", "coordinates": [ 84.898864, 56.412274 ] } },
	{ "type": "Feature", "properties": { "fid": 5 }, "geometry": { "type": "Point", "coordinates": [ 83.656622, 53.397293 ] } },
	{ "type": "Feature", "properties": { "fid": 6 }, "geometry": { "type": "Point", "coordinates": [ 86.011319, 54.596072 ] } },
	{ "type": "Feature", "properties": { "fid": 7 }, "geometry": { "type": "Point", "coordinates": [ 86.956906, 53.771514 ] } },
	{ "type": "Feature", "properties": { "fid": 8 }, "geometry": { "type": "Point", "coordinates": [ 85.139896, 52.277353 ] } },
[...]

PR #884, opened 2020-04-19
Fix slice range does not overflow available data

rbuffat:
This is a fix for https://github.com/Toblerity/Fiona/issues/813

Currently, we do not test if slices overflow the available data range. Depending on the driver, this results in different behavior. 

This PR adds a check if the start of a slices is within the feature count, and adapts it if not.
In these cases, it is important to have the number of features available. Thus, we need to force GetFeatureCount for Layers that do not support OLC_FASTFEATURECOUNT. This can bring a substantial performance penalty (potential full table scan).


tests/test_non_counting_layer.py
```diff
@@ -33,5 +33,5 @@ def test_slice(self):
             assert len(features) == 3
 
     def test_warn_slice_negative_index(self):
-        with pytest.warns(RuntimeWarning):
+        with pytest.warns((FionaDeprecationWarning, RuntimeWarning)) as record:
```
rbuffat, line 36: Stacking two pytest.warns does not work, but this does.

rbuffat:
I had a look at the coverage report using https://github.com/Toblerity/Fiona/pull/885 and modified the slice test to cover more cases. Thereby I noticed that the GeoJSON driver for gdal 2.1-2.2 returns the last feature if the index is set by OGR_L_SetNextByIndex to an index greater than the feature count.
Debug output can be seen here: https://travis-ci.org/github/rbuffat/Fiona/jobs/681455774

I fixed that by adding a check if next_index is greater ftcount. However, this fix requires that the feature count does not change over the lifetime of an Iterator. This is a new assumption. GDAL documentation says for OGR_L_GetFeatureCount: `Returns the number of features in the layer. For dynamic databases the count may not be exact.`

rbuffat:
"Just one last thing" I noticed that FeatureBuilder in __next__ of Iterator is wrapped in a try/finally while it isn't in __next__ of ItemsIterator. I stumbled on it because originally cython coverage didn't mark the finally block as covered, but it is executed. This is maybe a bug in cython coverage: https://github.com/cython/cython/issues/3567

sgillies:
@rbuffat thank you!

PR #885, opened 2020-04-19
Add coverage for cython pyx files

rbuffat:
It would be nice to have coverage also for code in pyx files.

This PR enables coverage analysis for cython files if the environment variable CYTHON_TRACING="true" is set. 

Coverage analysis can slow down the execution of cython code, thus it is important to be deactivated by default:
https://cython.readthedocs.io/en/latest/src/tutorial/profiling_tutorial.html#enabling-coverage-analysis


rbuffat:
Just noticed that many moons ago, rasterio had coverage for pyx but it was buggy: https://github.com/mapbox/rasterio/issues/526

sgillies:
Thank you again @rbuffat !

PR #889, opened 2020-04-20
ENH: Add include_fields option for Collection

### LINKED ISSUE #887: PROPOSAL: Add ability to specify fields to load
Currently, you can define which fields to not load in using `ignore_fields`:
https://github.com/Toblerity/Fiona/blob/2c1cef667160e226f6ee5af1cbb0815d1295d5bf/fiona/collection.py#L37
https://github.com/Toblerity/Fiona/blob/2c1cef667160e226f6ee5af1cbb0815d1295d5bf/fiona/ogrext.pyx#L129

Thoughts on adding a parameter with a name like `fields` or `include_fields` where only these fields will be loaded in?

Currently in `geopandas` this is a feature that would be nice to add. However, doing so though the Fiona API (which would be preferable and more efficient) would require first opening the file and getting a list of fields and then removing the fields not in the list the user specified. Then, you would have to open the file again with the list of fields to ignore. 

Ref: https://github.com/geopandas/geopandas/pull/1383

cc: @brendan-ward

snowman2:
Closes #887

Same as #888, but for maint-1.9.

Was tinkering and found this worked. Figured it would be useful to add to the discussion.

### DIFF
--- a/fiona/collection.py
+++ b/fiona/collection.py
@@ -35,7 +35,7 @@ class Collection(object):
     def __init__(self, path, mode='r', driver=None, schema=None, crs=None,
                  encoding=None, layer=None, vsi=None, archive=None,
                  enabled_drivers=None, crs_wkt=None, ignore_fields=None,
-                 ignore_geometry=False,
+                 ignore_geometry=False, include_fields=None,
                  **kwargs):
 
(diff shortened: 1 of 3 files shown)


snowman2:
Thanks @sgillies 👍

Issue #744, opened 2019-04-18
Support milliseconds for time / datetime types


jorisvandenbossche:
Fiona is currently using `OGR_F_GetFieldAsDateTime` to get the values to construct date/time/datetime values:

https://github.com/Toblerity/Fiona/blob/be02274a53da81a68eb92d5c2a4d7a163ebb1e23/fiona/ogrext.pyx#L241-L243

By using `OGR_F_SetFieldDateTimeEx` (https://www.gdal.org/ogr__api_8h.html#a748995fa28574b25f6b723013a405a8f), it could support milliseconds (the only difference between both is that the `&ss` seconds is now a float instead of int.

Would this be a welcome change?

sgillies:
Yes @jorisvandenbossche this would be very welcome.

rbuffat:
Support for milliseconds is introduced with Fiona 1.8.14

jorisvandenbossche:
This was implemented in #893, so I think this can be closed?

PR #893, opened 2020-05-01
Add support for milliseconds

rbuffat:
This PR adds support for milliseconds, as suggested in https://github.com/Toblerity/Fiona/issues/744

Additionally, the PR refactors test_datetime.py. In contrast to before, the output of drivers that convert date, datetime or time silently to str are not tested anymore. Otherwise, the test would grow immense in size for something we cannot really control.
=> https://github.com/rbuffat/Fiona/blob/917e208aab98b1d3b61a8628b771b241e30f516c/tests/test_datetime.py
On the plus side, the test covers now datetime.* objects.


fiona/_shim1.pyx
```diff
@@ -136,3 +136,21 @@ cdef void set_proj_search_path(object path):
 
 cdef (int, int, int) get_proj_version():
     return (-1, -1, -1)
+
+
+cdef void set_field_datetime(void *cogr_feature, int iField, int nYear, int nMonth, int nDay, int nHour, int nMinute, int nSecond, int nMicrosecond, int nTZFlag):
+    OGR_F_SetFieldDateTime(cogr_feature, iField, nYear, nMonth, nDay, nHour, nMinute, nSecond, nTZFlag)
```
rbuffat, line 142: Should we warn if nMicrosecond > 0 and is discarded?

fiona/_shim2.pyx
```diff
@@ -138,3 +138,28 @@ cdef void set_proj_search_path(object path):
...
+    cdef int nMonth = 0
+    cdef int nDay = 0
+    cdef int nHour = 0
+    cdef int nMinute = 0
+    cdef float fSecond = 0.0
+    cdef int nTZFlag = 0
+
+    retval = OGR_F_GetFieldAsDateTimeEx(cogr_feature, iField, &nYear, &nMonth, &nDay, &nHour, &nMinute, &fSecond, &nTZFlag)
+    nSecond = int(fSecond)
+    nMicrosecond = int(round((fSecond%1) * 10**6))
```
rbuffat, line 163: I'm not that comfortable with the round() here. However, it was necessary, as e.g. 5.220 seconds is converted to 5.2199... internally by gdal.  (Or I do something wrong)

fiona/_shim3.pyx
```diff
@@ -15,7 +15,7 @@ cdef extern from "ogr_srs_api.h" nogil:
     void OSRSetPROJSearchPaths(const char *const *papszPaths)
 
 
-from fiona.ogrext2 cimport *
+from fiona.ogrext3 cimport *
```
rbuffat, line 18: This is unrelated to this PR and should be changed anyhow.

rbuffat:
While expanding the date / datetime/ time tests to include more drivers I noticed that it would be advantageous to refactor the checks in collection.py if a driver converts a datetime field type to string and if a driver supports a datetime field type to drvsupport. This allows them to be also used in the tests.

rbuffat:
This is probably too much for this PR but somewhat related: some drivers (such as geojson for gdal 1.x or csv) convert date / datetime / time silently to string and do not follow a standard format. As Fiona already parses the datetimes, it would not take much to convert date/datetime/time to string in Fiona using isoformat(). This would ensure a consistent conversion among all drivers. 

I created a sample implementation based on this PR: https://github.com/rbuffat/Fiona/compare/add_support_for_milliseconds...rbuffat:test_str_convertion_format
If this is something that should be included in Fiona I can create a separate PR.

sgillies:
@rbuffat I think a function that can take either seconds or microseconds is a little awkward to use. I'd prefer that set_field_datetime and get_field_datetime take float seconds (allowing microseconds) if possible.

rbuffat:
@sgillies  I refactored set_field_datetime and get_field_datetime to take seconds as floats instead of seconds and microseconds as ints. 

Additionally, I refactored driver_converts_field_type_silently_to_str so that it is possible to test gdals behavior (and thus update the function if gdal behavior changes with future versions).

What is your opinion about using Pythons isoformat to convert date / datetime / time fields to string for drivers that convert these types to string silently?
 
I'm really not creative with names. driver_converts_field_type_silently_to_str is a horribly wrong function name. If you have any suggestion to rename it I will be gladly in favor.

sgillies:
@rbuffat I think the long name is fine. We could also make it "private" by changing the name to `_driver_converts_field_type_silently_to_str`.

> What is your opinion about using Pythons isoformat to convert date / datetime / time fields to string for drivers that convert these types to string silently?

I think I don't understand the situation well enough to say. Can you show me an example of the formats that these drivers use for datetimes?

So many changes have been made to the tests that I can't be sure how much behavior has changed. Is there any way you can revert the formatting changes?

rbuffat:
I modified the tests so that also drivers are tested that convert date / datetime / time to string.

Examples of conversions that are different from the isoformat formatting are the following:
- CSV: "2018-03-25T22:49:05" => "2018/03/25 22:49:05"
- PCIDSK: "22:49:05" => '0000/00/00 22:49:05'

rbuffat:
Manually conversion of the types to string would be quite easy to achieve:

The following code
```
                    # Add microseconds to seconds
                    ss += ms / 10**6
                    set_field_datetime(cogr_feature, i, y, m, d, hh, mm, ss, 0)
```

needs to be replaced with:

```
                if driver_converts_field_type_silently_to_str(collection.driver, schema_type):
                    # GDAL 1.x does not support milliseconds
                    if GDAL_VERSION_NUM < calc_gdal_version_num(2, 0, 0):
                        ms = 0
                    else:
                        # GDAL has only milliseconds precision
                        ms = int(ms / 1000.0) * 1000
                    if schema_type == 'date':
                        d = datetime.date(y, m, d)
                    elif schema_type == 'time':
                        d = datetime.time(hh, mm, ss, ms)
                    else:
                        d = datetime.datetime(y, m, d, hh, mm, ss, ms)

                    value_bytes = d.isoformat().encode(encoding)
                    string_c = value_bytes
                    OGR_F_SetFieldString(cogr_feature, i, string_c)
                else:
                    # Add microseconds to seconds
                    ss += ms / 10**6
                    set_field_datetime(cogr_feature, i, y, m, d, hh, mm, ss, 0)
```
Additionally, the field type must be overridden to string in WritingSession.

The full difference to the last commit can be seen here: https://github.com/rbuffat/Fiona/compare/add_support_for_milliseconds...rbuffat:add_support_for_milliseconds_convert_to_string

This would ensure that most of the supported drivers use the same format.

sgillies:
@rbuffat I'm a little concerned that changing the date/time formatting for the formats that silently convert to strings might break some assumptions of those drivers. Let's leave that aside for now.

rbuffat:
That's fine with me.

I would propose that once you are ok with the changes, I remove the old tests, as well as all tests for drivers converting the types to string. The old tests are covered by the new ones, and as we do not control the conversion, I think there is not much to gain with testing the special formatting of gdal drivers. It is sufficient to test if the field type is converted to string.

rbuffat:
Included in https://github.com/Toblerity/Fiona/pull/915

sgillies:
Thanks @rbuffat !

PR #894, opened 2020-05-06
Add gdal 3.1.0 to travis of maint-1.8, improve compile time

rbuffat:
As gdal 3.1.0rc3 is out I thought it would be good to enable it on travis. 

Furthermore, the PR backports the test fixes already applied to master and optimizes the compile time by disabling compiler optimizations (-O0). 

As currently the compile time of gdal, proj and fiona is >> the run time of the tests it decreases total run time of the travis build jobs. 
Compilation time of Fiona decreases from roughly 30 seconds to 15 seconds.
Compilation time of gdal master decreases from  20 minutes to 15 minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,31 +54,31 @@ matrix:
     # Test all supported python versions with latest stable gdal release
     - python: "2.7"
       env:
-        GDALVERSION="3.1.0rc3"
+        GDALVERSION="3.1.0"
         PROJVERSION="6.3.2"
     - python: "3.6"
       env:
-        GDALVERSION="3.1.0rc3"
+        GDALVERSION="3.1.0"
         PROJVERSION="6.3.2"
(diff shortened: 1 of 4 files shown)


rbuffat:
not really sure why the gdal 3.0.4 job is failing, here it was successful: https://travis-ci.org/github/rbuffat/Fiona/builds/684295388

.travis.yml
```diff
@@ -91,16 +91,16 @@ addons:
 
 before_install:
   - python -m pip install -U pip
+  - python -m pip wheel -r requirements-dev.txt
```
rbuffat, line 94: I put the pip install statements here as they do not depend on gdal / proj and are faster to execute. This avoids (in a worst case) to fail after 20 minutes when gdal is finally installed because a python dependency fails to install.

PR #899, opened 2020-05-10
Fix http paths

rbuffat:
GDAL changed the location of poly.zip and poly.shp. This PR adapts these changes. But long term it might be better to include a zip in the Fiona test directory.


sgillies:
@rbuffat I just made this one obsolete. I'll close it.

Issue #830, opened 2019-11-08
Write multiple layers to memory


aprilmay:
I need to create geopackages containing several layers, to be downloaded from a web server. Ideally in such a scenario, the data should be created in memory.

But Fiona has currently the following limitations:
- fiona.io.ZipMemoryFile supports multiple layers, but is for read only
- fiona.io.MemoryFile allows write, but switches to 'r' mode after writing the first layer.

So, my current implementation involves creating the file in the filesystem as a workaround, which is sub optimal.

Proposal: add a "mode" attribute in fiona.io.MemoryFile.open, which would be 'a' for this use case.

sgillies:
Completed!

Issue #754, opened 2019-06-01
Add methods to list and access multiple datasets within a zip file


sgillies:
See #753

snorfalorpagus:
It looks like `VSIReadDir` can be used to return a list of filenames in zip, or NULL if it doesn't exist. I'm not sure what the API should look like for this.

sgillies:
ZipMemoryFile was developed to serve the use case of a single zipped shapefile and its sidecar files, and multiple layers weren't a consideration. I believe we can extend it.

@snorfalorpagus for "ordinary" multilayer datasets we have `fiona.listlayers(path)` and  `fiona.open("path", layer="foo")`. These almost work for an in-memory zipfile, but you need to do some path hacking. Would this be a reasonable API?

mostfamiliar:
I'm waiting on listdir() or listlayers() as well ... I tried the suggestions in: https://github.com/Toblerity/Fiona/issues/753 and implemented it roughly the same except my file is an in memory upload:

uploaded = InMemoryUploaded File
```
                zf = zipfile.ZipFile(uploaded)
                names = zf.namelist()
                zf.close()
                regex = re.compile(r'.*\.shp')
                matching = self.match_filename(regex, names)

                with ZipMemoryFile(data["file"]) as zip_file:
                    with zip_file.open(matching[0]) as geo_file:
```
but zip_file.open results in: ERROR 4: `/vsizip/vsimem/cf972137877c410a9fd60388674726ae/cf972137877c410a9fd60388674726ae.zip/Trails_and_Paths.shp' does not exist in the file system, and is not recognized as a supported dataset name.

However:
strangely, if i comment out the ZipFile code i.e. `zipfile.ZipFile(uploaded)` and manually enter the .shp path into `zip_file.open('Trails_and_Paths.shp')` OR i use `zip_file.open()` it works:
<open Collection '/vsizip/vsimem/fac606d715af4501bd7f8a9871c3d23b/fac606d715af4501bd7f8a9871c3d23b.zip/Trails_and_Paths.shp:Trails_and_Paths', mode 'r' at 0x111535d90>

The problem is if file has a nested directory structure it will not open with `zip_file.open()` so I need to specify the path structure. Any ideas on why `zipfile.ZipFile` is causing a weird interaction? Is there something else I need to do besides closing it before I use `ZipMemoryFile`?

sgillies:
[...]

PR #901, opened 2020-05-20
Extend MemoryFile and ZipMemoryFile

rbuffat:
This PR extends the functionality of MemoryFile and ZipMemoryFile:
* Append support for MemoryFile
* Write support for ZipMemoryFile
* fiona.listdir() to list files in a directory
* ZipMemoryFile.listdir() and ZipMemoryFile.listlayers() to list files and layers in a ZipMemoryFile

Not all drivers seem to support MemoryFiles. The tests are extended to use all drivers available on the CI. A mechanism similar to drvsupport.driver_mode_mingdal is introduced to guard not supported drivers and modes. 

Issues:
* MemoryFileBase.exists() seems to have issues for some drivers when the memory files are created without the proper extension (e.g. the Shapefile driver)
* It would be preferable to use fiona.listdir() and fiona.listlayers() instead of ZipMemoryFile.listdir(), ZipMemoryFile.listlayers(), however, it is not that straightforward to create the path required as argument. 

Relevant issues: https://github.com/Toblerity/Fiona/issues/830 https://github.com/Toblerity/Fiona/issues/754

This PR is probably best suited for Fiona 1.9. I created it against 1.8, as this branch includes a more complete drvsupport list. Once 1.9 catches up it should be possible to change the PR to 1.9.


sgillies:
@rbuffat yesterday I finished a refactor of rasterio's MemoryFile after it was reported that its seek method was broken, preventing unordered multipart writes: https://github.com/mapbox/rasterio/pull/1935/files. I'd like to do the same for fiona's MemoryFile. After which, let's try to overlay this PR on top, and then I'll see able copying the directory support to rasterio :)

rbuffat:
@sgillies I will update this PR once you merge these changes.

I had a look at MemoryFileBase.exists().  It looks as CPLCheckForFile does more reliably detect the existence of a dataset when a name without extension is used. However, there are still a few issues:
- CPLCheckForFile returns True also for none dataset files. As long we control the filenames this should be not an issue. 
- GMT driver for GDAL 1.x requires an extension despite using CPLCheckForFile. With a .gmt extentions it works. (I did not yet test OGR_GMT)
- GPKG driver needs a file extension for GDAL 2.0 when appending. Otherwise GDAL uses the SQLite driver instead.  

A solution would be to set the appropriate extension in the initializationMemoryFile. However, this is a bit tricky because at this point it is unknown which driver will be used (or even written with write()).

rbuffat:
@sgillies I merged the new MemoryFile and tried to extend the test coverage using as many drivers as possible. This revealed some bugs and segfaults I need to look at (see TODOs in tests/test_memoryfile.py).

The most common issue is that drivers have problems opening filenames without the proper extension. The extensions of a driver can be accessed using GDALGetMetadataItem(cogr_driver, DMD_EXTENSION, NULL) if the driver is known. GDALGetMetadataItem should be available for > GDAL 2.0.

I would suggest refactoring MemoryFile to use the proper extension for the filenames when possible (= the driver is known).

sgillies:
Thank you! That's super useful. Fiona should certainly respect the GDAL/OGR driver capability API.

KeynesYouDigIt:
FWIW, ZipMemoryFile.listdir() and ZipMemoryFile.listlayers()  would be really helpful, I came here to raise an issue but it looks like its already in processes? these methods would be helpful for me :)

PR #902, opened 2020-05-20
Update drvsupport test (PCIDSK, GMT, PDS)

rbuffat:
I noticed that the test to append data only looks at the number of features is not enough to test the PCIDSK driver correctly. 

This PR modifies the drvsupport tests to compare the written data with the result. Ideally, the input records would be compared to the output records, but this is tricky, as some drivers add fields (e.g. the Shapefile driver adds a FID field). This PR only looks at one property (if the field type is supported), respectively checks if the coordinate is correct otherwise. 

Additionally, this PR enables the OGR_PDS and OGR_GMT drivers. It looks as they were renamed with GDAL 2.x. Thus I suppose they are not often used in combination with Fiona. 

```
from collections import OrderedDict
import pprint
import fiona
import os

schema = {'geometry': 'Point', 'properties': OrderedDict([('position', 'int')])}
records1 = [{'geometry': {'type': 'Point', 'coordinates': (0.0, float(i))}, 'properties': {'position': i}} for i in range(0, 2)]
records2 = [{'geometry': {'type': 'Point', 'coordinates': (0.0, float(i))}, 'properties': {'position': i}} for i in range(5, 10)]
filename = '/tmp/test.pix'
driver = "PCIDSK"

if os.path.exists(filename):
    os.unlink(filename)

with fiona.open(filename,
                mode='w',
                driver=driver,
                schema=schema) as out:
    out.writerecords(records1)

with fiona.open(filename,
                mode='a',
                driver=driver,
                schema=schema) as out:
    out.writerecords(records2)

with fiona.open(filename) as c:
    for f in c:
        pprint.pprint(f)
```

Output:
```
{'geometry': {'coordinates': (0.0, 5.0, 0.0), 'type': 'Point'},
 'id': '0',
 'properties': OrderedDict([('position', '5')]),
 'type': 'Feature'}
{'geometry': {'coordinates': (0.0, 6.0, 0.0), 'type': 'Point'},
 'id': '1',
 'properties': OrderedDict([('position', '6')]),
 'type': 'Feature'}
{'geometry': {'coordinates': (0.0, 7.0, 0.0), 'type': 'Point'},
 'id': '0',
 'properties': OrderedDict([('position', '7')]),
 'type': 'Feature'}
{'geometry': {'coordinates': (0.0, 8.0, 0.0), 'type': 'Point'},
 'id': '1',
 'properties': OrderedDict([('position', '8')]),
 'type': 'Feature'}
{'geometry': {'coordinates': (0.0, 9.0, 0.0), 'type': 'Point'},
 'id': '2',
 'properties': OrderedDict([('position', '9')]),
 'type': 'Feature'}
{'geometry': None,
 'id': '3',
 'properties': OrderedDict([('position', '')]),
 'type': 'Feature'}
{'geometry': None,
 'id': '4',
 'properties': OrderedDict([('position', '')]),
 'type': 'Feature'}
```

### DIFF
--- a/fiona/drvsupport.py
+++ b/fiona/drvsupport.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from fiona._env import get_gdal_version_num, calc_gdal_version_num
 
 from fiona.env import Env
 
@@ -159,6 +160,19 @@ driver_mode_mingdal = {
           'GeoJSON': (2, 1, 0),
           'MapInfo File': (2, 0, 0)}
 }
+
+
+def driver_supports_mode(driver, mode):
+    """ Returns True if driver supports mode, False otherwise"""
+
+    if driver not in supported_drivers:
+        return False
+    if mode not in supported_drivers[driver]:
+        return False
+    if driver in driver_mode_mingdal[mode]:
+        if get_gdal_version_num() < calc_gdal_version_num(*driver_mode_mingdal[mode][driver]):
+            return False
+    return True
 
 
 # Removes drivers in the supported_drivers dictionary that the
--- a/fiona/drvsupport.py
+++ b/fiona/drvsupport.py
@@ -63,7 +63,8 @@ supported_drivers = dict([
     # GML 	GML 	Yes 	Yes 	Yes (read support needs Xerces or libexpat)
     ("GML", "rw"),
     # GMT 	GMT 	Yes 	Yes 	Yes
-    ("GMT", "raw"),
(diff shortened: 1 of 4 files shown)


rbuffat:
Actually, to compare records would be handy, and the same code could also be used in the tests of https://github.com/Toblerity/Fiona/pull/901

I will try to find a suitable implementation.

rbuffat:
Included in https://github.com/Toblerity/Fiona/pull/915

PR #904, opened 2020-05-26
Select Python version on OSX Travis Jobs using pyenv

rbuffat:
Using pyenv it is possible to select a Python Version on OSX. The drawback is, that the installation requires roughly 3 additional minutes, however, it can be cached. I thought about using the same build matrix for OSX as for Linux, but I did not because the build time of OSX jobs is a lot slower and it seems as there are only two concurrent jobs active. Thus, it might be good if there are fewer OSX jobs running. 

The Python version must be one listed in "pyenv install --list". I noticed that different versions of Xcode support different Python versions. Thus, if a newer Python version needs to be supported, a newer Xcode image might also be needed. I choose Xcode9.4, as it was faster compared to newer images (in a single run, so this might be arbitrary).

Furthermore, this PR syncs travis with the recent changes in maint-1.8 (rstcheck currently disabled as tests would fail) and uses SVG for the travis / coveralls / appveyor badges (I'm not sure why it irritates me that they currently do not have the same height...).

Additionally, a Python 3.9-dev job is added, which is marked as allowed to fail. A test fails due to botocore, which is currently not compatible with Python 3.9, as it imports xml.etree.cElementTree, which was removed in Python 3.9.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[run]
+plugins = Cython.Coverage
+omit = *pxd
--- a/.github/workflows/ci_macos.yml
+++ b/.github/workflows/ci_macos.yml
@@ -1,0 +1,92 @@
+
+name: MacOS CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: macos-10.15
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - {python-version: 3.6, GDALVERSION: "2.3.3",   PROJVERSION: "4.9.3"}
+          - {python-version: 3.6, GDALVERSION: "2.4.4",   PROJVERSION: "4.9.3"}
+          - {python-version: 3.6, GDALVERSION: "3.0.4",   PROJVERSION: "6.2.1"}
+          - {python-version: 3.6, GDALVERSION: "3.1.0",   PROJVERSION: "6.3.2"}
+          - {python-version: 3.7, GDALVERSION: "3.1.0",   PROJVERSION: "6.3.2"}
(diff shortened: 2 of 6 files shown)


rbuffat:
I added a configuration for Github CI for OSX builds. It was a bit more complicated than I expected. There are two big differences between travis and github CI. With Github, each step is run in a different process. Sharing environment variables between steps requires thus a workaround. Furthermore, the caching mechanism is slightly different. 
While working on this I notices a bug I introduced with the OSX builds on travis. The issue is, that as far I see it is not possible to pin a specific version of homebrew packages. Thus, always the most recent packages are installed. This will lead eventually to problems if the homebrew packages are incompatible with cached gdal / proj binaries (shared libraries such as sqlite, goes, spatiallite, ...).
With the Github CI a key is defined, with them the cache is identified:

```
brew list --versions | grep 'expat\|libspatialite\|sqlite\|geos\|curl\|libxml2\|freexl\|zstd' > $GITHUB_WORKSPACE/brew_list
```
```
      - name: Dependency Cache
        uses: actions/cache@v2
        with:
          path: gdalinstall
          key: ${{ runner.os }}-gdal-${{ matrix.config.GDALVERSION }}-proj-${{ matrix.config.PROJVERSION }}-${{ hashFiles('**/brew_list')}}
```

Adding `${{ hashFiles('**/brew_list')}}` to the cache key should invalidate the cache when a version of a homebrew package changed. I tried to filter packages that are used by gdal, I hope I caught all relevant ones. 

I'm not yet sure what has to be done to activate the github ci here. In my repository, it seemed to be enough to add the yml file in .github/workflows.

rbuffat:
Travis is obsolete now.

Issue #771, opened 2019-07-16
Overwriting GeoJSON logs an ERROR but succeeds


kannes:
## Expected behavior and actual behavior.

Overwriting a GeoJSON file results in a `logging` error message about the file not existing.

## Steps to reproduce the problem.

Same example code as in https://github.com/Toblerity/Fiona/issues/438#issuecomment-343154689 but with logging enabled.

```
import logging

logging.basicConfig(
    level=logging.INFO,
)

import fiona
from fiona.crs import from_epsg
from shapely.geometry import Point, mapping

schema = {
    "geometry": "Point",
    "properties": {"id": "int"}
}

feature = {
    "geometry": mapping(Point(0,0)),
    "properties": {"id": "0"}
}

for i in range(2):
    with fiona.open("/tmp/test.geojson", "w", 
                    driver="GeoJSON",
                    schema=schema, 
                    crs=from_epsg(4326)) as sink:
        sink.write(feature)
        print(f"Write #{i+1} succeeded")
```

results in

```
Write #1 succeeded
ERROR:fiona._env:/tmp/test.geojson: No such file or directory
WARNING:fiona._env:driver GeoJSON does not support creation option ENCODING
Write #2 succeeded
```

The error message is false. The file exists after the first successful write. At least from my user perspective.

## Operating system

Archlinux

## Fiona and GDAL version and provenance

gdal 3.0.0-2 from Archlinux repositories.
Fiona 1.8.6 via pip.

sgillies:
Thanks for the report, @kannes. When the `with` block ends, the file is definitely flushed to disk and will exist. I'll look into correcting the error message.

Issue #568, opened 2018-04-05
Problems with format guessing when writing Shapefile/GeoJSON


kannes:
I am sorry in advance, because this is dumb, but it is a bug.

If I have a shapefile of name `foo.shp` (plus auxiliary files) and try to create a GeoJSON file named `foo.shp` (because I forgot to update the filename), I will get a Shapefile named `OgrGeoJSON.shp` (plus aux. files) instead. Here is a test script:

```
import fiona
from fiona.crs import from_epsg
import glob

outdir = "/tmp"
file = "O"  # "O" so that we will see the OgrGeoJSON files in globbing

p = {
    'geometry': {'coordinates': (370262, 5615696), 'type': 'Point'},
    'id': '0', 'properties': {}, 'type': 'Feature'
}

schema = {'geometry': 'Point', 'properties': {}}

with fiona.open(f"{outdir}/{file}.shp", "w", crs=from_epsg(25832), schema=schema, driver="ESRI Shapefile") as sink:
    sink.write(p)
        
print(glob.glob(f"{outdir}/{file}*"))
# -> ['/tmp/O.prj', '/tmp/O.cpg', '/tmp/O.dbf', '/tmp/O.shx', '/tmp/O.shp']

# here we have .shp but driver="GeoJSON"
with fiona.open(f"{outdir}/{file}.shp", "w", crs=from_epsg(25832), schema=schema, driver="GeoJSON") as sink:
    sink.write(p)
        
print(glob.glob(f"{outdir}/{file}*"))
# -> ['/tmp/OgrGeoJSON.prj', '/tmp/OgrGeoJSON.cpg', '/tmp/OgrGeoJSON.dbf', '/tmp/OgrGeoJSON.shx', '/tmp/OgrGeoJSON.shp', '/tmp/O.prj', '/tmp/O.cpg', '/tmp/O.dbf', '/tmp/O.shx', '/tmp/O.shp']
```

I would have expected either an error or the foo.shp file being overwritten (aux. files left in place for extra confusion because screw the Shapefile format clutter).

The other way around is not much better as Fiona now tries to overwrite the existing foo.geojson with another GeoJSON even though ESRI Shapefile was set as the driver.

```
import fiona
from fiona.crs import from_epsg
import glob

outdir = "/tmp"
file = "O"  # "O" so that we will see the OgrGeoJSON files in globbing

p = {
    'geometry': {'coordinates': (370262, 5615696), 'type': 'Point'},
[...]

Issue #504, opened 2017-11-16
Avoid passing encoding creation option to format drivers that don't support it


sgillies:
See https://github.com/Toblerity/Fiona/issues/468#issuecomment-344111416: the GeoJSON driver warns that ENCODING isn't supported.

@mattayes the warning is harmless but we can avoid it.

snorfalorpagus:
I can't see any way to ask GDAL which creation options are supported by a driver/dataset/layer. Do we need to build our own list for the drivers we support?

sgillies:
According to http://www.gdal.org/gdal_8h.html#aec318ffe797555656fe678558c576b23 `GDALValidateCreationOptions` will validate options, and there's a hint about a `GDAL_DMD_CREATIONOPTIONLIST` driver metadata item.

Still, I think the immediate thing for me to do is to eliminate the encoding option in the GeoJSON case.

snorfalorpagus:
```
Warning 6: driver MapInfo File does not support creation option ENCODING
```

Looks like MapInfo is the same.

kurt-rhee:
KML driver as well

`not recognized as a supported file format.
driver KML does not support creation option ENCODING`

sgillies:
Done!

PR #909, opened 2020-06-05
Expose driver metadata, refactor dataset creation

rbuffat:
Since GDAL 2.0, OGR drivers include more metadata. Currently, there is no way to access this metadata (e.g. to query supported dataset open options, dataset creation options, layer creation options, file extensions etc.) through Fiona. This PR adds a new meta module that allows to query metadata. I decided to create a new module as ogrext seems to grow quite large. If you think that this functionality should belong in Fiona we should think about the best architecture (which might be sharable with rasterio). 

File extensions could be used to create in memory datasets, as some drivers have troubles with files without the proper extension (https://github.com/Toblerity/Fiona/pull/901).   With the current implementation,  `fiona.meta.print_driver_options('GeoJSON')`
outputs the following information:  
```
Dataset Open Options:
    FLATTEN_NESTED_ATTRIBUTES:
        Description: Whether to recursively explore nested objects and produce flatten OGR attributes
        Type: boolean
        Default value: NO
    NESTED_ATTRIBUTE_SEPARATOR:
        Description: Separator between components of nested attributes
        Type: string
        Default value: _
    FEATURE_SERVER_PAGING:
        Description: Whether to automatically scroll through results with a ArcGIS Feature Service endpoint
        Type: boolean
    NATIVE_DATA:
        Description: Whether to store the native JSon representation at FeatureCollection and Feature level
        Type: boolean
        Default value: NO
    ARRAY_AS_STRING:
        Description: Whether to expose JSon arrays of strings, integers or reals as a OGR String
        Type: boolean
        Default value: NO
    DATE_AS_STRING:
        Description: Whether to expose date/time/date-time content using dedicated OGR date/time/date-time types or as a OGR String
        Type: boolean
        Default value: NO

Dataset Creation Options:
    No options available.

Layer Creation Options:
    WRITE_BBOX:
        Description: whether to write a bbox property with the bounding box of the geometries at the feature and feature collection level
        Type: boolean
        Default value: NO
    COORDINATE_PRECISION:
        Description: Number of decimal for coordinates. Default is 15 for GJ2008 and 7 for RFC7946
        Type: int
    SIGNIFICANT_FIGURES:
        Description: Number of significant figures for floating-point values
        Type: int
        Default value: 17
    NATIVE_DATA:
        Description: FeatureCollection level elements.
        Type: string
    NATIVE_MEDIA_TYPE:
        Description: Format of NATIVE_DATA. Must be "application/vnd.geo+json", otherwise NATIVE_DATA will be ignored.
        Type: string
    RFC7946:
        Description: Whether to use RFC 7946 standard. Otherwise GeoJSON 2008 initial version will be used
        Type: boolean
        Default value: NO
    WRITE_NAME:
        Description: Whether to write a "name" property at feature collection level with layer name
        Type: boolean
        Default value: YES
    DESCRIPTION:
        Description: (Long) description to write in a "description" property at feature collection level
        Type: string
    ID_FIELD:
        Description: Name of the source field that must be used as the id member of Feature features
        Type: string
    ID_TYPE:
        Description: Type of the id member of Feature features
        Type: string-select
        Accepted values: AUTO,String,Integer
    WRITE_NON_FINITE_VALUES:
        Description: Whether to write NaN / Infinity values
        Type: boolean
        Default value: NO 
```

Knowing the available options allows to filter them in ogrext WritingSession before being passed to gdal_open_vector, gdal_create, respectively GDALDatasetCreateLayer to avoid warnings. (Currently only for write mode, as only there a driver is known). Adding filtering for read and append mode could be added by first querying the driver, such as in _remove() using gdal_open_vector and GDALGetDatasetDriver.
This is related to  https://github.com/Toblerity/Fiona/issues/504 . However, I'm unsure how much of this issue is already solved with https://github.com/Toblerity/Fiona/commit/5e687a06e55d1c634da7792b950deb4b6ea169b3

Adding options filtering to these functions meant to refactor the creation of dataset in the WritingSession.  Especially this part needs a careful review. The biggest change involves using _remove() instead of os.unlink() for all drivers instead of just GeoJSON.  This allows in the following example that all auxiliary files of a Shapefile are deleted when it is overwritten by a GeoJSON dataset. However, as deleting files is potentially very dangerous, I'm unsure if the default of Fiona should be to overwrite files or to fail. What about adding an overwrite=True/False argument to Collection?  The refactoring also fixes https://github.com/Toblerity/Fiona/issues/771 and https://github.com/Toblerity/Fiona/issues/568

```
with fiona.open('foo.shp', "w", driver="ESRI Shapefile", schema=schema1) as dst:
    dst.writerecords(features1)

assert set(os.listdir(tmpdir.strpath)) == {'foo.cpg', 'foo.dbf', 'foo.shx', 'foo.shp'}


with fiona.open('foo.shx', "w", driver="GeoJSON", schema=schema1) as dst:
    dst.writerecords(features1)

assert os.listdir(tmpdir.strpath) == ['foo.shx']
```

If you think this PR is too big or you would like only a subset of the features, I'm happy to refactor this PR or create a new one. 
Documentation of the new features is currently missing, as it is probably best to wait until the implementation is stable before writing it.

### DIFF
--- a/fiona/_meta.pyx
+++ b/fiona/_meta.pyx
@@ -34,10 +34,15 @@ def _get_metadata_item(driver, metadata_item):
     cdef char* metadata_c = NULL
     cdef void *cogr_driver
 
-    metadata = ""
     cogr_driver = exc_wrap_pointer(GDALGetDriverByName(driver.encode('utf-8')))
     metadata_c = GDALGetMetadataItem(cogr_driver, strencode(metadata_item), NULL)
+
+    metadata = None
     if metadata_c != NULL:
         metadata = metadata_c
         metadata = metadata.decode('utf-8')
+
+        if len(metadata) == 0:
+            metadata = None
+
     return metadata
--- a/fiona/_meta.pyx
+++ b/fiona/_meta.pyx
@@ -1,0 +1,43 @@
+
+include "gdal.pxi"
+
+from fiona._err cimport  exc_wrap_pointer
+from fiona.compat import strencode
+from fiona._shim cimport gdal_open_vector
+from fiona.env import require_gdal_version
+import logging
+
+# This is required otherwise GDALGetDriverByName returns NULL for GDAL 1.x
+from fiona._shim cimport *
+
+
+log = logging.getLogger(__name__)
+
+
+@require_gdal_version('2.0')
+def _get_metadata_item(driver, metadata_item):
+    """Query metadata items
+
+    Parameters
+    ----------
+    driver : str
+        Driver to query
+    metadata_item : str
+        Metadata item to query
+
+    Returns
+    -------
+    str
+        XML of metadata item or empty string
+
+    """
+    cdef char* metadata_c = NULL
+    cdef void *cogr_driver
+
+    metadata = ""
+    cogr_driver = exc_wrap_pointer(GDALGetDriverByName(driver.encode('utf-8')))
+    metadata_c = GDALGetMetadataItem(cogr_driver, strencode(metadata_item), NULL)
+    if metadata_c != NULL:
+        metadata = metadata_c
+        metadata = metadata.decode('utf-8')
+    return metadata
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -994,24 +994,33 @@ cdef class WritingSession(Session):
                 driver_layer_creation_options = layer_creation_options(collection.driver)
 
                 # filter options not supported by driver
-                open_kwargs = {}
-                for k, v in kwargs.items():
-                    if k.upper() in driver_dataset_open_options:
-                        open_kwargs[k] = v
-                    else:
-                        log.debug("Ignore '{}' as dataset open option.".format(k))
-                create_kwargs = {}
-                for k, v in kwargs.items():
-                    if k.upper() in driver_dataset_creation_options:
-                        create_kwargs[k] = v
-                    else:
-                        log.debug("Ignore '{}' as dataset creation option.".format(k))
-                layer_kwargs = {}
-                for k, v in kwargs.items():
-                    if k.upper() in driver_layer_creation_options:
-                        layer_kwargs[k] = v
-                    else:
-                        log.debug("Ignore '{}' as layer creation option.".format(k))
+                def filter_options(kwargs, driver_options):
+                    # If driver does not specify options we cannot filter
+                    if driver_options is None:
+                        return kwargs, []
+
+                    filtered_options = {}
+
+                    invalid_options = []
+                    for k, v in kwargs.items():
+                        if k.upper() in driver_options:
+                            filtered_options[k] = v
+                        else:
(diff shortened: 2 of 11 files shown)


fiona/_meta.pyx
```diff
@@ -0,0 +1,43 @@
...
+include "gdal.pxi"
+
+from fiona._err cimport  exc_wrap_pointer
+from fiona.compat import strencode
+from fiona._shim cimport gdal_open_vector
+from fiona.env import require_gdal_version
+import logging
+
+# This is required otherwise GDALGetDriverByName returns NULL for GDAL 1.x
+from fiona._shim cimport *
```
rbuffat, line 11: I don't know why GDALGetDriverByName returns NULL here but not in ogrext without this import. As metadata is mostly only available for gdal >= 2.0, this is not that important, but I find it still strange.

fiona/ogrext.pyx
```diff
@@ -984,31 +984,57 @@ cdef class WritingSession(Session):
...
+                driver_dataset_creation_options = dataset_creation_options(collection.driver)
+                driver_layer_creation_options = layer_creation_options(collection.driver)
+
+                # filter options not supported by driver
+                open_kwargs = {}
+                for k, v in kwargs.items():
+                    if k.upper() in driver_dataset_open_options:
+                        open_kwargs[k] = v
+                    else:
+                        log.debug("Ignore '{}' as dataset open option.".format(k))
```
rbuffat, line 1002: Would a log.info be more appropriate instead of log.debug?

fiona/ogrext.pyx
```diff
@@ -1575,7 +1609,7 @@ def _remove(path, driver=None):
         cogr_driver = GDALGetDatasetDriver(cogr_ds)
         GDALClose(cogr_ds)
     else:
-        cogr_driver = OGRGetDriverByName(driver.encode("utf-8"))
+        cogr_driver = exc_wrap_pointer(GDALGetDriverByName(driver.encode("utf-8")))
```
rbuffat, line 1612: OGRGetDriverByName is deprecated

fiona/_meta.pyx
```diff
@@ -0,0 +1,43 @@
...
+    str
+        XML of metadata item or empty string
+
+    """
+    cdef char* metadata_c = NULL
+    cdef void *cogr_driver
+
+    metadata = ""
+    cogr_driver = exc_wrap_pointer(GDALGetDriverByName(driver.encode('utf-8')))
+    metadata_c = GDALGetMetadataItem(cogr_driver, strencode(metadata_item), NULL)
```
rbuffat, line 39: Currently, only the default domain is used (NULL). The domains can be queried with GDALGetMetadataDomainList, however for the driver I tested it with only the default domain (NULL) was reported. I have to admit I don't yet fully understand the domain concept and where it is used.

rbuffat:
Closed in favor of https://github.com/Toblerity/Fiona/pull/950

Issue #905, opened 2020-05-30
1.8.13.post1 causes mask filter to act like a bounding box filter


nikmolnar:
## Expected behavior and actual behavior.

I have a script that summarizes features in dataset A that intersect with features in dataset B. This code relies on `.items(mask=geometry_B)` to select the features of dataset A that exist within each feature of dataset B.

This code works fine in Fiona 1.8.13. But in 1.8.13.post1, it seems to filter by bounding box, rather than the actual feature geometry.

## Steps to reproduce the problem.

This is a simplified version of the code in question. I am unable to share the data, but I could come up with other data to reproduce the problem, if it becomes necessary.

```python
with fiona.open('dataset_b.gpkg') as dataset_b:
    for feature in dataset_b:
        with fiona.open('dataset_a.gpkg') as dataset_a:
            items = dataset_a.items(mask=feature["geometry"])
```

## Operating system

macOS 10.15.3

## Fiona and GDAL version and provenance

Fiona: 1.8.13.post1
GDAL: 2.4.2 (built from source)

sgillies:
@nikmolnar I think there must be something else going on. The fiona code in that area hasn't changed between those versions: https://github.com/Toblerity/Fiona/blame/4750ae99119c7bd2bc38d7842fd57c1c2db2cd2e/fiona/ogrext.pyx#L1233-L1236.

Now, the GDAL/OGR function that fiona calls, `OGR_L_SetSpatialFilter`, is format dependent. Is it possible that your application changed formats and that the new one doesn't support geometry filtering? GeoPackage files might have special indexing needs. I'm not very familiar with the format, I confess.

It's also noted in https://gdal.org/api/vector_c_api.html#_CPPv422OGR_L_SetSpatialFilter9OGRLayerH12OGRGeometryH that the test isn't 100% accurate and that there may be some false positives. It doesn't say that this is format dependent, but that is my take.

nikmolnar:
@sgillies l suspected format at first, too, since I had previously used shapefiles. However, with no code or data changes, I've observed this working correctly at version 1.8.13 and incorrectly at version 1.8.13.post1.

I will see if the same happens with shapefiles, and try to put together a minimally complete example to reproduce.

nikmolnar:
Ok, here is a minimum reproduction of the problem, using shapefiles.

```python
import fiona


def main():
    with fiona.open('california/california.shp') as california_shp:
        feature = next(california_shp)

[...]

PR #912, opened 2020-06-11
Improve mask documentation

rbuffat:
Improve documentation for the inaccuracy of spatial filtering.
Add test that fails with wheel. 
(See https://github.com/Toblerity/Fiona/issues/905)

### DIFF
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -1,8 +1,11 @@
 # Testing collections and workspaces
 
 import datetime
+import os
+import random
 import sys
 import re
+from collections import OrderedDict
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


tests/test_collection.py
```diff
@@ -911,3 +914,28 @@ def test_collection_env(path_coutwildrnp_shp):
...
+
+    with fiona.open(path, 'w',
+                    driver=driver,
+                    schema=schema,) as c:
+        c.writerecords(records)
+
+    with fiona.open(path) as c:
+        items = list(
+            c.items(mask={'type': 'Polygon', 'coordinates': (((2.0, 2.0), (4.0, 4.0), (4.0, 6.0), (2.0, 2.0)),)}))
+        assert len(items) == 15
```
QuLogic, line 941: This test fails on s390x because it returns 24 items. I don't know if that's a bug in GDAL or not, but given the caveats added to the documentation, shouldn't this be a `>=`?

rbuffat: This test was introduced to test if the GDAL build included in Fiona's wheels are correctly setup to use GEOS for spatial filtering. When this is not setup correctly, if I correctly remember, GDAL uses a bounding box of the mask as filter criteria.

QuLogic: Yes, looking through related issues, it does appear that this is due to GDAL not being compiled against GEOS. And it turns out that GDAL was not compiled with GEOS on s390x, but was everywhere else, because their autotools checks are a bit buggy.

Issue #914, opened 2020-06-16
Unwanted removal of ISO8601 datetime string timezone information when reading from GeoJSON


EigenJT:
Bringing this over from `geopandas`. It seems like reading an ISO8601 datetime string from a `GeoJSON` erases the timezone information at some point.
[https://github.com/geopandas/geopandas/issues/1472](https://github.com/geopandas/geopandas/issues/1472)

## Expected behavior and actual behavior.
Expected to read in timezone aware ISO8601 datetime strings in a `GeoJSON` as strings. Got string that has been stripped of timezone information.

I.E. in the `GeoJSON`, we have `2010-10-20T12:00:00+07:00` but when read in using `Fiona`, got `2010-10-20T12:00:00`.

## Operating system
Mac OS X 10.15.5.

## Fiona and GDAL version and provenance
Fiona: 1.8.13.post1 installed via Pypi
GDAL: 2.4.4 installed via Homebrew

rbuffat:
Currently, Fiona supports RFC3339 and is not aware of timezones (= Ignoring potential timezone information from GDAL). 

Read support could probably be implemented quite easily, writing datetimes with timezones could potential be more tricky. 

GDAL' s  documentation regarding timezone information is as follows:
```
OGR_F_GetFieldAsDateTimeEx: pnTZFlag: (0=unknown, 1=localtime, 100=GMT, see data model for details)
CPLParseRFC822DateTime: pnTZFlag:  (0=unknown, 100=GMT, 101=GMT+15minute, 99=GMT-15minute), or NULL
```

EigenJT:
> Currently, Fiona supports RFC3339 and is not aware of timezones (= Ignoring potential timezone information from GDAL).
> 
> Read support could probably be implemented quite easily, writing datetimes with timezones could potential be more tricky.
> 
> GDAL' s documentation regarding timezone information is as follows:
> 
> ```
> OGR_F_GetFieldAsDateTimeEx: pnTZFlag: (0=unknown, 1=localtime, 100=GMT, see data model for details)
> CPLParseRFC822DateTime: pnTZFlag:  (0=unknown, 100=GMT, 101=GMT+15minute, 99=GMT-15minute), or NULL
> ```

Thanks for the response @rbuffat . I'm not too familiar with configuring GDAL. Would there be a way of disabling the parsing and just treat datetimes as regular strings

rbuffat:
For the GeoJSON driver, you can use the DATE_AS_STRING option:
```
[...]

PR #915, opened 2020-06-22
Add support for time zones

rbuffat:
This PR introduces support for time zones.

The PR includes https://github.com/Toblerity/Fiona/pull/893 and https://github.com/Toblerity/Fiona/pull/902. They are related and help here. I will update the PR once these PRs are merged or closed.

Timezones in GDAL can be set using the TZFlag of OGR_F_SetFieldDateTime[Ex]. Valid values are  0 for an unknown timezone and 100=GMT, 101=GMT+15minute, 99=GMT-15minute, .. for known timezones. 

The current behavior of Fiona is to always use unknown timezones (TZFlag = 0). If a field with a timezone is read, the timezone information is discarded. Same for writing data. This leads to issues such as in https://github.com/Toblerity/Fiona/issues/914 

Support for timezones varies by driver, and unfortunately, no metadata is available for it.
Issues are:
- Some drivers do not support timezones
- Some drivers do only support GMT timezone
- Some drivers do not support unknown timezones
 
This PR adds a data structure to identify these drivers in drvsupport and the functions  _driver_supports_timezones(driver, field_type),  _driver_supports_unknown_timezones(driver, field_type).

When reading data, the TZFlag of OGR_F_GetFieldAsDateTimeEx returns a TZFlag is now taken into account. E.g. for the case of https://github.com/Toblerity/Fiona/issues/914, "2015-04-22T00:00:00+07:00" is now returned instead of "2015-04-22T00:00:00".

When writing data, it is checked if the driver supports the timezone. If yes, the datetime / time data is saved with the timezones. 
If a driver does not support time zones (or only UTC), and the data to be written has a timezone other than UTC, the data is converted to UTC.

The GPKG driver, and the GPX driver until GDAL 2.4 do not support unknown timezones. These drivers interpret the unknown flag as GMT (Writing a TZFlag of 0 results in reading a TZFlag of 100). Thus, while with the old implementation, where the timezone is discarded, when a datetime value was returned as "2015-04-22T00:00:00" it will now be returned as "2015-04-22T00:00:00+00:00".

@jorisvandenbossche if you have time, could you look at the changes in ogrext.pyx and especially test_datetime.py


rbuffat:
I found an issue with this implementation for the 'time' field type: the utcoffset of atimezone is not fixed.  E.g. if the timezone has daylight saving times. The timezone Europe/Zurich is either UTC+1 or UTC+2. For datetime, this is no problem, as the date is known.

Some drivers support timezones for times (e.g. PCDISK, although this driver converts the time to string). I would suggest raising an Exception if a driver does not support timezones and a feature contains a time with a timezone different than UTC.

fiona/drvsupport.py
```diff
@@ -151,14 +155,25 @@
...
+    """ Returns True if driver supports mode, False otherwise"""
+
+    if driver not in supported_drivers:
+        return False
+    if mode not in supported_drivers[driver]:
+        return False
+    if driver in driver_mode_mingdal[mode]:
+        if get_gdal_version_num() < calc_gdal_version_num(*driver_mode_mingdal[mode][driver]):
+            return False
+    return True
```
sgillies, line 174: @rbuffat would you be okay with making this a "private" function in 1.8.14, i.e. changing the name to `_driver_supports_mode`?

tests/data/test_tz.geojson
```diff
@@ -0,0 +1,18 @@
+{
+  "type": "FeatureCollection",
+  "features": [
+    {
+      "type": "Feature",
+      "properties": {
+        "datetime": "2015-04-22T00:00:00+07:00"
```
sgillies, line 7: Here, I'd like to see the property changed from "datetime" to something like "timestamp" to make sure that the implementation doesn't depend on a special name.

sgillies:
@rbuffat thank you! Can't wait to see this fixed in 1.8.14.

Issue #916, opened 2020-06-24
Field names mixed up writing via "KML" driver


wheeled:
## Expected behavior and actual behavior.

Attempting to save a GeoPandas dataframe using:
```
        with fiona.Env():
            gdf.to_file(kml_path, driver='KML', NameField='name')
```
produces a KML file in which the value for `<name>` is taken from another field, and likewise `<description>`.  If the file is written the hard way (using `ogr`) it is correct.  Perhaps there are some kwargs that need to be passed along to correct this?  In the example above `NameField='name'` was an unsuccessful attempt.

## Correct KML using `ogr`:

```
<?xml version="1.0" encoding="utf-8" ?>
<kml xmlns="http://www.opengis.net/kml/2.2">
<Document id="root_doc">
<Schema name="walk" id="walk">
	<SimpleField name="id" type="string"></SimpleField>
	<SimpleField name="date" type="string"></SimpleField>
	<SimpleField name="distance" type="string"></SimpleField>
	<SimpleField name="speed" type="string"></SimpleField>
	<SimpleField name="time" type="string"></SimpleField>
	<SimpleField name="activity_name" type="string"></SimpleField>
</Schema>
<Folder><name>walk</name>
  <Placemark>
	<name>Walked 3.97 km on 22/1/17</name>
	<Style><LineStyle><color>ff0000ff</color></LineStyle><PolyStyle><fill>0</fill></PolyStyle></Style>
	<ExtendedData><SchemaData schemaUrl="#walk">
		<SimpleData name="id">1965703340</SimpleData>
		<SimpleData name="date">01/22/2017</SimpleData>
		<SimpleData name="distance">3.96978493824</SimpleData>
		<SimpleData name="speed">1.397811597971831</SimpleData>
		<SimpleData name="time">2840</SimpleData>
		<SimpleData name="activity_name">Walk</SimpleData>
	</SchemaData></ExtendedData>
      <LineString><coordinates>151.165393442,-33.7688339688 ... 151.16551338,-33.7687453898</coordinates></LineString>
  </Placemark> ...
```

## Incorrect KML using `Fiona` (see Placemark name and description):

```
<?xml version="1.0" encoding="utf-8" ?>
<kml xmlns="http://www.opengis.net/kml/2.2">
<Document id="root_doc">
<Schema name="walk" id="walk">
	<SimpleField name="id" type="string"></SimpleField>
	<SimpleField name="date" type="string"></SimpleField>
[...]

PR #918, opened 2020-07-05
Fix schema mapping if driver includes default fields

rbuffat:
Problem:
Some drivers create fields by default. Fiona is currently not aware of these fields. Additionally, some drivers modify the field names (e.g. Shapefile has a length restriction for field names). Fiona maintains a schema mapping between field names in the schema provided to Fiona, and the resulting field names. However, as the default fields are not considered, the resulting mapping is wrong (see https://github.com/Toblerity/Fiona/issues/916).

This PR fixes the schema mapping if a driver contains default fields. This is kind of a minimal solution, as the existence of default fields also influence the schema validation, which is not touched with this PR. This PR does not modify that only the fields in the schema provided to Fiona are validated.

Currently, schema validation requires that records include all fields of the schema. However, drivers include also "optional" default fields, that do not need to be present. E.g. the default fields for the GPX driver are:
```
        'GPX': {'properties': OrderedDict([('ele', 'float'),
                                           ('time', 'datetime'),
                                           ('magvar', 'float'),
                                           ('geoidheight', 'float'),
                                           ('name', 'str'),
                                           ('cmt', 'str'),
                                           ('desc', 'str'),
                                           ('src', 'str'),
                                           ('link1_href', 'str'),
                                           ('link1_text', 'str'),
                                           ('link1_type', 'str'),
                                           ('link2_href', 'str'),
                                           ('link2_text', 'str'),
                                           ('link2_type', 'str'),
                                           ('sym', 'str'),
                                           ('type', 'str'),
                                           ('fix', 'str'),
                                           ('sat', 'int'),
                                           ('hdop', 'float'),
                                           ('vdop', 'float'),
                                           ('pdop', 'float'),
                                           ('ageofdgpsdata', 'float'),
                                           ('dgpsid', 'int')]),
```

E.g. the`geoidheight` element is defined with `minOccurs="0"` in https://www.topografix.com/GPX/1/1/gpx.xsd

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -921,6 +921,8 @@ cdef class WritingSession(Session):
 
             else:
                 self._fileencoding = userencoding or self._get_fallback_encoding()
+                
+            before_fields = self.get_schema()['properties']
 
         elif collection.mode == 'w':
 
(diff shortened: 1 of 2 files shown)


sgillies:
@rbuffat thank you! @snorfalorpagus thanks for the review!

PR #922, opened 2020-08-18
Include FileGDB in travis and fix related issues

rbuffat:
FileGDB is included in supported_drivers, but currently not tested. 

This PR:
- Adds FileGDB support to travis using https://github.com/Esri/file-geodatabase-api
- Updates drvsupport for  FileGDB
- Fixes an issue when a driver does not support transitions (such as FileGDB), writing records results in `fiona.errors.TransactionError: Failed to commit transaction` (https://travis-ci.org/github/rbuffat/Fiona/jobs/718900437)

It seems that ints written with the FileGDB driver are read as floats back. Not sure why.

I'm not that much of a fan of using the FileGDB API, but as long as we include FileGDB in drvsupport, I prefer if we test the driver. If it causes problems, the support can be removed again.

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -1227,11 +1227,13 @@ cdef class WritingSession(Session):
             if record["geometry"] is None:
                 return True
             return record["geometry"]["type"].lstrip("3D ") in valid_geom_types
-
-        log.debug("Starting transaction (initial)")
-        result = gdal_start_transaction(self.cogr_ds, 0)
-        if result == OGRERR_FAILURE:
-            raise TransactionError("Failed to start transaction")
+        transactions_supported = check_capability_transaction(self.cogr_ds)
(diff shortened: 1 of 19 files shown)


fiona/ogrext.pyx
```diff
@@ -1227,11 +1227,13 @@ cdef class WritingSession(Session):
...
                 return True
             return record["geometry"]["type"].lstrip("3D ") in valid_geom_types
-
-        log.debug("Starting transaction (initial)")
-        result = gdal_start_transaction(self.cogr_ds, 0)
-        if result == OGRERR_FAILURE:
-            raise TransactionError("Failed to start transaction")
+        transactions_supported = check_capability_transaction(self.cogr_ds)
+        log.debug("Transaction supported: {}".format(transactions_supported))
+        if transactions_supported:
```
sgillies, line 1232: Great! Alternative would be to move the test into `gdal_start_transaction` itself but we could do that layer.

Issue #919, opened 2020-07-15
fiona.transform.transform - different behavior based on GDAL version


calebrob6:
fiona.transform.transform exhibits different behavior based on the underlying gdal library installed.

## Expected behavior and actual behavior.

If I run `fiona.transform.transform("epsg:4326", "epsg:3857", [-75.71], [38.06])` I expect to get: `([-8427998.647958742], [4587905.271362515])` regardless of library version, however with gdal 3.0.2 and fiona 1.8.13.post1 I get the incorrect result.

In particular, the `xs`, `ys` position are swapped in the arguments, i.e. if I run `fiona.transform.transform("epsg:4326", "epsg:3857", [38.06], [-75.71])` then I do get the correct result.

@snowman2 suggested that this could be related to a PYPROJ issue, https://github.com/pyproj4/pyproj/issues/538 (see https://fiona.groups.io/g/main/topic/75506516#60 for comments).

## Steps to reproduce the problem.

```
conda create --name test python gdal fiona # this installs gdal 3.0.2 and fiona 1.8.13.post1
conda activate test
python
>>> import fiona.transform
>>> fiona.transform.transform("epsg:4326", "epsg:3857", [-75.71], [38.06])
([4236819.819591993], [-13244913.18166699])
```

If I switch the lat/lon position then I get the correct result (http://epsg.io/transform#s_srs=4326&t_srs=3857&x=-75.7169053&y=38.0624956):
```
>>> fiona.transform.transform("epsg:4326", "epsg:3857", [38.06], [-75.71])
([-8427998.647958742], [4587905.271362515])
```

If I instead do the conda install with a gdal version less than 3 I get the expected result:
```
conda create --name test2 python "gdal<3" fiona # this installs gdal 2.3.3 and fiona 1.8.4
conda activate test2
python
>>> fiona.transform.transform("epsg:4326", "epsg:3857", [-75.71], [38.06])
([-8427998.647958742], [4587905.27136252])
```

## Operating system

I have tested this with conda on Windows 10 and Ubuntu 18.04.

## Fiona and GDAL version and provenance

@snorfalorpagus has reproduced with different versions on Travis - https://travis-ci.org/github/snorfalorpagus/Fiona/builds/708143904.

sgillies:
[...]

PR #926, opened 2020-08-26
Fix axis ordering for fiona.transform

rbuffat:
Fix axis ordering of fiona.transform (see https://github.com/Toblerity/Fiona/issues/919)


sgillies:
Thanks @rbuffat !

PR #929, opened 2020-08-28
Silence errors on overwriting GeoJSON

### LINKED ISSUE #771: Overwriting GeoJSON logs an ERROR but succeeds

sgillies:
Resolves #771

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -26,6 +26,7 @@ from fiona._env import GDALVersion, get_gdal_version_num, calc_gdal_version_num
 from fiona._err import cpl_errs, FionaNullPointerError, CPLE_BaseError, CPLE_OpenFailedError
 from fiona._geometry import GEOMETRY_TYPES
 from fiona import compat
+from fiona.env import Env
 from fiona.errors import (
     DriverError, DriverIOError, SchemaError, CRSError, FionaValueError,
     TransactionError, GeometryTypeValidationError, DatasetDeleteError,
@@ -964,7 +965,7 @@ cdef class WritingSession(Session):
 
             else:
                 self._fileencoding = userencoding or self._get_fallback_encoding()
-                
+
             before_fields = self.get_schema()['properties']
 
         elif collection.mode == 'w':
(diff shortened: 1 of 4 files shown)


fiona/ogrext.pyx
```diff
@@ -991,27 +992,36 @@ cdef class WritingSession(Session):
             # TODO: revisit the logic in the following blocks when we
             # change the assumption above.
             else:
-                if collection.driver == "GeoJSON" and os.path.exists(path):
-                    # manually remove geojson file as GDAL doesn't do this for us
+                log.debug("Path exists: path=%r, driver=%r", path, collection.driver)
+
```
sgillies, line 996:
```suggestion


```

sgillies, line 996:
```suggestion
```

rbuffat:
@sgillies 
Just FYI / inspiration: In https://github.com/Toblerity/Fiona/pull/909 some parts of the dataset creation are modified.  (https://github.com/Toblerity/Fiona/blob/6ee5939f1bcfe64cbedea454f689f8a1649922f6/fiona/ogrext.pyx#L1026-L1064 (all *_kwargs can be ignored))

Currently, there is a special treatment for the GeoJSON that overwrites local files.  Would it make sense to have an overwrite=True/False flag in fiona.open() instead, that depending of the result of CPLCheckForFile either returns an error or overwrites? If you are interested, I could create a new PR with just this part.

sgillies:
@rbuffat I'm :-1: on an overwrite option. I think the mode argument should cover everything. That said, I think I might be implementing "w" wrong for multi-layer formats. It doesn't truncate to zero (overwrite) in that case, where maybe it should. We could require "r+" to add layers to a multi-layer format?

rbuffat:
> It doesn't truncate to zero (overwrite) in that case, where maybe it should

If I understand correctly the issue is that is not possible to differentiate between creating a new layer, or overwriting an existing file.  If I understand correctly the OGR model strictly separates read and write mode, thus "r+" might be a bit confusing.

Regarding the overwrite aspect, would you just use the existing behavior or having something to separate "w" for writing if the file or layer does not exist and "w!" or "w+" to overwrite a file or layer if it already exists.

PR #930, opened 2020-08-28
Test default overwrite behavior

rbuffat:
I was not sure what the default behavior of Fiona is to overwrite an existing file, thus I created this test. The test seems to fail for some drivers. I did not yet look closely at why they are failing, it could also be due to the test. Also, I'm unsure if in this case, Fiona needs to behave similarly for all drivers.


sgillies:
@rbuffat in the interest of reinvigorating this project, I'm going to close this one. Testing driver-specific overwriting behavior should wait until the 2.0 milestone.

PR #931, opened 2020-08-29
New and improved MemoryFile

### LINKED ISSUE #906: Bring MemoryFile improvements from rasterio 1.1 (seeking, etc) here
https://github.com/mapbox/rasterio/pull/1935

To be done before #901.

sgillies:
Based on rasterio's.

Resolves #906


fiona/ogrext.pyx
```diff
@@ -1663,65 +1659,75 @@ def remove_virtual_file(vsi_filename):
     return VSIUnlink(vsi_cfilename)
 
 
-cdef class MemoryFileBase(object):
+cdef class MemoryFileBase:
     """Base for a BytesIO-like class backed by an in-memory file."""
```
rbuffat, line 1664: Would it be an option to put `cdef VSILFILE * _vsif`here before the init() instead of the new MemoryFileBase C class in ogrext.pxd?

sgillies: That's a good suggestion. I'll try it.

sgillies: 🎉

PR #934, opened 2020-08-31
Keep VSI filenames in sync between write and read

sgillies:
As @rbuffat suggested, we will use driver-supported file extensions when creating a Collection in the context of a default, unnamed MemoryFile.


fiona/ogrext.pyx
```diff
@@ -1723,13 +1722,31 @@ cdef class MemoryFileBase:
...
+
+        if extension_c != NULL:
+            extension_b = extension_c
+            recommended_extension = extension_b.decode("utf-8")
+            if not recommended_extension.startswith("."):
+                recommended_extension = "." + recommended_extension
+            root, ext = os.path.splitext(self.name)
+            if not ext:
+                log.info("Setting extension: root=%r, extension=%r", root, recommended_extension)
+                self.name = root + recommended_extension
```
sgillies, line 1749: Here's the code to find and use the DMD_EXTENSION.

fiona/ogrext.pyx
```diff
@@ -1741,31 +1758,30 @@ cdef class MemoryFileBase:
...
         cdef vsi_l_offset buffer_len = 0
         cdef unsigned char [:] buff_view
 
-        buffer = VSIGetMemFileBuffer(self._path, &buffer_len, 0)
+        name_b = self.name.encode('utf-8')
+        buffer = VSIGetMemFileBuffer(name_b, &buffer_len, 0)
 
         if buffer == NULL or buffer_len == 0:
-            buff_view = None
+            buff_view = memoryview(b"")
```
sgillies, line 1784: Allows us to revert to a simpler `__len__()` implementation.

fiona/ogrext.pyx
```diff
@@ -1723,13 +1722,36 @@ cdef class MemoryFileBase:
...
         if self._vsif == NULL:
-            self._vsif = VSIFOpenL(self.name.encode("utf-8"), self.mode.encode("utf-8"))
+            self._vsif = VSIFOpenL(name_b, self.mode.encode("utf-8"))
+
+    def _ensure_extension(self, drivername=None):
+        """Ensure that the instance's name uses a file extension supported by the driver."""
+        # Avoid a crashing bug with GDAL versions < 2.
+        gdal_version = GDALVersion.runtime()
+        if gdal_version.major < 2:
+            return
```
sgillies, line 1739: We got a segfault on Travis when this was called. I'm not inclined to try to make Fiona do the right thing with GDAL 1.11.

Issue #935, opened 2020-09-01
test_write_or_driver_error[DGN] & test_append_or_driver_error[DGN] fail on several architectures


sebastic:
The Debian package build for Fiona 1.8.14 failed on several architectures due to test failures:

**i386** ([full buildlog](https://buildd.debian.org/status/fetch.php?pkg=fiona&arch=i386&ver=1.8.14-1&stamp=1598936045&file=log))
```
=================================== FAILURES ===================================
_______________________ test_write_or_driver_error[DGN] ________________________

tmpdir = local('/tmp/pytest-of-buildd/pytest-0/test_write_or_driver_error_DGN0')
driver = 'DGN'
testdata_generator = <function testdata_generator.<locals>._testdata_generator at 0xef24d190>

    @pytest.mark.parametrize('driver', [driver for driver, raw in supported_drivers.items() if 'w' in raw])
    def test_write_or_driver_error(tmpdir, driver, testdata_generator):
        """
            Test if write mode works.
    
        """
    
        if driver == "BNA" and GDALVersion.runtime() < GDALVersion(2, 0):
            # BNA driver segfaults with gdal 1.11
            return
    
        schema, crs, records1, _, test_equal = testdata_generator(driver, range(0, 10), [])
        path = str(tmpdir.join(get_temp_filename(driver)))
    
        if driver in driver_mode_mingdal['w'] and GDALVersion.runtime() < GDALVersion(
                *driver_mode_mingdal['w'][driver][:2]):
    
            # Test if DriverError is raised for gdal < driver_mode_mingdal
            with pytest.raises(DriverError):
                with fiona.open(path, 'w',
                                driver=driver,
                                crs=crs,
                                schema=schema) as c:
                    c.writerecords(records1)
    
        else:
    
            # Test if we can write
            with fiona.open(path, 'w',
                            driver=driver,
                            crs=crs,
                            schema=schema) as c:
    
                c.writerecords(records1)
    
            if driver in {'FileGDB', 'OpenFileGDB'}:
                open_driver = driver
            else:
                open_driver = None
            with fiona.open(path, driver=open_driver) as c:
                assert c.driver == driver
                items = list(c)
                assert len(items) == len(records1)
                for val_in, val_out in zip(records1, items):
>                   assert test_equal(driver, val_in, val_out), "in: {val_in}, out: {val_out}".format(val_in=val_in,
                                                                                                      val_out=val_out)
[...]

Issue #933, opened 2020-08-30
[Tests] Tests should not rely on write order


rbuffat:
Currently, some tests assume that the reading order of records is the same as they are written. This is not the case for all drivers.

See https://github.com/Toblerity/Fiona/pull/924

sgillies:
@rbuffat historically, the tests were all based on shapefiles, which was at the time the most used and most tested driver. And we did assume preservation of order.

I would like us to be conservative about changing tests to be more specific to particular drivers. It's not Fiona's job, in my view, to find and document bugs and quirks in GDAL/OGR drivers. And let's be clear: many of the GDAL vector drivers are of historical interest at best and are fine to ignore. I'd like for us to focus on implementing properly our feature access API without thinking too much about individual GDAL drivers. Concretely, let's firm up the semantics for Fiona opening modes and make sure that they map to the set of OGR driver capabilities, but let's not write tests that are specific to a particular driver or add code to Fiona that works around limitations of drivers.

I'm sorry that this comment is a little vague and hand-waving.

rbuffat:
@sgillies I was thinking about refactoring the test_equal method of the testdata_generator to work with a collection of records instead of individual records and only refactoring the tests using it. https://github.com/Toblerity/Fiona/blob/7559619bef3e2e3a2d712b674c3c7c358551ccae/tests/conftest.py#L398-L409

I see the benefit of testing all enabled drivers in having more confidence if a new driver is enabled. But I also see that the complexity of the tests increases. There needs to be a balance between benefits and complexity.

PR #936, opened 2020-09-01
Refactor testdata_generator

rbuffat:
Refactor test function of testdata_generator:
- Test a collection of records instead of only single records and do not assume any order (resolves https://github.com/Toblerity/Fiona/issues/933)
- Only compare ints (should resolve https://github.com/Toblerity/Fiona/issues/935 )

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,6 +2,7 @@
 
 import copy
 import json
+import logging
 import os.path
 import shutil
 import tarfile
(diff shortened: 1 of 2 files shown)


rbuffat:
Would it make sense that all tests for drvsupport.py using the testgenerator run only on the CI or with a special marker?

QuLogic:
I attempted to rebase this to apply to 1.8.18, but unfortunately, it does not appear to fix #935. The logging says "Could not extract unique elements from test_data."

tests/conftest.py
```diff
@@ -395,18 +396,38 @@ def get_records2(driver, range):
...
+            logging.error("Could not extract unique elements from test_data.")
+            return False
+
+        for test_record in test_data:
+            pos = _get_pos(driver, test_record)
+            if pos not in positions:
+                logging.error("Test record {} corresponds not to a record from true_data.".format(test_record))
+                return False
+
+        return True
```
QuLogic, line 430: You should just `assert` these things and let pytest print out the parts that don't match.

rbuffat: The assert is in the corresponding tests.

QuLogic: It's less effective; the tests will just `assert False` and that's all pytest will show. Then you have to go through the log to figure out what from here failed. If you assert here, then pytest will show the individual parts of what are not matching, and the backtrace will point to the actual failure.

snorfalorpagus: I agree with @QuLogic. We should let pytest do the work for us.

```python
assert len(actual_data) == len(expected_data)
```

Gives:

```
    def test_me():
        expected_data = [1,2,3]
        actual_data = [1,2]
>       assert len(actual_data) == len(expected_data)
E       assert 2 == 3
E        +  where 2 = len([1, 2])
E        +  and   3 = len([1, 2, 3])

tests/test_me.py:4: AssertionError
```

sgillies:
@rbuffat I'm going to reject this one for 1.8 and 1.9.

Issue #937, opened 2020-09-01
Memfile segfault for 'MapInfo File' driver


rbuffat:
The following code segfaults Fiona. The code works with other drivers, thus I assume it is a GDAL issue.

```
from collections import OrderedDict
from io import BytesIO

import fiona

driver = 'MapInfo File'
schema = {'geometry': 'Point', 'properties': OrderedDict([('position', 'str')])}

with BytesIO() as fout:
    with fiona.open(fout,
                    'w',
                    driver=driver,
                    schema=schema) as c:
        pass
```

sgillies:
Confirmed with GDAL 3.1.

sgillies:
Oh, I see. We're calling "MITAB" "MapInfo File" in the drvsupport module. I don't remember whether that was the original name of the format driver or just an error.

rbuffat:
@sgillies I think it is this driver: https://gdal.org/drivers/vector/mitab.html#driver-capabilities
https://github.com/OSGeo/gdal/blob/73117389a4e0c3ce530b5029c1b5dafed47c7274/gdal/ogr/ogrsf_frmts/mitab/mitab_ogr_driver.cpp#L204

sgillies:
The MapInfo File driver writes a vsimem file that VSIFOpenL fails to open. I don't understand why, but I'll make a change so that we raise OSError instead of crashing from the null pointer.

sgillies:
@rbuffat it occurs to me that the multi-file nature of the format may be involved.

rbuffat:
> The MapInfo File driver writes a vsimem file that VSIFOpenL fails to open. I don't understand why, but I'll make a change so that we raise OSError instead of crashing from the null pointer.

This is a good point. Using BytesIO() makes only sense for single file formats!

rbuffat:
Fixed in Fiona 1.8.17

PR #938, opened 2020-09-01
Raise OSError when VSIFOpenL fails

sgillies:
Prevents the crash reported in #937. Look reasonable to you @rbuffat ?


rbuffat:
Looks good! Thanks, @sgillies

PR #939, opened 2020-09-02
Add  function to list contents of a directory

rbuffat:
I feel https://github.com/Toblerity/Fiona/pull/901 has grown too big. I think it's best to split it into several parts.

This PR adds functionality to list the content of a directory using GDAL VSI function. Thus, this works also to list directories in a zip file.


sgillies:
Thanks for splitting it up @rbuffat ! I think this will have to wait for the 1.9.0 release since it adds a new function to the API.

sgillies (approved):

PR #940, opened 2020-09-02
this_schema not used for fp_writer

rbuffat:
`this_schema` and the conversion to OrderedDict is not used. Unsure though if it is needed at all, as there seemed to be no issue without it.

https://github.com/Toblerity/Fiona/blob/c2cbbab04ff7d3eaf1a11ab767d7f86e918a5edf/fiona/__init__.py#L219-L232


sgillies (approved):
Thanks!

Issue #941, opened 2020-09-03
TypeError on import with GDAL version 3.1+


tomplex:
## Expected behavior and actual behavior.

`import fiona` succeeds without error when built against the most recent version of GDAL. However with GDAL 3.1.2 and 3.2.0, there is a `TypeError` on import.

## Steps to reproduce the problem.

```Dockerfile
FROM osgeo/gdal

RUN apt update && \
    apt install -y \
    python3-pip

RUN pip3 install --no-binary fiona fiona

ENTRYPOINT ["python3", "-c", "import fiona"]
```

```bash
docker build -t test-fiona .
docker run test-fiona
```

This results in a `TypeError`:

```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/fiona/__init__.py", line 86, in <module>
    from fiona.collection import BytesCollection, Collection
  File "/usr/local/lib/python3.8/dist-packages/fiona/collection.py", line 11, in <module>
    from fiona.ogrext import Iterator, ItemsIterator, KeysIterator
  File "fiona/ogrext.pyx", line 1, in init fiona.ogrext
TypeError: C function fiona._shim.get_field_as_datetime has wrong signature (expected __pyx_ctuple_e69d__int__and_int__and_int__and_int__and_int__and_int__and_float__and_int__etc (void *, int), got __pyx_ctuple_1cb47__int__and_int__and_int__and_int__and_int__and_int__and_float__and_int__etc (void *, int))
```

## Operating system & GDAL version

I first observed this issue on Debian Stretch using GDAL 3.1.2, but I can't share that image as it's work-related. Fortunately I was able to replicate using the latest GDAL images.

```bash
# lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04 LTS
Release:        20.04
Codename:       focal

# gdal-config --version
3.2.0

# ogr2ogr --version
GDAL 3.2.0dev-603eb8ac0327b25c8b85ad8ef8958290405482ef, released 2020/09/03
```

jorisvandenbossche:
We are seeing the same error on the conda-forge builds: https://github.com/conda-forge/fiona-feedstock/pull/166 
And indeed, the GDAL version might be it, on conda-forge it is building with GDAL 3.1.2. Locally it was working for me, but it seems I was using GDAL 3.0.4. After updating to 3.1.2, it also fails locally.

[...]

PR #942, opened 2020-09-03
Refactor _shim*.pyx to not return tuples

rbuffat:
It looks as for each tuple and shim a different tuple typedef is created. This leads to issues for releases: https://github.com/Toblerity/Fiona/issues/941

This PR removes all tuple return values in the _shim modules.


rbuffat:
Updating the pointer loos really strange, but:

> Note that Cython uses array access for pointer dereferencing, as *x is not valid Python syntax, whereas x[0] is.
https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html

sgillies (approved):

jorisvandenbossche:
Thanks @rbuffat !

PR #945, opened 2020-09-04
Refactor _loading to avoid circular import

rbuffat:
@sgillies is this what you had in mind? 

I'm still a bit concerned regarding the `import fiona._err`, especially as we do not test this case on the CI. Would a dedicated module for just this task make sense?

### DIFF
--- a/fiona/_loading.py
+++ b/fiona/_loading.py
@@ -16,36 +16,11 @@ log.addHandler(logging.NullHandler())
 
 @contextlib.contextmanager
 def add_gdal_dll_directories():
-    dll_directories = []
-
-    if platform.system() == 'Windows' and sys.version_info >= (3, 8):
-
-        # if loading of extension modules fails, search for gdal dll directories
-        try:
-            import fiona._err
-        except ImportError as e:
-
-            def directory_contains_gdal_dll(path):
-                """ Checks if a directory contains a gdal dll """
-                return len(glob.glob(os.path.join(path, "gdal*.dll"))) > 0
-
-            def search_gdal_dll_directories():
-                """ Search for gdal dlls
-
-                    Finds directories in PATH containing gdal.dll.
-                """
-                # Parse PATH for gdal/bin
-                for path in os.getenv('PATH', '').split(os.pathsep):
-                    if directory_contains_gdal_dll(path):
-                        dll_directories.append(path)
-                if len(dll_directories) == 0:
-                    log.warning("No dll directories found.")
-
-            search_gdal_dll_directories()
-
     dll_dirs = []
-    for dll_directory in dll_directories:
-        dll_dirs.append(os.add_dll_directory(dll_directory))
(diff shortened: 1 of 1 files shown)


sgillies:
@rbuffat yes, that's what I had in mind. What would be the impact of always adding DLL directories for Windows/3.8 instead of conditionally? There would be some duplicated effort, yes, but DLL directories would only be searched and added once for each extension module import.

rbuffat:
A potential issue I see is that for conda or Christoph Gohlke packages there is no need to load additional DLLs (I assume @jorisvandenbossche ). If a user of these packages has additional software installed, that requires GDAL in its path, then these DLLs are also added. I'm not sure what the consequences are if incompatible gdal DLLs are added. I guess that on Windows more Fiona users will use these packages instead of compiling Fiona from source.

sgillies:
Here's what the numpy project does: https://github.com/numpy/numpy/blob/25fa87a92d39c4a7de72f79263c9048d657b1226/numpy/distutils/misc_util.py#L2307-L2317. It standardizes a ".libs" directory and isn't extendable. Could we do just this for our own appveyor-produced wheels? Gohlke has his own approach. Conda has patched Python. We don't need to support any other packagers of fiona as far as I know.

rbuffat:
On appveyor we manually copy the dlls to the .libs folder.  Should this be included in setup.py for users that want to install Fiona from source for themself? Or should we ignore this use case? I'm not sure if there are a lot of users doing that, as this is rather complicated.

sgillies:
@rbuffat let's ignore that use case for now.

fiona/_loading.py
```diff
@@ -13,45 +13,14 @@
...
 @contextlib.contextmanager
 def add_gdal_dll_directories():
-
     dll_dirs = []
-    for dll_directory in dll_directories:
-        dll_dirs.append(os.add_dll_directory(dll_directory))
+    if platform.system() == 'Windows' and sys.version_info >= (3, 8):
+        dll_directory = os.path.join(os.path.dirname(__file__), '.libs')
+        if os.path.exists(dll_directory):
+            dll_dirs.append(os.add_dll_directory(dll_directory))
```
sgillies, line 23: Yep, this looks great!

sgillies (approved):

jorisvandenbossche:
> A potential issue I see is that for conda or Christoph Gohlke packages there is no need to load additional DLLs (I assume @jorisvandenbossche ).

While I am a "maintainer" of the conda packages to help out, I am not very well versed in the building details. Probably better to ping eg @ocefpaf for specific questions

jorisvandenbossche:
@rbuffat and thanks for looking into this!

ocefpaf:
I'm a complete dummy on Windows. Let's test this as a patch in https://github.com/conda-forge/fiona-feedstock/pull/167, I'm sending the PR in a moment.

ocefpaf:
BTW, it worked as expected. Thanks!

PR #948, opened 2020-09-09
ENH: Auto-detect driver in writing mode

snowman2:
Is this something you would be interested in? If so, I can add some tests, etc. Target version 1.9.

### DIFF
--- a/fiona/drvsupport.py
+++ b/fiona/drvsupport.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
-from fiona.env import Env
+import os
+
+from fiona.env import Env, require_gdal_version
 from fiona._env import get_gdal_version_tuple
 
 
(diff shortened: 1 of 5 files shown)


rbuffat:
There is no 1:1 relation between drivers and extensions. A driver can have none or multiple extensions. The same extension can be understood by multiple drivers (e.g. gdb FileGDB / OpenFileGDB, EsriJSON / TOPOJSON / GeoJSON, KML/LIBKLM, ...
). This will make it challenging to determine the correct driver for each situation.

snowman2:
> This will make it challenging to determine the correct driver for each situation.

It won't be perfect for sure. One option would be to limit it to a subset for the guess that we are more certain about. For example, it could only attempts to guess shapefiles, geopackages, and geojson and raise a `DriverError` if it cannot find the appropriate driver for the extension.

snowman2:
> The same extension can be understood by multiple drivers (e.g. gdb FileGDB / OpenFileGDB, EsriJSON / TOPOJSON / GeoJSON, KML/LIBKLM, ...
).

It looks like only a subset support the `w` mode. So, filtering by that mode resolves the duplicate issue for the moment I think.

snowman2:
Would be good to be able to build off of #950

fiona/drvsupport.py
```diff
@@ -191,12 +223,67 @@ def _filter_supported_drivers():
...
+    dict:
+        Map of extensions to the driver.
+    """
+    from fiona.meta import extensions  # prevent circular import
+
+    extension_to_driver = _extension_to_driver.copy()
+    for drv, modes in supported_drivers.items():
+        # update extensions based on driver suppport
+        for extension in extensions(drv) or ():
+            if "w" in modes:
```
snowman2, line 248: I figured this would be better than checking if GDAL supports it so that a better error message will be raised if a driver is selected that is not supported downstream.

snowman2:
Should be ready for review. Apologies for the noise - attempting to minimize the diff.

fiona/drvsupport.py
```diff
@@ -197,6 +229,60 @@ def _filter_supported_drivers():
...
+
+@require_gdal_version("2.0")
+def vector_driver_extensions():
+    """
+    Returns
+    -------
+    dict:
+        Map of extensions to the driver.
+    """
+    from fiona.meta import extensions  # prevent circular import
```
snowman2, line 241: One thought I had was to add a `_meta.pyx` file and move the method(s) from `ogrext.pyx` there to prevent circular imports. Another option, would be to move these methods to the `meta.py`. Thoughts?

sgillies:
Let's keep this for 1.9.

sgillies (approved):
Let's do this!

snowman2:
Thanks @sgillies 👍

PR #955, opened 2020-09-12
BUG: Allow property name as empty string

snowman2:
Ref: https://github.com/geopandas/geopandas/issues/1611

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -665,11 +665,9 @@ cdef class Session:
                 raise ValueError("Null field definition")
 
             key_c = OGR_Fld_GetNameRef(cogr_fielddefn)
+            if key_c == NULL:
+                raise ValueError("Empty field name at index: %s" % i)
             key_b = key_c
-
-            if not bool(key_b):
-                raise ValueError("Invalid field name ref: %s" % key)
-
             key = key_b.decode(encoding)
 
             if key in ignore_fields:
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -666,7 +666,7 @@ cdef class Session:
 
             key_c = OGR_Fld_GetNameRef(cogr_fielddefn)
             if key_c == NULL:
-                raise ValueError("Empty field name at index: %s" % i)
+                raise ValueError("Field name NULL at index: %s" % i)
             key_b = key_c
             key = key_b.decode(encoding)
 
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -192,12 +192,14 @@ cdef class FeatureBuilder:
         for i in range(OGR_F_GetFieldCount(feature)):
             fdefn = OGR_F_GetFieldDefnRef(feature, i)
             if fdefn == NULL:
-                raise ValueError("Null feature definition")
+                raise ValueError("NULL field definition at index {}".format(i))
             key_c = OGR_Fld_GetNameRef(fdefn)
             if key_c == NULL:
-                raise ValueError("Null field name reference")
+                raise ValueError("NULL field name reference at index {}".format(i))
             key_b = key_c
             key = key_b.decode(encoding)
+            if not key:
+                warnings.warn("Empty field name at index {}".format(i))
 
             if key in ignore_fields:
                 continue
@@ -662,13 +664,15 @@ cdef class Session:
         for i from 0 <= i < n:
             cogr_fielddefn = OGR_FD_GetFieldDefn(cogr_featuredefn, i)
             if cogr_fielddefn == NULL:
-                raise ValueError("Null field definition")
+                raise ValueError("NULL field definition at index {}".format(i))
 
             key_c = OGR_Fld_GetNameRef(cogr_fielddefn)
             if key_c == NULL:
-                raise ValueError("Field name NULL at index: %s" % i)
+                raise ValueError("NULL field name reference at index {}".format(i))
             key_b = key_c
             key = key_b.decode(encoding)
+            if not key:
+                warnings.warn("Empty field name at index {}".format(i))
 
             if key in ignore_fields:
                 log.debug("By request, ignoring field %r", key)
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -944,14 +944,19 @@ def test_mask_polygon_triangle(tmpdir, driver, filename):
 def test_collection__empty_column_name(tmpdir):
     """Based on pull #955"""
     tmpfile = str(tmpdir.join("test_empty.geojson"))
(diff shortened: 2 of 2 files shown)


sgillies:
@snowman2 the less driver-specific code we have in this part of Fiona, the better, so I hesitate to accept this. I think we must also check with GDAL's maintainer about whether empty field names are allowed in the OGR data model (see https://gdal.org/api/ogrfeature_cpp.html#_CPPv412OGRFieldDefn) or if this only works by accident. Would you be willing to ask on gdal-dev?

snowman2:
> I think we must also check with GDAL's maintainer about whether empty field names are allowed in the OGR data model

I did some testing with `ogr2ogr` and `ogrinfo` and it seems like the empty string is supported across drivers:

```
ogr2ogr -f GPKG samplecafe_halfhalfhalfhalfhalfEmpty2.gpkg samplecafe_halfhalfhalfhalfhalfEmpty2.geojson
ogr2ogr -f "ESRI Shapefile" samplecafe_halfhalfhalfhalfhalfEmpty2.shp samplecafe_halfhalfhalfhalfhalfEmpty2.geojson
```
```
INFO: Open of `samplecafe_halfhalfhalfhalfhalfEmpty2.gpkg'
      using driver `GPKG' successful.

Layer name: samplecafe_halfhalfhalfhalfhalfEmpty2
Geometry: Point
Feature Count: 2
Extent: (8.690787, 49.409640) - (8.694292, 49.414744)
Layer SRS WKT:
GEOGCS["WGS 84",
    DATUM["WGS_1984",
        SPHEROID["WGS 84",6378137,298.257223563,
            AUTHORITY["EPSG","7030"]],
        AUTHORITY["EPSG","6326"]],
    PRIMEM["Greenwich",0,
        AUTHORITY["EPSG","8901"]],
    UNIT["degree",0.0174532925199433,
        AUTHORITY["EPSG","9122"]],
    AXIS["Latitude",NORTH],
    AXIS["Longitude",EAST],
    AUTHORITY["EPSG","4326"]]
Data axis to CRS axis mapping: 2,1
FID Column = fid
Geometry Column = geom
@osmId: String (0.0)
@validFrom: DateTime (0.0)
@validTo: DateTime (0.0)
addr:city: String (0.0)
addr:housenumber: String (0.0)
addr:postcode: String (0.0)
addr:street: String (0.0)
addr:suburb: String (0.0)
amenity: String (0.0)
cuisine: String (0.0)
diet:vegan: String (0.0)
diet:vegetarian: String (0.0)
name: String (0.0)
opening_hours: String (0.0)
outdoor_seating: String (0.0)
phone: String (0.0)
smoking: String (0.0)
source: String (0.0)
toilets:wheelchair: String (0.0)
website: String (0.0)
wheelchair: String (0.0)
: String (0.0)
created_by: String (0.0)
```

> ... or if this only works by accident

That is definitely a question for the `gdal-dev` mailing list. I will check the GDAL tests first and if I don't find anything I will post the question there.

snowman2:
From what I can tell from the code, it allows any string for the field name, even an empty one:

[pszName](https://github.com/OSGeo/gdal/blob/255f08f93b70e23643aca62af61448f9295c67a8/gdal/ogr/ogr_feature.h#L97) is of type `char *`.

And [OGRFieldDefn::SetName](https://github.com/OSGeo/gdal/blob/255f08f93b70e23643aca62af61448f9295c67a8/gdal/ogr/ogrfielddefn.cpp#L162) doesn't have any constraints on what can be set to the field name, so even an empty field name is allowed.

snowman2:
Based on that, this check for an empty field name could be removed entirely. I will send a note to the list to ask about whether this behavior is intended as I don't see any tests explicitly testing this.

snowman2:
Interesting discovery ... the CSV driver seems to give the field a name based on the index (`field_22`):
```
ogrinfo -so samplecafe_halfhalfhalfhalfhalfEmpty2.csv samplecafe_halfhalfhalfhalfhalfEmpty2
INFO: Open of `samplecafe_halfhalfhalfhalfhalfEmpty2.csv'
      using driver `CSV' successful.

Layer name: samplecafe_halfhalfhalfhalfhalfEmpty2
Geometry: None
Feature Count: 2
Layer SRS WKT:
(unknown)
@osmId: String (0.0)
@validFrom: String (0.0)
@validTo: String (0.0)
addr:city: String (0.0)
addr:housenumber: String (0.0)
addr:postcode: String (0.0)
addr:street: String (0.0)
addr:suburb: String (0.0)
amenity: String (0.0)
cuisine: String (0.0)
diet:vegan: String (0.0)
diet:vegetarian: String (0.0)
name: String (0.0)
opening_hours: String (0.0)
outdoor_seating: String (0.0)
phone: String (0.0)
smoking: String (0.0)
source: String (0.0)
toilets:wheelchair: String (0.0)
website: String (0.0)
wheelchair: String (0.0)
field_22: String (0.0)
created_by: String (0.0)
```

snowman2:
For reference: https://lists.osgeo.org/pipermail/gdal-dev/2020-September/052647.html

fiona/ogrext.pyx
```diff
@@ -665,11 +665,9 @@ cdef class Session:
                 raise ValueError("Null field definition")
 
             key_c = OGR_Fld_GetNameRef(cogr_fielddefn)
+            if key_c == NULL:
+                raise ValueError("Empty field name at index: %s" % i)
```
sgillies, line 669: "Empty" isn't correct because you intend to allow empty strings, yes? How about "Field name is NULL at index"?

I think it's highly unlikely we'll see a NULL field name, but we do need to avoid passing a NULL on in our code.

sgillies (changes requested):
I'm not in favor of helping users publish bad data (empty field names) but this does fix a reported bug and will help people read and transform bad data, so let's do it.

fiona/ogrext.pyx
snowman2: Agreed :+1:

snowman2:
> I'm not in favor of helping users publish bad data (empty field names) but this does fix a reported bug and will help people read and transform bad data, so let's do it.

Yeah, not super great. How about adding a warning instead? Probably good for the user to know that they are playing with :fire:.

sgillies:
Great idea @snowman2 . A warning would be appropriate and very welcome.

snowman2:
I found that the `FeatureBuilder` was doing the same thing and modified it so the errors and warnings were the same.

sgillies (approved):
@snowman2 okay with you if I customize the warning type after I merge this?

snowman2:
Sure, feel free to adjust as you want 👍

snowman2:
Thanks for the review @sgillies :+1:

PR #956, opened 2020-09-17
Merge maint 1.8

sgillies:


sgillies:
@rbuffat I've done it! The merge-maint-1.8 branch has commits from maint-1.8 and resolves conflicts. Unless anything strikes you as wrong, I'll merge this later today and maint-1.9 will be up to date.

docs/manual.rst
```diff
@@ -1007,7 +1004,7 @@ a schema of ``{'properties': {'bar': 'int', 'foo': 'str'}}`` will produce
 a shapefile where the first field is 'foo' and the second field is 'bar'. If
 you want 'bar' to be the first field, you must use a list of property items
 
-.. sourcecode:: python
+.. sourcecode:: pycon
```
snowman2, line 1007: Is that correct?

sgillies: Good catch!

docs/manual.rst
```diff
@@ -271,10 +270,6 @@ collection to get back to the beginning.
 Collection indexing
 -------------------
 
-.. admonition::
-
-   New in version 1.1.6
-
 Features of a collection may also be accessed by index.
 
 .. code-block:: pycon
```
snowman2, line 275: Another one here.

docs/manual.rst
```diff
@@ -303,13 +298,15 @@ Features of a collection may also be accessed by index.
...
 -------------
 
 A :py:class:`~fiona.collection.Collection` involves external resources. There's
 no guarantee that these will be released unless you explicitly
 :py:meth:`~fiona.collection.Collection.close` the object or use
-a :py:keyword:`with` statement. When a :py:class:`~fiona.collection.Collection`
+a :keyword:`with` statement. When a :py:class:`~fiona.collection.Collection`
 is a context guard, it is closed no matter what happens within the block.
 
 .. sourcecode:: pycon
```
snowman2, line 312: Another one here

docs/manual.rst
```diff
@@ -1157,7 +1155,7 @@ indexes.
     3
 
 To filter features by property values, use Python's builtin :py:func:`filter` and
-:py:keyword:`lambda` or your own filter function that takes a single feature
+:keyword:`lambda` or your own filter function that takes a single feature
 record and returns ``True`` or ``False``.
 
 .. sourcecode:: pycon
```
snowman2, line 1161: Another one here

docs/manual.rst
sgillies: This one looks okay to me. The following code block does have python prompts and "pycon" is intended.

We have inconsistent use of code-block and sourcecode directives, but that's out of scope.

snowman2: Sounds good, thanks for getting this MR ready :+1:

rbuffat:
@snowman2 @sgillies 

The following code block is not valid Python and rstcheck returns this as an error. I assumed that `.. sourcecode:: pycon` results in the same formatting but without the code checking. Maybe it is better to [ignore python code checks ](https://github.com/myint/rstcheck/blob/master/README.rst#ignore-specific-languages) or do not use rstcheck on travis if it is annoying. 

```
c = fiona.open(
    '/tmp/file.shp',
    'w',
    schema={'properties': [('bar', 'int'), ('foo', 'str')], ...},
    ... )
```

sgillies:
@rbuffat I fixed the example code. Thanks for the review, both!

rbuffat:
@sgillies Thank you for all the work for merging this! 

Before I only had time to look at the rstcheck issues. I now had a glance at all the changes, the only thing I found so far is that Python 2.7 and GDAL < 2.4 was eliminated from travis but not from appveyor. Should then also the _shim modules be removed?

sgillies:
@rbuffat it was a bad time :sweat_smile: I am determined to never let our branches become strangers again.

No, we won't drop the shims for 1.9. That's for 2.0. But I think it is okay to test compatibility with GDAL versions 1.11-2.3 less thoroughly than we have done.

PR #958, opened 2020-09-18
Use get_gdal_version_tuple in drvsupport

snowman2:
Using the static tuple will make it so things don't have to be re-calculated each time.

### DIFF
--- a/fiona/drvsupport.py
+++ b/fiona/drvsupport.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 from fiona.env import Env
-from fiona._env import get_gdal_version_num, calc_gdal_version_num
+from fiona._env import get_gdal_version_tuple
 
+
+_GDAL_VERSION = get_gdal_version_tuple()
 
 # Here is the list of available drivers as (name, modes) tuples. Currently,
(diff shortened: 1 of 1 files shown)


sgillies:
@snowman2 unless there's a big performance improvement that outweighs other considerations, let's keep drvsupport as it was. Statements that are executed on module import can make testing of module functions more complicated.

snowman2:
> it's trivial to test some aspects drvsupport functions against different versions of GDAL if we can patch get_gdal_version_num() after drvsupport is imported, but before the function is called.

Have you tried this?
```python
@patch("fiona.drvsupport._GDAL_VERSION", GDALVersion(1, 0, 0))
```

sgillies:
@snowman2 you're right, patching that value is even simpler than patching the function. I'm keeping my general concerns about executing statements on import, but in this specific case there's no problem at all. Thank you for the PR!

rbuffat:
I know I'm a bit late. There are currently multiple ways of checking gdal versions: GDALVersion (One in env.py and one in env.pyx), get_gdal_version_tuple and get_gdal_version_num. Each of them is used in different parts of Fiona. Would it make sense to refactor the code to only use one method? E.g. by extending GDALVersion of env.py to support also the revision/patch level?

PR #960, opened 2020-10-01
Append support for MemoryFile

rbuffat:
This PR adds append support for MemoryFile. 
It is the second part of https://github.com/Toblerity/Fiona/pull/901

~~Currently, there are a few TODO's left. GDAL includes the DCAP_VIRTUALIO metadata item: "Capability set by a driver that can read/create datasets through the VSI*L API." I will update the PR once https://github.com/Toblerity/Fiona/pull/950 is merged.~~


rbuffat:
Appveyor is failing as the DGN driver gained DCAP_VIRTUALIO support with GDAL 2.3:  https://github.com/OSGeo/gdal/blob/release/2.2/gdal/ogr/ogrsf_frmts/dgn/ogrdgndriver.cpp
https://github.com/OSGeo/gdal/blob/release/2.3/gdal/ogr/ogrsf_frmts/dgn/ogrdgndriver.cpp

rbuffat:
The test for MapInfo File fails as the geometry': {'coordinates': (0.0, 4.0), 'type': 'Point'} is read back as 'geometry': {'coordinates': (0.0, 4.01), 'type': 'Point'}. I suspect this is something related to GDAL, respectively the driver format. This test issue should be adressed in https://github.com/Toblerity/Fiona/pull/936.

Thus, I disabled the Mapinfo driver for now.

fiona/io.py
```diff
@@ -52,33 +52,33 @@ def open(self, driver=None, schema=None, crs=None, encoding=None,
...
+        if driver is not None and not supports_vsi(driver):
+            raise DriverError("Driver {} does not support virtual files.")
+
+        if mode in ('r', 'a') and not self.exists():
+            raise IOError("MemoryFile does not exist.")
+        if layer is None and mode == 'w' and self.exists():
+            raise IOError("MemoryFile already exists.")
+
+        if not self.exists() or mode == 'w':
+            if driver is not None:
```
rbuffat, line 64: Should we introduce a check that driver and schema need to be specified in write mode?

rbuffat:
When I split https://github.com/Toblerity/Fiona/pull/901 I lost the big picture. The original intention was to support not only append but all modes, including the creation of multiple layers (See https://github.com/Toblerity/Fiona/issues/830)

snorfalorpagus:
@rbuffat This looks sensible to me. Other than the MapInfo issue is this complete?

sgillies:
Let's keep this for 1.9 (I'm going rapidly through open PRs today).

sgillies:
What looked like an easy conflict resolution broke everything 😦

PR #961, opened 2020-10-05
Add 'where' SQL clause to set attribute filter

mwtoews:
This PR implements a `where` argument to set an attribute filter ([where docs](https://gdal.org/user/ogr_sql_dialect.html#where), [API docs](https://gdal.org/api/vector_c_api.html#_CPPv424OGR_L_SetAttributeFilter9OGRLayerHPKc)), which could work alongside existing `bbox` and `mask` spatial filter parameters.

Basic usage is:

```python
import fiona

c = fiona.open('tests/data/coutwildrnp.shp', 'r')

print(len(list(c.items(where="WILDRNP020 BETWEEN 450 AND 480"))))  # 14
print(len(list(c.items(where="WILDRNP020 BETWEEN 450 AND 480 AND STATE='UT'"))))  # 3
print(len(list(c.items(where="STATE='UT'"))))  # 13
```
TODO/Questions:

- [x] How to capture/handle error messages? E.g. `c.items(where="id=2")` logs "ERROR 1: "id" not recognised as an available field." to the console.
- [x] Should this also integrate with `fio cat`? (I think so, as this is effectivley ogr2ogr's [-where](https://gdal.org/programs/ogr2ogr.html#cmdoption-ogr2ogr-where) option)
- [x] <s>Should arguments like this be available to be passed from the `kwargs` from `fiona.open`? This would allow, e.g. `geopandas.from_file(..., where='id > 5000')` from 3rd party interfaces. This question applies to `bbox` too, so could be deferred to a later issue.</s> (See [geopandas/io/file.py](https://github.com/geopandas/geopandas/blob/63de9bd1820080eedc443c09df7686e1022304a0/geopandas/io/file.py#L48-L158) where `bbox` and `mask` are handled separately from `kwargs` -- this would need to be followed up on geopandas directly)
- [x] Does the API header need to be repeated in all `ogrext*.pxd` files? This isn't clear.
- [x] Add doc (where is the main docstrings?), and add example.
- [x] Add tests.

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -1212,7 +1212,7 @@ cdef class Iterator:
     cdef stepsign
 
     def __cinit__(self, collection, start=None, stop=None, step=None,
-                  bbox=None, mask=None):
+                  bbox=None, mask=None, where=None):
         if collection.session is None:
             raise ValueError("I/O operation on closed collection")
         self.collection = collection
@@ -1237,6 +1237,14 @@ cdef class Iterator:
 
         else:
             OGR_L_SetSpatialFilter(cogr_layer, NULL)
+
+        if where:
+            where_b = where.encode('utf-8')
+            where_c = where_b
+            OGR_L_SetAttributeFilter(cogr_layer, <const char*>where_c)
+
+        else:
+            OGR_L_SetAttributeFilter(cogr_layer, NULL)
 
         self.encoding = session._get_internal_encoding()
 
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -1241,7 +1241,8 @@ cdef class Iterator:
         if where:
             where_b = where.encode('utf-8')
             where_c = where_b
-            OGR_L_SetAttributeFilter(cogr_layer, <const char*>where_c)
+            exc_wrap_int(
+                OGR_L_SetAttributeFilter(cogr_layer, <const char*>where_c))
 
         else:
             OGR_L_SetAttributeFilter(cogr_layer, NULL)
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -20,16 +20,17 @@ from fiona._geometry cimport (
     GeomBuilder, OGRGeomBuilder, geometry_type_code,
     normalize_geometry_type_code, base_geometry_type_code)
 from fiona._err cimport exc_wrap_int, exc_wrap_pointer, exc_wrap_vsilfile
-
 import fiona
 from fiona._env import GDALVersion, get_gdal_version_num
-from fiona._err import cpl_errs, FionaNullPointerError, CPLE_BaseError, CPLE_OpenFailedError
+from fiona._err import (
+    cpl_errs, FionaNullPointerError, CPLE_BaseError, CPLE_AppDefinedError,
+    CPLE_OpenFailedError)
 from fiona._geometry import GEOMETRY_TYPES
 from fiona import compat
 from fiona.errors import (
     DriverError, DriverIOError, SchemaError, CRSError, FionaValueError,
     TransactionError, GeometryTypeValidationError, DatasetDeleteError,
-    FionaDeprecationWarning)
+    AttributeFilterError, FionaDeprecationWarning)
 from fiona.compat import strencode
 from fiona.rfc3339 import parse_date, parse_datetime, parse_time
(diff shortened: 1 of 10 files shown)


mwtoews:
Some progress with error capture/handling and `fio cat` has been completed. 

One question is whether to throw a different exception than `CPLE_AppDefinedError` if there is an issue handling the WHERE clause. This currently looks a bit harsh, when e.g.:
```
$ fio cat --where NAME=3 coutwildrnp.shp
ERROR:fiona._env:Type mismatch or improper type of arguments to = operator.
ERROR:fiona.fio.cat:Exception caught during processing
  File "/tmp/venv/lib/python3.6/site-packages/fiona/fio/cat.py", line 77, in cat
    for i, feat in src.items(bbox=bbox, where=where):
  File "/tmp/venv/lib/python3.6/site-packages/fiona/collection.py", line 281, in items
    self, start, stop, step, bbox, mask, where)
  File "fiona/ogrext.pyx", line 1244, in fiona.ogrext.Iterator.__cinit__
  File "fiona/_err.pyx", line 246, in fiona._err.exc_wrap_int
fiona._err.CPLE_AppDefinedError: Type mismatch or improper type of arguments to = operator.
Aborted!
```
should these OGR errors be re-thrown differently?

fiona/ogrext.pyx
```diff
@@ -1238,6 +1238,15 @@ cdef class Iterator:
         else:
             OGR_L_SetSpatialFilter(cogr_layer, NULL)
 
+        if where:
+            where_b = where.encode('utf-8')
+            where_c = where_b
+            exc_wrap_int(
+                OGR_L_SetAttributeFilter(cogr_layer, <const char*>where_c))
```
sgillies, line 1245: @mwtoews let's catch that CPLEx exception here and raise something like fiona.errors.AttributeFilterError. In general, I'd like this project improve in the direction of only intentionally raising its own errors, with Python or GDAL errors being a signal of buggy code.

fiona/ogrext3.pxd
```diff
@@ -322,6 +322,7 @@ cdef extern from "ogr_api.h":
     void *  OGROpenShared (char *path, int mode, void *x)
     int     OGRReleaseDataSource (void *datasource)
     OGRErr  OGR_L_SetIgnoredFields (void *layer, const char **papszFields)
+    OGRErr  OGR_L_SetAttributeFilter(void *layer, const char*)
```
sgillies, line 325: Thanks for doing this in all the .pxd files. In an upcoming PR we're going to delete ogrext1.pxd and ogrext2.pxd as we'll be dropping support for GDAL 1.x and 2.x.

mwtoews:
@sgillies thanks for clarification. A new exception is now available, and the earlier `fio cat` example shows a new output:
```
$ fio cat --where NAME=3 coutwildrnp.shp
ERROR:fiona._env:Type mismatch or improper type of arguments to = operator.
Usage: fio cat [OPTIONS] INPUTS...

Error: Invalid value: 'where' clause is invalid: Type mismatch or improper type of arguments to = operator.
$ fio cat --where NAM=3 coutwildrnp.shp
ERROR:fiona._env:"NAM" not recognised as an available field.
Usage: fio cat [OPTIONS] INPUTS...

Error: Invalid value: 'where' clause is invalid: "NAM" not recognised as an available field.
```

fiona/ogrext.pyx
```diff
@@ -1238,6 +1239,18 @@ cdef class Iterator:
...
             OGR_L_SetSpatialFilter(cogr_layer, NULL)
 
+        if where:
+            where_b = where.encode('utf-8')
+            where_c = where_b
+            try:
+                exc_wrap_int(
+                    OGR_L_SetAttributeFilter(cogr_layer, <const char*>where_c))
+            except CPLE_AppDefinedError as e:
+                raise AttributeFilterError(e) from None
```
mwtoews, line 1249: a wee note here, as this is a new Python feature to me [documented here](https://docs.python.org/3/tutorial/errors.html#exception-chaining) and [shown here](https://stackoverflow.com/q/52725278/). It shortens the repetition in the exception trace, without "During handling of the above exception...", but importantly it keeps the original exception message from OGR telling the user what's wrong.

sgillies: 😲

mwtoews:
@sgillies would it be problematic if I were to backport this to maint-1.9? I'd need to adjust the change log for master too, as it would be switched to the 1.9 milestone.

sgillies:
@mwtoews fine with me!

PR #962, opened 2020-10-07
Use fiona.meta.extensions for tests

rbuffat:
- Use fiona.meta.extensions in conftest
- Refactor test_bounds to use testdata_generator


sgillies:
There I go again. Somehow, the changed number of return values from the testdata generator slipped by git.

PR #963, opened 2020-10-08
Travis-ci: added support for ppc64le along with amd64

sanjaymsh:
Hi,
I have added support for ppc64le build on travis-ci in the branch . The travis-ci build log can be tracked on the link :https://travis-ci.com/github/sanjaymsh/Fiona/builds/188878827 . I believe it is ready for the final review and merge.
Please have a look on it and if everything looks fine for you then please approve it for merge.

Thanks !!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,14 +41,6 @@ matrix:
         GDALVERSION="2.3.3"
         PROJVERSION="4.9.3"
     - os: linux
-      arch: ppc64le
-      dist: xenial
-      python: "3.6"
-      env:
-        GDALVERSION="2.3.3"
-        PROJVERSION="4.9.3"
-
-    - os: linux
       arch: amd64
       dist: xenial
       python: "3.6"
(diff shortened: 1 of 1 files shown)


sgillies:
Hi @sanjaymsh can you give us some motivation for this PR? Do you have concerns that the standard linux builds do not address? Keep in mind that the fiona package is a fairly thin wrapper around GDAL. If the GDAL project isn't being tested, this may not be worth the effort.

sanjaymsh:
Hi @sgillies This package is included in the distros for ppc64le - we are trying to make sure that integration into the distros is pain free by making sure they have a clean bill of health from their development community. This simplifies the integration process for ppc64le based distros.

sgillies:
@sanjaymsh what would you think about trimming this to a single ppc64le job to keep the build time down? I don't think 3 jobs will catch any more architecture-specific problems than 1 job.

sanjaymsh:
@sgillies Ok, yes, this way the time taken will be down. This approach looks good to me as well. In case if it is required to test on other matrix for ppc64 then it can be changed. For now, we can use the one which uses python:3.8. I will make changes accordingly.

Thanks!!

sanjaymsh:
@sgillies I have updated the .travis.yml and removed ppc64le from matrix for all env , just kept one to make sure that it supports the ppc64le arch as well. And now it looks like all the issues are resolved and it is ready for the final review and merge. Please have a look on this.

Thanks !!

sgillies:
Closing since we're leaving Travis CI.

PR #964, opened 2020-10-11
Test merge maint 1.8 master

sgillies:


sgillies:
Tests pass on my macbook but it looks like I have some appveyor and travis work to do.

rbuffat:
@sgillies If it is ok for you, I should have some time to look at it tomorrow.

Issue #194, opened 2015-01-06
RuntimeWarning when using list on collection


bekozi:
I am getting ``RuntimeWarning``s about a potentially altered cursor when calling ``list`` on a freshly opened collection object. I put together a test case recreating the message: https://gist.github.com/bekozi/2005f589fe8a87233543#file-test_read_cursor_warning_fiona-py-L45. There may be a very good reason for this, but I can’t sort out why...

rbuffat:
I implemented this feature a while ago. 

The reason for this warning is that the underlying ogr api does not guarantee for certain functions that the internal read cursor of the layer is modified. 

```
    records = []
    for f in source:
        records.append(f)
```

does not issue a runtime warning, while 

```
    records = list(source)
```

issues a runtime warning while list() executes the get_length function of the session, that calls the ogr function OGR_L_GetFeatureCount: http://www.gdal.org/ogr__api_8h.html#abfbba8f45edd5dc44ced91f9228f9124 (Note that some implementations of this method may alter the read cursor of the layer.)

I don't know how many of the drivers (currently or in the future) supported by fiona are in danger of this behavior. 

@sgillies
I could implement a patch that stores the index and only issues a warning if the read cursor actually was moved.
Probably this feature causes more confusion that it helps, thus it might also be removed. Or a test suite should be created to test the behavior of the OGR api for all supported drivers by fiona to check if one of them modifies the read cursor.

sgillies:
@rbuffat let's do remove the warning.

I'm not quite sure what the warning in the docs about read cursor alteration means. There are formats for which OGR_L_GetFeatureCount must read entirely through the file to get the count (GeoJSON for example) and the cursor would be set back to the beginning afterwards. Is this what the docs warn about or is there something else?

rbuffat:
@sgillies 
I interpret it the same way. From the documentation the following methods are affected: OGR_L_GetExtent
OGR_L_GetExtentEx
OGR_L_GetFeatureCount

The only thing that bothers me is if a driver modifies the read cursor this will lead to bugs that will be very nasty to find because from fionas documentation you will not be able to see that this is allowed behavior according to the OGR API. Then again, the need to get count or extent during an iteration is not that likely to happen.

sgillies:
[...]

PR #965, opened 2020-10-12
Fix* travis for "Test merge maint 1.8 master" branch

rbuffat:
@sgillies I did not spot what exactly caused that gdal install did not find the proj libs. I used the travis setup from maint-1.8 and tried to apply the same changes. I hope I caught everything. I

I moved the OSX builds to a github workflow. If you don't like this feel free to delete the  .github/workflows/ci_macos.yml. The advantage of using the github workflow is that it allows specifying the Python version and more jobs can run in parallel.

There are still some issues:
- An appveyor test is still failing
- Pytest should not display runtime warnings, such as `RuntimeWarning: Sequential read of iterator was interrupted. Resetting iterator. This can negatively impact the performance.`

The build of gdal 3.1.3 fails for ppc64le, but it succeeds with gdal 3.0.4.

### DIFF
--- a/.github/workflows/ci_macos.yml
+++ b/.github/workflows/ci_macos.yml
@@ -67,7 +67,6 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install -r requirements-dev.txt
-          python -m pip install coveralls>=1.1 --upgrade
 
       - name: Build PROJ
         run: |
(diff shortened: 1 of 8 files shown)


StefanBrand:
The `RuntimeWarning` does not appear in 1.8.17. I think this is a regression introduced in 1.8.18. Related SO question: https://stackoverflow.com/q/64995369/9778755

StefanBrand:
This issue sounds familiar: https://github.com/Toblerity/Fiona/issues/194

rbuffat:
Travis is now obsolete

Issue #968, opened 2020-10-16
GML Version 3.1.1 Export


rebot:
Hi,

I'm trying to make a GML export of a Polygon using Fiona. I've managed to make the export, but it seems like the export is in version 2 format of GML. How can I enable Fiona to make an export to GML Version 3.1.1. Should I add a new kind of schema? Is this related to my GDAL version? 

This issue is related to a question I've asked on another [project](https://github.com/DOV-Vlaanderen/pydov/issues/132#issuecomment-709186850)

## Steps to reproduce the problem.

I make use of GeoPandas, Fiona and Shapely. I've installed Fiona and GDAL using the wheel provided on the following website.

GDAL Version 3.1.3

```txt
Fiona @ file:///C:/Users/BEGILT/Development/Python/RockWorks/Fiona-1.8.17-cp36-cp36m-win_amd64.whl
GDAL @ file:///C:/Users/BEGILT/Development/Python/RockWorks/GDAL-3.1.3-cp36-cp36m-win_amd64.whl
```

Next, I make a shapely Polygon object and use the `mapping` function to make sure Fiona can save it as a GML. 
I've written a function to convert the GML from v2 tot v3.1.1, but I hope it's possible without the function!

```python
# Schema = http://schemas.opengis.net/gml/ - Schema definitie voor de .gml
schema = {'properties': OrderedDict([]),'geometry': 'Polygon'}# Schema = http://schemas.opengis.net/gml/ - Schema definitie voor de .gml
schema = {'properties': OrderedDict([]),'geometry': 'Polygon'}

# https://gis.stackexchange.com/questions/52705/how-to-write-shapely-geometries-to-shapefiles
# https://geoscripting-wur.github.io/PythonVector/

def exporteer_vorm(segment):
    # Haal GML op schrijf de file weg in memory
    gml_bytes = BytesIO()
    gml_filename = os.path.join('hulpmiddelen','projectzone',f'{segment.segment}.gml')
    with fiona.open(gml_bytes, 'w', driver='GML', schema=schema, crs=from_epsg(31370)) as f:
        f.write({
            'geometry': mapping(segment.geometry),
            'properties': {}
        })
    gml_bytes.seek(0)
    # Omzetting .gml van versie 2 naar versie 3.1.1 dat bruikbaar is in pyDOV
    gml_content = gml_bytes.read().decode('utf-8')
    gml_content = gml_content.replace('Box','Envelope')
[...]

PR #969, opened 2020-10-17
Pass dataset creation options to gdal_create when creating new file

rbuffat:
Currently, Fiona does only pass options when a new layer is created, but not when a dataset is created. 
This PR passes options to gdal_create when a new dataset is created.

I'm not sure if in this case gdal_create should be wrapped with `with Env(GDAL_VALIDATE_CREATION_OPTIONS="NO")`.

This fixes https://github.com/Toblerity/Fiona/issues/968

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -1007,7 +1007,7 @@ cdef class WritingSession(Session):
 
             if not CPLCheckForFile(path_c, NULL):
                 log.debug("File doesn't exist. Creating a new one...")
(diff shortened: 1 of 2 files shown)


rbuffat:
There seems to be an intermittent segfault with gdal 1.11. I'm not sure why. I deactivated the test for gdal 1.x.

sgillies (approved):
@rbuffat thank you! Until we come up with a way to separate the dataset and layer creation options (if we do), disabling validation seems like a good approach.

Issue #971, opened 2020-10-19
Parameter precision not working for transform_geom()


SaCodematix:
I do get the error `local variable 'new_coords' referenced before assignment` when I call `transform.transform_geom()` with an indicated value for the optional `precision` parameter other than the default value of `-1`.

## Expected behavior and actual behavior.
I expected to be able to transform a GeometryCollection with the precision of 3 decimal places through the optional parameter `precision` of `transform.transform_geom()`. 

## Steps to reproduce the problem.
```
from fiona import transform

source_crs = "epsg:25832"
dest_src = "epsg:4326"
geom =  {'type': 'GeometryCollection', 'geometries': [{'type': 'LineString', 'coordinates': [(512381.8870945257, 5866313.311218272), (512371.23869999964, 5866322.282500001), (512364.6014999999, 5866328.260199999)]}]}

geom_transformed = transform.transform_geom(source_crs, dest_src, geom)
print("Transformed geom without optional parameter 'precision':", geom_transformed, "\n")

geom_transformed = transform.transform_geom(source_crs, dest_src, geom, precision=-1)
print("Transformed geom with optional default parameter 'precision=-1':", geom_transformed, "\n")

geom_transformed = transform.transform_geom(source_crs, dest_src, geom, precision=3)
print("Transformed geom with optional parameter 'precision=3':", geom_transformed)
```

Output:
```
# python3 transform_geom_test.py
Transformed geom without optional parameter 'precision': {'type': 'GeometryCollection', 'geometries': [{'type': 'LineString', 'coordinates': [(9.184271907836777, 52.94630492951154), (9.184113777710168, 52.94638582212857), (9.184015228795353, 52.9464397114269)]}]} 

Transformed geom with optional default parameter 'precision=-1': {'type': 'GeometryCollection', 'geometries': [{'type': 'LineString', 'coordinates': [(9.184271907836777, 52.94630492951154), (9.184113777710168, 52.94638582212857), (9.184015228795353, 52.9464397114269)]}]} 

Traceback (most recent call last):
  File "transform_geom_test.py", line 16, in <module>
    geom_transformed = transform.transform_geom(source_crs, dest_src, geom, precision=3)
  File "/home/sarnold/anaconda3/envs/wind/lib/python3.8/site-packages/fiona/transform.py", line 92, in transform_geom
    return _transform_geom(
  File "fiona/_transform.pyx", line 220, in fiona._transform._transform_geom
[...]

PR #972, opened 2020-10-19
Support precision for GeometryCollection geometries in fiona.transform

rbuffat:
Currently, fiona.transform does not support GeometryCollection geometries when the precision argument is set (see https://github.com/Toblerity/Fiona/issues/971). 

This PR refactors the way geometries are rounded after the transformation and supports GeometryCollection.
@sgillies I'm not sure if this is the best way to solve this issue. The code looks a bit complicated.


sgillies (approved):
Excellent!

PR #979, opened 2020-11-19
ENH: Add WKT format support

### LINKED ISSUE #977: WKT2 support
Related to:
- https://github.com/OSGeo/PROJ/issues/2426#issuecomment-727937449
- https://github.com/geopandas/geopandas/issues/1697

Currently error is produced here:
```python
import fiona
schema = {"geometry": "Point", "properties": {"zero": "int"}}
feature = {
    "geometry": {"type": "Point", "coordinates": (0, 0)},
    "properties": {"zero": "0"},
}
target_crs = 'PROJCS["IaRCS_04_Sioux_City-Iowa_Falls_NAD_1983_2011_LCC_US_Feet",GEOGCS["GCS_NAD_1983_2011",DATUM["D_NAD_1983_2011",SPHEROID["GRS_1980",6378137.0,298.257222101]],PRIMEM["Greenwich",0.0],UNIT["Degree",0.0174532925199433]],PROJECTION["Lambert_Conformal_Conic"],PARAMETER["False_Easting",14500000.0],PARAMETER["False_Northing",8600000.0],PARAMETER["Central_Meridian",-94.83333333333333],PARAMETER["Standard_Parallel_1",42.53333333333333],PARAMETER["Standard_Parallel_2",42.53333333333333],PARAMETER["Scale_Factor",1.000045],PARAMETER["Latitude_Of_Origin",42.53333333333333],UNIT["Foot_US",0.3048006096012192]]'
with fiona.open("./test.shp", "w", driver="ESRI Shapefile", schema=schema, crs=target_crs) as collection:
    collection.write(feature)
```

Solution: use WKT2. Need to do some research on how to do that.

snowman2:
Closes #977

### DIFF
--- a/fiona/__init__.py
+++ b/fiona/__init__.py
@@ -202,6 +202,16 @@ def open(fp, mode='r', driver=None, schema=None, crs=None, encoding=None,
     crs_wkt : str
         An optional WKT representation of a coordinate reference
         system.
+    ignore_fields : list
+        List of field names to ignore on load.
+    ignore_geometry : bool
+        Ignore the geometry on load.
+    include_fields : list
+        List of a subset of field names to include on load.
+    wkt_version : str
+        WKT Version to use to for the CRS
+        (WKT1, WKT1_GDAL, WKT1_ESRI, WKT2, WKT2_2015, WKT2_2018).
+        Defaults to GDAL's default (WKT1_GDAL for GDAL 3).
     kwargs : mapping
         Other driver-specific parameters that will be interpreted by
         the OGR library as layer creation or opening options.
--- a/fiona/_crs.pyx
+++ b/fiona/_crs.pyx
@@ -8,7 +8,7 @@ from __future__ import absolute_import
 import logging
 
 from fiona cimport _cpl
-from fiona._err cimport exc_wrap_pointer
+from fiona._err cimport exc_wrap_pointer, exc_wrap_ogrerr
 from fiona._err import CPLE_BaseError
 from fiona._shim cimport osr_get_name, osr_set_traditional_axis_mapping_strategy
 from fiona.compat import DICT_TYPES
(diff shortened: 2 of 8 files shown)


mwtoews:
@snowman2 Fiona 1.9 has been updated to require GDAL >=3.1 (see b4dcc4001e91d6cbd9530471b0de599f5b83bd48), so a few simplifications could be made throughout this PR. I can review this when ready.

fiona/__init__.py
```diff
@@ -202,6 +202,16 @@ def open(fp, mode='r', driver=None, schema=None, crs=None, encoding=None,
     crs_wkt : str
         An optional WKT representation of a coordinate reference
         system.
+    ignore_fields : list
+        List of field names to ignore on load.
+    ignore_geometry : bool
+        Ignore the geometry on load.
+    include_fields : list
```
snowman2, line 209: These appear to be unrelated to this MR, but were missing from the docstring.

fiona/_crs.pyx
```diff
@@ -20,8 +20,38 @@ logger = logging.getLogger(__name__)
...
+        OSRExportToWktEx(cogr_srs, &wkt_c, options_wkt)
+
+    if wkt_c != NULL:
+        wkt_b = wkt_c
+        wkt = wkt_b.decode('utf-8')
+
+    _cpl.CPLFree(wkt_c)
+
+    if not wkt:
+        raise CRSError("Invalid input to create CRS: {}".format(crs))
```
mwtoews, line 50: This error should somehow also describe `wkt_version`, e.g. `fiona._crs.crs_to_wkt(target_crs, wkt_version="abc")` will raise an error that makes no mention of a potentially bad `wkt_version` value.

Not sure if it's possible to capture possible invalid `wkt_version` values earlier, or if that's a bad idea, as other values work, e.g. `wkt_version="WKT2"`.

fiona/__init__.py
```diff
@@ -202,6 +202,16 @@ def open(fp, mode='r', driver=None, schema=None, crs=None, encoding=None,
...
+        List of a subset of field names to include on load.
+    wkt_version : str
+        WKT Version to use to for the CRS.
+        (WKT1_GDAL, WKT1_ESRI, WKT2_2015, WKT2_2018)
+        Only works with GDAL 3+. Default is WKT1_GDAL.
```
mwtoews, line 214: Version GDAL 3+ is now irrelevant.
```suggestion
        Default is WKT1_GDAL.
```
Note the implementation here has a default of None, where FORMAT is not passed as an option. Does GDAL assume the default WKT1_GDAL? ([ref](https://gdal.org/api/ogrspatialref.html#_CPPv4NK19OGRSpatialReference11exportToWktEPPcPPCKc))

fiona/__init__.py
mwtoews: Figured so!

fiona/__init__.py
snowman2: It is the default until GDAL 4 (https://github.com/OSGeo/gdal/issues/4681).

snowman2: I updated it to hopefully be more clear.

fiona/_crs.pyx
mwtoews: there are a few ways to capture an error message, like this https://github.com/Toblerity/Fiona/blob/9c8fe736d4381526905c6976c399a33d10e3ecbf/fiona/ogrext.pyx#L1512-L1516

snowman2: Thanks!

Current error message:
```
fiona.errors.CRSError: Invalid input to create CRS: EPSG:4326. Please choose a different WKT version (WKT2 is recommended): WKT1 | WKT1_GDAL | WKT1_ESRI | WKT2 | WKT2_2018 | WKT2_2015. GDAL ERROR: Unsupported value for FORMAT
```

mwtoews: Useful error capture, as sometimes `WKT1_GDAL` will be specified, but the error message would contain (e.g.) "GDAL ERROR: PROJ: proj_as_wkt: Unsupported conversion method: Lambert Conic Conformal (2SP Michigan)" as the chosen wkt format is incompatible.

mwtoews (approved):
Thanks, looks great! Last request is to add an entry to CHANGES.txt (in the "New features" section), then I'll merge it after a few days (in case there is any further review by others).

mwtoews:
Thanks for the time an patience putting this together! The last CI failure is unrelated and is fixed with 7fa6bcceb3f383b41c44983d3c534b367c5f64a7

snowman2:
Thanks @mwtoews 👍

fiona/__init__.py
```diff
@@ -202,6 +202,16 @@ def open(fp, mode='r', driver=None, schema=None, crs=None, encoding=None,
...
         system.
+    ignore_fields : list
+        List of field names to ignore on load.
+    ignore_geometry : bool
+        Ignore the geometry on load.
+    include_fields : list
+        List of a subset of field names to include on load.
+    wkt_version : str
+        WKT Version to use to for the CRS
+        (WKT1, WKT1_GDAL, WKT1_ESRI, WKT2, WKT2_2015, WKT2_2018).
```
rouault, line 213: for the Fiona API, I would suggest using WKT2_2019 which is the canonical name and remap that to WKT2_2018 when passing it to GDAL for compatibility with early GDAL 3.1 (WKT2_2019 was added in GDAL 3.2)

PR #980, opened 2020-11-20
Switch from travis to github workflows

rbuffat:
This PR replaces travis with a github workflow

It includes the same features compared to travis except testing the GDAL master branch (I did not figure out how to mark a job with allow_failure). But as the GDAL master job is typically not cached and thus the most time/resource consuming, it would probably be removed from travis anyway.

The action is designed to emulate a travis environment. Thus, no changes to the proj/gdal install script were necessary. Also, this would allow to easily switch back to travis.

~~The github workflows only work once this PR is merged.~~

The workflow in action can be seen here: https://github.com/rbuffat/Fiona/actions/runs/374781042


jorisvandenbossche:
> I did not figure out how to mark a job with allow_failure

See eg here how we did this for pygeos: https://github.com/pygeos/pygeos/blob/62ddf67cf623c6f66da3cc9b3f941763b810a664/.github/workflows/test-linux.yml#L90

rbuffat:
@jorisvandenbossche Thanks for the template, this helped a lot!

rbuffat:
@sgillies The github action seem to still work. One of the limitations of github actions is that the cache is only available for one week:

> A repository can have up to 5GB of caches. Once the 5GB limit is reached, older caches will be evicted based on when the cache was last accessed. Caches that are not accessed within the last week will also be evicted.

https://github.com/actions/cache

An alternative to caching gdal would be to use docker container. Gdal seems to provide a range of different docker images already: https://github.com/OSGeo/gdal/tree/master/gdal/docker It should be possible to create tailored containers for our CI based on these containers. But for now, it is probably better to have a working CI as soon as possible.

sgillies:
@rbuffat thank you!

PR #981, opened 2020-11-21
Switch windows ci to a github workflow

rbuffat:
Appveyor allows only one parallel job. Typically, the windows ci is thus quite slow compared to the rest. This PR switches the appveyor CI to a github workflow. 

The only drawback is that currently no wheel is built and saved as artifacts.

Together with https://github.com/Toblerity/Fiona/pull/980 only one CI would be used. 

fio --version and fio --gdal-version currently do not load the windows dlls with Python 3.8. This is fixed with this PR, but other fio commands could still fail. Probably, we should increase the test coverage for the fio commands.


sgillies:
@rbuffat let's not do this for 1.8.x (which we will close out). How about for 1.9?

sgillies:
@rbuffat this one is out of date. I'm going to close it and we'll do another for 1.9.

PR #985, opened 2020-11-30
Remove 'six' module use; reduce footprint of 'fiona.compat'

mwtoews:
While `six` was handy transitioning thru the Python 2 to 3 conversion, now that Fiona is fully Python 3, it is no longer needed.

Also, there are fewer instances where `fiona.compat` is needed.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -15,6 +15,7 @@ Changes:
   fiona.model.Properties.
 - GDAL 3.1.0 is the minimum GDAL version.
 - Drop Python 2, and establish Python 3.7 as the minimum version (#1079).
+- Remove six and reduce footprint of fiona.compat (#985).
 
 Deprecations:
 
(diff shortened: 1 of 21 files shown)


rbuffat:
It's certainly a good idea to get rid of the Python 2.x past. However, what I'm a bit concerned about is that at one point maint-1.9 needs to be merged into master, and the more this branch deviates from maint-1.9, the more merge conflicts could arise.

mwtoews:
@rbuffat ah, I see the concern. I normally contribute features to `master`, but I see features are also in `maint-1.9` too (I normally expect "maint-*" branches to be maintenance branches with bug fixes.) Let me know if there is something that I can help out with.

sgillies:
@mwtoews I'm so sorry to let this get old and for the unfortunate maint-1.8/1.9/master split. Would you be willing to retarget this at maint-1.9 (the next release)?

mwtoews:
This is ready to go for `maint-1.9`. The rstcheck failure is unrelated.

sgillies:
Thanks @mwtoews !

Issue #986, opened 2020-12-02
RuntimeWarning: Sequential read of iterator was interrupted. Resetting iterator. This can negatively impact the performance.


rbuffat:
The `RuntimeWarning` does not appear in 1.8.17. I think this is a regression introduced in 1.8.18. Related SO question: https://stackoverflow.com/q/64995369/9778755

_Originally posted by @StefanBrand in https://github.com/Toblerity/Fiona/issues/965#issuecomment-737073718_

rbuffat:
@StefanBrand I believe this RuntimeWarning is not a regression. This behavior was introduced in https://github.com/Toblerity/Fiona/commit/e50b4ad8a31ecf65c56b95a9f05863c995039618 and the log message was later changed to a RuntimeWarning.

GDAL (the library Fiona uses to access the geodata) maintains a iterator over the features that are currently read. There a some operations that, for some drivers, can influence this iterator. Thus, after such operations we have to ensure that the iterator is set to the correct position that a continuous read of the data is ensured. Such operations include counting all features in a dataset, respectively calculating its extent.

There are different types of drivers in GDAL. Some drivers support random access, while some do not. For the drivers that do not support random access, the resetting of the iterator involves reading all features again up to the iterator position. As this is a possible costly operation, this RuntimeWarning is emitted, so that users are aware of this behavior.

It seems that this warning is not clear and we should probably improve the wording.

From https://stackoverflow.com/questions/64995369/geopandas-warning-on-read-file, the warning is shown when the bbox parameter is set in the read_file method of geopandas.

StefanBrand:
I'm simply reading a small GPKG and creating a `list` from it:

```
    with fiona.open('sample.gpkg') as src:
        sample = list(src)  # issues a RuntimeWarning in fiona 1.8.18
```

rbuffat:
So what's going on is that the list() method seems to call len() before it starts iterating over the Iterator. I assume this to create an efficient data structure for the list. Would be interesting to look at the implementation of list(). We probably should not emit a warning when the iterator was interrupted before an iteration has actually started. 

```
class Iterator:

    def __init__(self):
        print("__init__")
        self.counter = 0
        self.len = 5

    def __iter__(self):
        print("__iter__")
        return self

    def __next__(self):
        self.counter += 1
        if self.counter > self.len:
            raise StopIteration
        print("__next__", self.counter)
        return self.counter

    def __len__(self):
        print("__len__")
        return self.len

it = Iterator()
[...]

PR #987, opened 2020-12-02
Only warn iterator reset when iteration has started

rbuffat:
This PR fixes that a warning is raised in cases where a collection is converted to a list (https://github.com/Toblerity/Fiona/issues/986), such as:

```
    with fiona.open('sample.gpkg') as src:
        sample = list(src)  # issues a RuntimeWarning in fiona 1.8.18
```

However, it does not prevent that OGR_L_SetNextByIndex(start_index) is executed twice. When the start_index is 0, this is probably not a costly operation, but if the start_index is large, and the driver does not support random access, this could be costly. 

This could be solved by moving https://github.com/Toblerity/Fiona/blob/maint-1.9/fiona/ogrext.pyx#L1554-L1556 into _next() However, _next() is potentially executed many times for large datasets, thus this could generate some overhead for large datasets.

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -1467,7 +1467,7 @@ cdef class Iterator:
 
         # Set read cursor to next_item position
         if session.cursor_interrupted:
-            if not self.fastindex:
+            if not self.fastindex and not self.next_index == self.start:
                 warnings.warn("Sequential read of iterator was interrupted. Resetting iterator. "
                               "This can negatively impact the performance.", RuntimeWarning)
             exc_wrap_int(OGR_L_SetNextByIndex(session.cogr_layer, self.next_index))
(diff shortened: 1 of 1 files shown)


sgillies:
Though I'd rather not have another 1.8.x release, it seems like a 1.8.19 might be warranted.

rbuffat:
@sgillies sorry for that. The warning message seems to cause more confusion than it is helpful. Maybe it is better to remove it? 

Also, it looks as Travis seems to be not working anymore. Not sure why, could be just a hiccup.

sgillies:
@rbuffat the "toblerity" organization's open source Travis credits are all spent. I haven't had the time or energy to ask for more and maybe we should switch permanently to GitHub Actions or AppVeyor?

I think the warning is useful in some contexts, but can be confusing when it occurs during normal usage due to details of the GDAL/OGR driver.

Issue #992, opened 2020-12-15
Spatialite support


jamesmyatt:
I'm not clear about why Fiona doesn't support Spatialite files, especially since it does support GeoPackage file and OGR also supports Spatialite.

I assume that all SQLite files fall under the "RDBMS" category in https://fiona.readthedocs.io/en/latest/manual.html#rules-of-thumb, and Spatialite uses SQLite, but so does GeoPackage. Also https://github.com/Toblerity/Fiona/issues/121. This rule-of-thumb makes sense for a generic SQLite file where there are no standards for the tables and columns, but I think it's good for Fiona to support specific formats like Spatialite and GeoPackage, where SQLite is merely the container.

In particular, I am able to read Spatialite by patching `fiona.supported_drivers`. So the following code works exactly the same for both GeoPackage and Spatialite files (in fact, it should work the same for any multi-layer file supported by Fiona).

```
import fiona
import geopandas

# Force SQLite support(!?)
fiona.supported_drivers["SQLite"] = "r"

# Read layers
layers = fiona.listlayers(data_file)

# Load file using Fiona
dfs = {name: geopandas.read_file(data_file, layer=name) for name in layers}
```

Here's my "read_postgis" version of this code for Spatialite only and Windows only, as well as lots of other assumptions that I don't even know. I'm sure this can be improved, but it's the limit of my SQL and GIS knowledge/googling.

```
import contextlib
import sqlite3
import geopandas

dfs = {}
with contextlib.closing(sqlite3.connect(data_file)) as con:
    con.enable_load_extension(True)
    con.load_extension("mod_spatialite.dll")

    with contextlib.closing(con.cursor()) as c:
        layers = c.execute("SELECT f_table_name, f_geometry_column FROM geometry_columns").fetchall()

    for name, geom_col in layers:
        dfs[name] = pd.concat(
            [
                pd.read_sql_query(f"SELECT * FROM {name};", con).drop(
                    columns=["GEOMETRY", "OGC_FID"]
                ),
                geopandas.GeoDataFrame.from_postgis(
                    f"SELECT ST_AsBinary(GEOMETRY) as {geom_col} FROM {name};",
                    con,
                    geom_col=geom_col,
                ),
            ],
            axis=1,
        )
```

I think we can all agree that the `read_file` version of this code is far better, even if the `read_postgis` version is 3x faster.

I suppose, my main questions are:

- Am I way off base here?
- Have I misunderstood the parallels between GeoPackage and Spatialite?
[...]

PR #999, opened 2021-02-09
Ensable sqlite support

caspervdw:
This closes https://github.com/Toblerity/Fiona/issues/992. As Geopackage is a standard within the sqlite file format, the corresponding unittests are are copy of `test_geopackage.py`.

### DIFF
--- a/fiona/drvsupport.py
+++ b/fiona/drvsupport.py
@@ -115,6 +115,7 @@ supported_drivers = dict([
     ("SEGY", "r"),
     # Norwegian SOSI Standard 	SOSI 	No 	Yes 	No, needs FYBA library
     # SQLite/SpatiaLite 	SQLite 	Yes 	Yes 	No, needs libsqlite3 or libspatialite
+    ("SQLite", "raw"),
     # SUA 	SUA 	No 	Yes 	Yes
     ("SUA", "r"),
     # SVG 	SVG 	No 	Yes 	No, needs libexpat


caspervdw:
Test fails with GDAL 3.*, it seems unrelated:

```
Extracting archive: gdalbin.zip
ERROR: gdalbin.zip
gdalbin.zip
Open ERROR: Can not open the file as [zip] archive
ERRORS:
Is not archive
Can't open as archive: 1
Files: 0
Size:       0
Compressed: 0
Command exited with code 2
```

rbuffat:
@caspervdw Currently, the master branch is a bit behind maint-1.8. maint-1.8 contains automatic tests for all drivers enabled in [drvsupport.py](https://github.com/Toblerity/Fiona/blob/maint-1.8/fiona/drvsupport.py): https://github.com/Toblerity/Fiona/blob/maint-1.8/tests/test_drvsupport.py

sgillies:
@caspervdw as far as I am concerned, we don't need the tests. I am confident that all the sqlite-specific concerns are covered by the existing geopackage tests (geopackage driver also uses sqlite).

The more driver-specific tests we add, the higher the maintenance cost of this package, and the harder it will be to change code in the future.

caspervdw:
I changed base to `maint-1.8` and omitted the tests. Please review again!

sgillies:
There are some test failures I don't understand on Windows. Any ideas @rbuffat ?

rbuffat:
@sgillies  I guess the rebase did not trigger the maint-1.8 tests. I guess for maint-1.8 there need to be some date/time settings in https://github.com/Toblerity/Fiona/blob/maint-1.8/fiona/drvsupport.py necessary. I will look into this.

rbuffat:
It looks as the SQLite driver supports timezone for datetimes with GDAL 2.4.0 and never supports timezones for time.  Aslo there seems to be an issue with GDAL 2.0:

```
__________________________ test_write_memfile_crs_wkt __________________________

    @pytest.mark.xfail(raises=DriverError)
    def test_write_memfile_crs_wkt():
        example_schema = {
            "geometry": "Point",
            "properties": [("title", "str")],
        }
    
        example_features = [
            {
                "geometry": {"type": "Point", "coordinates": [0.0, 0.0]},
                "properties": {"title": "One"},
            },
            {
                "geometry": {"type": "Point", "coordinates": [1.0, 2.0]},
                "properties": {"title": "Two"},
            },
            {
                "geometry": {"type": "Point", "coordinates": [3.0, 4.0]},
                "properties": {"title": "Three"},
            },
        ]
    
        with io.BytesIO() as fd:
            with fiona.open(
                fd,
                "w",
                driver="GPKG",
                schema=example_schema,
                crs_wkt=crs_to_wkt("EPSG:32611"),
            ) as dst:
                dst.writerecords(example_features)
    
            fd.seek(0)
            with fiona.open(fd) as src:
>               assert src.crs == {"init": "epsg:32611"}
E               AssertionError: assert {} == {'init': 'epsg:32611'}
E                 Right contains 1 more item:
E                 {'init': 'epsg:32611'}
E                 Use -v to get the full diff
```

I would suggest to drop the test for GDAL 2.0. GDAL 2.1.0 was released 2016-05-02.

caspervdw:
Work is continuing on #1002, thanks @rbuffat for picking this up

PR #1001, opened 2021-02-14
WIP: Use docker container in Linux CI

rbuffat:
**Problem:** 

Building PROJ/GDAL takes a lot of time and Github CI invalidates caches after a week.

**Solution:** 

Use a custom docker container for the CI that includes required GDAL binaries already.
*  Docker containers are built in a custom project and uploaded to the GitHub Container Registry: https://github.com/rbuffat/buildcontainer
* In Fiona's CI these docker containers are used to build and test Fiona. 

**Pro:**
* Docker container can be reused (e.g. for the rasterio CI)
* As GDAL is only compiled once it is not necessary to use a stripped-down version of GDAL and more drivers can be included
* The docker container could be shared with the wheel build setup, allowing faster wheel builds as well as testing Fiona with the GDAL included in the wheels

**Cons:**
* Increased complexity. GitHub Container Registry is currently only a feature preview and availability and pricing can change at any time
* Knowledge about docker is required to maintain the CI
(I only recently started using docker, thus there could be more things I'm not yet aware)

@sgillies I was curious how difficult it was to get this working. For me, it would be perfectly fine to keep the existing CI pipeline and keep this archived in case in the future it gets more urgent. Personally, I think the sharing of containers for wheel building and CI could be interesting, but this would definitely take some more work.


sgillies:
@rbuffat what would you think about dropping this and switching to the solution from the rasterio repo (using official GDAL images)?

sgillies:
Closing.

PR #1002, opened 2021-02-17
Enable SQLite driver

rbuffat:
See https://github.com/Toblerity/Fiona/pull/999

### DIFF
--- a/fiona/drvsupport.py
+++ b/fiona/drvsupport.py
@@ -120,6 +120,7 @@ supported_drivers = dict([
     ("SEGY", "r"),
     # Norwegian SOSI Standard 	SOSI 	No 	Yes 	No, needs FYBA library
     # SQLite/SpatiaLite 	SQLite 	Yes 	Yes 	No, needs libsqlite3 or libspatialite
+    ("SQLite", "raw"),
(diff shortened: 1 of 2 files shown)


caspervdw:
It seems that GDAL 2.0 picks SQLite as the preferred driver for opening a GPKG file, thereby losing CRS WKT support.

Probably the easiest way to fix it is adding a conditional skip `pytest.mark.skipif` for this old GDAL version.

jamesmyatt:
Closes #992

jamesmyatt:
> It seems that GDAL 2.0 picks SQLite as the preferred driver for opening a GPKG file, thereby losing CRS WKT support.
> 
> Probably the easiest way to fix it is adding a conditional skip `pytest.mark.skipif` for this old GDAL version.

> It seems that GDAL 2.0 picks SQLite as the preferred driver for opening a GPKG file, thereby losing CRS WKT support.
> 
> Probably the easiest way to fix it is adding a conditional skip `pytest.mark.skipif` for this old GDAL version.

@caspervdw , is it better to drop support for GDAL 2.0, which was release in Jun 2015? GDAL 2.1 was released in Apr 2016 and 5 years seems like a very generous amount of support time for an open-source project.

PR #1005, opened 2021-03-02
CI: convert Travis CI to GitHub for Linux only (for now)

mwtoews:
This aims to convert most of the Travis CI testing over to Github Actions.

The install scripts have more-or-less been rewritten to be CI-agnostic, and share a common install prefix.

Currently tests on macOS have been left out, for now. If this is a big priority, they could be added either as a second `macos.yml` file, or perhaps the main file could be renamed `ci.yml` with some additions to the matrix for `os`.


rbuffat:
@mwtoews I hope you were aware of https://github.com/Toblerity/Fiona/pull/980

mwtoews:
no, I wasn't. Probably should have checked first!

PR #1008, opened 2021-04-20
Explicitly require setuptools, fiona/fio/main.py and tests/test_fio_info.py imports pkg_resources

hrnciar:


jorisvandenbossche:
Instead of adding setuptools to the `install_requires`, it might be the better option to remove it's usage in fiona/fio/main.py. The setuptools docs also say it is discouraged in favor of stdlib `importlib` (https://setuptools.readthedocs.io/en/latest/pkg_resources.html), or otherwise there are also dedicated alternatives like `entrypoints` (https://entrypoints.readthedocs.io/en/latest/index.html).

hrnciar:
Yes, that's certainly also another option for how to do this. The reason why I opened this PR is that in Fedora we introduced this change (https://fedoraproject.org/w/index.php?title=Changes/Reduce_dependencies_on_python3-setuptools). We would like packages to explicitly state that they require setuptools on runtime. We identified about 200 packages with missing `setuptools` in `install_requires` and we tried to help upstream to put it there to avoid any future breakages in Fedora. I am sorry, but I cannot dedicate more time to do it the way you propose.

jorisvandenbossche:
> I am sorry, but I cannot dedicate more time to do it the way you propose.

No problem, and thanks a lot for bringing it up, as it's currently indeed a missing requirement

sgillies:
Let's require setuptools for 1.8.20 (see #1014) and switch to importlib in the future.

PR #1019, opened 2021-06-21
BUG: Remove unreachable return in ogrext

snowman2:
Saw a warning when compiling about this being unreachable.


mwtoews (approved):
This warning was indeed still there, thanks @snowman2 !

snowman2:
Thanks @mwtoews 👍

PR #1021, opened 2021-06-21
REF: Use compile_time_env in place of shims

snowman2:
Same thing as: https://github.com/mapbox/rasterio/pull/2016

This assumes GDAL 2.3+, same as the rasterio PR.


snowman2:
This will help with #979

sgillies:
@snowman2 can we target this for 1.9?

snowman2:
> can we target this for 1.9?

Hopefully it won't be too messy to do after the merge from 1.8 :crossed_fingers:

PR #1046, opened 2021-11-21
Add support for Python 3.10

mwtoews:
The main fix is to upgrade cython to a version that supports Python 3.10. Package classifier metadata is also updated.

Also add recent versions of GDAL and PROJ in the CI matrix.

Appyveyor (Windows) is not updated here, as it appears to have generally lagged without Python 3.9 support yet.

Closes #1043

PR #1049, opened 2021-11-30
Add support for Python 3.10

akrherz:
An attempt to bring Python 3.10 support to maint-1.8 branch?  I paired back the original commit some out of my blissful ignorance of what may not be necessary.  Please feel free to forcibly reject the PR with fire!

In my fork, the python3.6 build failure matches current [maint-1.8 branch build failure](https://github.com/Toblerity/Fiona/runs/2713722152?check_suite_focus=true)

```
def test_write_memfile_crs_wkt():
E               AssertionError: assert 'SQLite' == 'GPKG'
E                 - SQLite
E                 + GPKG
```

To get python 3.10 green in CI, I needed to additionally:
-  add a setup.py shim since `enum34` was not available.
- bump pytest and boto3 to a version that works on python 3.10

The please-with-a-cherry-on-top, would be to then cut a 1.8.21 release :)

### DIFF
--- a/fiona/drvsupport.py
+++ b/fiona/drvsupport.py
@@ -99,7 +99,7 @@ supported_drivers = dict([
     # multi-layer
     #   ("OpenAir", "r"),
     # PCI Geomatics Database File 	PCIDSK 	No 	No 	Yes, using internal PCIDSK SDK (from GDAL 1.7.0)
-    ("PCIDSK", "rw"),
+    ("PCIDSK", "raw"),
     # PDS 	PDS 	No 	Yes 	Yes
     ("PDS", "r"),
     # PDS renamed to OGR_PDS for GDAL 2.x
@@ -160,6 +160,7 @@ driver_mode_mingdal = {
           'FlatGeobuf': (3, 1, 3)},
 
     'a': {'GPKG': (1, 11, 0),
+          'PCIDSK': (3, 3, 0),
           'GeoJSON': (2, 1, 0),
           'MapInfo File': (2, 0, 0)}
 }
--- a/setup.py
+++ b/setup.py
@@ -352,7 +352,11 @@ setup_args = dict(
         'License :: OSI Approved :: BSD License',
         'Operating System :: OS Independent',
         'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.8',
+        'Programming Language :: Python :: 3.9',
+        'Programming Language :: Python :: 3.10',
         'Topic :: Scientific/Engineering :: GIS'])
 
 if os.environ.get('PACKAGE_DATA'):
--- a/setup.py
+++ b/setup.py
@@ -295,6 +295,9 @@ requirements = [
     'ordereddict; python_version < "2.7"',
     'enum34; python_version < "3.4"'
 ]
+# Python 3.10 workaround as enum34 not available
+if sys.version_info >= (3, 10):
+    requirements.remove('enum34; python_version < "3.4"')
 
 extras_require = {
     'calc': ['shapely'],
--- a/tests/test_open.py
+++ b/tests/test_open.py
@@ -9,6 +9,8 @@ import fiona
 from fiona._crs import crs_to_wkt
 from fiona.errors import DriverError
 
+from .conftest import requires_gdal21
+
 
 def test_open_shp(path_coutwildrnp_shp):
     """Open a shapefile"""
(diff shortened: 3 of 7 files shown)


mwtoews:
See #1037 for `test_write_memfile_crs_wkt`; I've just pushed that here to see if that clears the status...

mwtoews (approved):
Looks great, and I'd like to see a release soon too!

sgillies:
Me, too! We don't have wheel-building infrastructure right now and it'll take a few days to set it up. On top of that, I'm not really sure how much it will cost to build the wheels on Travis now since the support for open source projects is gone. $.01? $10? I didn't get any insight into that when I was building the wheels from the Mapbox account.

fiona/drvsupport.py
```diff
@@ -99,7 +99,7 @@
     # multi-layer
     #   ("OpenAir", "r"),
     # PCI Geomatics Database File 	PCIDSK 	No 	No 	Yes, using internal PCIDSK SDK (from GDAL 1.7.0)
-    ("PCIDSK", "rw"),
+    ("PCIDSK", "raw"),
```
sgillies, line 102: @akrherz is this change intended?

setup.py
```diff
@@ -295,6 +295,9 @@ def run(self):
     'ordereddict; python_version < "2.7"',
     'enum34; python_version < "3.4"'
 ]
+# Python 3.10 workaround as enum34 not available
+if sys.version_info >= (3, 10):
+    requirements.remove('enum34; python_version < "3.4"')
```
sgillies, line 300: This shouldn't be needed. The intent of `enum34; python_version < "3.4"` is that enum34 only gets installed for python versions lower than 3.4.

tests/test_open.py
```diff
@@ -21,6 +23,7 @@ def test_open_filename_with_exclamation(data_dir):
     assert fiona.open(path), "Failed to open !test.geojson"
 
 
+@requires_gdal21
```
sgillies, line 26: Tests were passing without this before, no?

fiona/drvsupport.py
akrherz: Yes, it was necessary to make the build work and copied from Toblerity/Fiona#1046

sgillies (commented):
@akrherz Thanks! I have some questions inline.

setup.py
akrherz: I hit a bug without this and how python=3.10 gets resolved vs python 3.4 IIRC

tests/test_open.py
akrherz: shrug, [maybe](https://github.com/Toblerity/Fiona/runs/2713722384)

akrherz: Oh, this change is something @mwtoews added to make things greeen, sorry, I forgot what I did here :)

setup.py
sgillies: Oof. I dug and found that this bug was fixed 2 years ago: https://github.com/pypa/pip/pull/7166. What version of pip are you using?

akrherz: I wasn't doing anything, it was whatever CI here is using :)

fiona/drvsupport.py
mwtoews: This is from https://github.com/Toblerity/Fiona/pull/1046#issuecomment-974906111

tests/test_open.py
mwtoews: This is from #1037

rbuffat:
> Me, too! We don't have wheel-building infrastructure right now and it'll take a few days to set it up. On top of that, I'm not really sure how much it will cost to build the wheels on Travis now since the support for open source projects is gone. $.01? $10? I didn't get any insight into that when I was building the wheels from the Mapbox account.

@sgillies A solution to this would be to build wheels locally. In https://github.com/rbuffat/manylinux_gdal I extended the pypa/manylinux_2_24_x86_64 Docker image with precompiled versions of gdal (inspired from fiona-wheels). With the commands outlined in https://github.com/rbuffat/manylinux_gdal/blob/main/Readme.md it is possible to build wheels locally using Docker. Though, I have not tested if they actually work. Probably it will require some more work in order to get it working properly. If you are interested, please contact me so that we can coordinate.

dopplershift:
@sgillies It would be really nice to get a fiona that works on 3.10. As far as wheel-building goes, is there anything special you're using Travis for (e.g. powerpc, arm)? Otherwise a lot of the community has moved to Azure and GitHub actions. I'd be happy to lend my expertise on the GitHub Actions side if there's hope of the work getting merged. Currently, this is keeping me from building docs on 3.10 on some of my projects.

sgillies:
@dopplershift thanks for letting me know! I had assumed that you and your users were getting everything you needed from conda-forge.

Two weeks ago I finished migrating rasterio's wheel infra to GitHub Actions: https://github.com/rasterio/rasterio-wheels. If you've got time, this multibuild configuration could be copied over to a PR against https://github.com/sgillies/fiona-wheels.

sgillies:
Apologies for the slowness, everyone! Merging.

Issue #1031, opened 2021-08-20
Test failures with GDAL 3.3.1


sebastic:
Fiona 1.8.20 fails to build with GDAL 3.3.1 due to test failures:
```
=========================== short test summary info ============================
FAILED tests/test_datetime.py::test_datefield[GPSTrackMaker-datetime] - fiona...
FAILED tests/test_datetime.py::test_datefield_null[GPSTrackMaker-datetime] - ...
FAILED tests/test_drvsupport.py::test_no_append_driver_cannot_append[GPSTrackMaker]
FAILED tests/test_drvsupport.py::test_no_append_driver_cannot_append[PCIDSK]
======= 4 failed, 847 passed, 6 skipped, 2 xfailed, 21 xpassed in 13.20s =======
```
[Full buildlog](https://bugs.debian.org/cgi-bin/bugreport.cgi?att=1;bug=992526;filename=fiona_1.8.20-2_amd64.build;msg=5)

This seems to be caused by deprecated drivers:
```
E   fiona._err.CPLE_AppDefinedError: Driver GTM is considered for removal in GDAL 3.5. You are invited to convert any dataset in that format to another more common one .If you need this driver in future GDAL versions, create a ticket at https://github.com/OSGeo/gdal (look first for an existing one first) to explain how critical it is for you (but the GDAL project may still remove it), and to enable it now, set the GDAL_ENABLE_DEPRECATED_DRIVER_GTM configuration option / environment variable to YES
```
Setting the environment variable resolves leaves just one test failure:
```
=================================== FAILURES ===================================
_________________ test_no_append_driver_cannot_append[PCIDSK] __________________

tmpdir = local('/tmp/pytest-of-pbuilder/pytest-0/test_no_append_driver_cannot_a7')
driver = 'PCIDSK'
testdata_generator = <function testdata_generator.<locals>._testdata_generator at 0x7f2589d6ba60>
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f258980f040>

    @pytest.mark.parametrize('driver', [driver for driver, raw in supported_drivers.items() if
                                        'w' in raw and 'a' not in raw])
    def test_no_append_driver_cannot_append(tmpdir, driver, testdata_generator, monkeypatch):
        """
        Test if a driver that supports write and not append cannot also append
    
        If this test fails, it should be considered to enable append support for the respective driver in drvsupport.py.
    
        """
    
[...]

PR #1060, opened 2022-02-01
Make tests less specific about feature equality

sgillies:
Also reformat.

Towards resolving #935 #1031

### DIFF
--- a/tests/test_drvsupport.py
+++ b/tests/test_drvsupport.py
@@ -311,6 +311,7 @@ def test_no_append_driver_cannot_append(
             c.writerecords(records2)
     except Exception as exc:
         log.exception("Caught exception in trying to append.")
+        return
 
     if driver in {"FileGDB", "OpenFileGDB"}:
         open_driver = driver
--- a/tests/test_drvsupport.py
+++ b/tests/test_drvsupport.py
@@ -10,94 +10,104 @@ from fiona.errors import DriverError
 
 
 @requires_gdal24
-@pytest.mark.parametrize('format', ['GeoJSON', 'ESRIJSON', 'TopoJSON', 'GeoJSONSeq'])
+@pytest.mark.parametrize("format", ["GeoJSON", "ESRIJSON", "TopoJSON", "GeoJSONSeq"])
 def test_geojsonseq(format):
     """Format is available"""
     assert format in fiona.drvsupport.supported_drivers.keys()
 
 
-@pytest.mark.parametrize('driver', [driver for driver, raw in supported_drivers.items() if 'w' in raw])
+@pytest.mark.parametrize(
+    "driver", [driver for driver, raw in supported_drivers.items() if "w" in raw]
+)
 def test_write_or_driver_error(tmpdir, driver, testdata_generator):
     """
-        Test if write mode works.
+    Test if write mode works.
 
     """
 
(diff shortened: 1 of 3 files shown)


tests/test_drvsupport.py
```diff
@@ -10,94 +10,104 @@
...
             open_driver = driver
         else:
             open_driver = None
-        with fiona.open(path, driver=open_driver) as c:
-            assert c.driver == driver
-            items = list(c)
-            assert len(items) == len(records1)
-            for val_in, val_out in zip(records1, items):
-                assert test_equal(driver, val_in, val_out), "in: {val_in}, out: {val_out}".format(val_in=val_in,
-                                                                                                  val_out=val_out)
```
sgillies, line 64: #935 and #1031 report test failures involving coordinate comparison. So I'm removing the feature comparisons. We are really only concerned about the counting of features here. The exact values can be driver/platform dependent.

tests/test_drvsupport.py
```diff
@@ -225,73 +251,73 @@ def test_no_write_driver_cannot_write(tmpdir, driver, testdata_generator, monkey
...
-                    crs=crs,
-                    schema=schema,
-                    **create_kwargs) as c:
+    with fiona.open(
+        path, "w", driver=driver, crs=crs, schema=schema, **create_kwargs
+    ) as c:
 
         c.writerecords(records1)
 
-    is_good = True
```
sgillies, line 268: This complicated the test and made it hard to understand. I've tried to rewrite it without this flag.

tests/test_drvsupport.py
```diff
@@ -225,73 +251,73 @@ def test_no_write_driver_cannot_write(tmpdir, driver, testdata_generator, monkey
...
     try:
-        with fiona.open(path, 'a',
-                        driver=driver) as c:
+        with fiona.open(path, "a", driver=driver) as c:
             c.writerecords(records2)
+    except Exception as exc:
+        log.exception("Caught exception in trying to append.")
```
sgillies, line 307: Reduced the scope of the try/except block.

tests/test_drvsupport.py
```diff
@@ -225,73 +251,73 @@ def test_no_write_driver_cannot_write(tmpdir, driver, testdata_generator, monkey
...
-        is_good = False
+    if driver in {"FileGDB", "OpenFileGDB"}:
+        open_driver = driver
+    else:
+        open_driver = None
 
-    assert not is_good
+    with fiona.open(path, driver=open_driver) as collection:
+        assert collection.driver == driver
+        assert len(collection) == len(records1)
```
sgillies, line 316: I think this is the crux of it. If no new records were added and there was no other exception, the format does not support appending.

sgillies:
PCIDSK is a difficult vector format. It doesn't appear in https://gdal.org/drivers/vector/index.html and got a rewrite at 3.3.0. The driver in versions back to 2.1 seems to support append, at least in my more general rewrite of the tests, so I've adjusted the metadata.

mwtoews:
Re PCIDSK, see https://github.com/Toblerity/Fiona/pull/1046#issuecomment-974906111 which was put into the master branch. I recall looking at the release notes for that GDAL release, but didn't see much for PCIDSK.

PR #1064, opened 2022-02-13
Add pytz to test requirements

QuLogic:
It is used in [`tests/test_datetime.py`](https://github.com/Toblerity/Fiona/blob/master/tests/test_datetime.py#L17).


sgillies:
Thanks @QuLogic !

PR #1020, opened 2021-06-21
BUG: requires_python->python_requires in setup

snowman2:
Saw this warning: `Unknown distribution option: 'requires_python'`

PR #1079, opened 2022-03-08
Establish minimum Python version 3.6 for maint-1.9 branch

mwtoews:
This PR increases the minimum Python version to 3.6 for the maint-1.9 branch, for the upcoming Fiona 1.9.x release series.

It is mainly based on a few cherry picks from:

- #770
- #984
- #1032
- #1020
- #1051

A few manual edits were required. It should have minimal conflicts with `master`. Compatibility with the Fiona 1.8.x series should be maintained.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -6,6 +6,10 @@ All issue numbers are relative to https://github.com/Toblerity/Fiona/issues.
 1.9dev
 ------
 
+Changes:
+
+- Drop Python 2, and establish Python 3.6 as the minimum version (#1079).
+
 Deprecations:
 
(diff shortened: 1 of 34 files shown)


mwtoews:
Any glaring issues here @sgillies ? I might merge this sooner than later before it gets stale.

sgillies:
I'm going to remove 3.6 from the classifiers in setup.py. 3.7 will be the minimum Python version and we may yet increase to 3.8 before 1.9.0 as Python 3.11 is close.

mwtoews:
@sgillies are you doing this? Otherwise, I can increase it to Python 3.8. There are other simple upgrades for Python 3.7 that I can see that I could slip in.

PR #1096, opened 2022-05-11
BUG: force sdist to cythonize files, rather than keeping cythonized files

### LINKED ISSUE #1087: Support Python 3.11
Currently Python 3.11 isn't fully supported in Fiona yet, including testing in CI and publishing wheels to PyPI.

Python 3.11 is expected to be released as stable in October 2022, with many new features including:

> - [PEP 657](https://www.python.org/dev/peps/pep-0657/) -- Include Fine-Grained Error Locations in Tracebacks
> - [PEP 654](https://www.python.org/dev/peps/pep-0654/) -- Exception Groups and except*
> - [PEP 673](https://www.python.org/dev/peps/pep-0673/) -- Self Type
> - [PEP 646](https://www.python.org/dev/peps/pep-0646/)-- Variadic Generics
> - [PEP 680](https://www.python.org/dev/peps/pep-0680/)-- tomllib: Support for Parsing TOML in the Standard Library
> - [PEP 675](https://www.python.org/dev/peps/pep-0675/)-- Arbitrary Literal String Type
> - [PEP 655](https://www.python.org/dev/peps/pep-0655/)-- Marking individual TypedDict items as required or potentially-missing
> - [bpo-46752](https://bugs.python.org/issue46752)-- Introduce task groups to asyncio
> - The [Faster Cpython Project](https://github.com/faster-cpython) is already yielding some exciting results: this version of CPython 3.11 is ~ 19% faster on the geometric mean of the [PyPerformance benchmarks](https://speed.python.org/), compared to 3.10.0.

Fiona is an import package in the geospatial stack, so early support will help speed up Python 3.11 adoption. The last alpha release in the 3.11 series, [Python 3.11.0a7](https://www.python.org/downloads/release/python-3110a7/), has been released earlier this week, and early May the first beta release will be published, after which no new features will be added.

I think it would be an amazing feat to have full Python 3.11 support in Fiona by the time Python 3.11.0 beta 1 gets released, ([expected](https://peps.python.org/pep-0664/) Friday May 6th, 2022). This includes testing in CI and publishing wheels to PyPI.

mwtoews:
This should resolve #1087, where a future Python version is unable to build from a source distribution that include cythonized `*.c` and `*.cpp` files. The source distributions should only keep the source cython files, and a build should cythonize its own files.

Tested locally using `python -m build` with success. (Using Python 3.8 -- Python 3.110b1 not tested)

Other minor changes to MANIFEST.in:

- Include CITATION.txt, which was recently added
- Include pyproject.toml, which was necessary to build binary wheel
- Remove exclude MANIFEST.in, <s>as this is normally excluded</s> (update: this is [normally included](https://packaging.python.org/en/latest/guides/using-manifest-in/), so perhaps there's a good reason to leave it alone)

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,11 +2,11 @@ global-exclude .DS_Store
 global-exclude *.pyc
 recursive-exclude docs/data *
 recursive-exclude docs/_build *
-exclude MANIFEST.in
 exclude *.txt *.py
 recursive-include docs *.rst *.txt
 recursive-include tests *.py
 recursive-include tests/data *
-include fiona/*.c fiona/*.cpp
-include CHANGES.txt CREDITS.txt LICENSE.txt VERSION.txt README.rst
-include benchmark.py setup.py requirements.txt
+recursive-include fiona *.pyx *.pxd *.pxi
+recursive-exclude fiona *.c *.cpp
+include CHANGES.txt CITATION.txt CREDITS.txt LICENSE.txt VERSION.txt README.rst
+include benchmark.py pyproject.toml setup.py requirements.txt
--- a/setup.py
+++ b/setup.py
@@ -49,10 +49,6 @@ with open('CREDITS.txt', **open_kwds) as f:
 with open('CHANGES.txt', **open_kwds) as f:
     changes = f.read()
 
-# Set a flag for builds where the source directory is a repo checkout.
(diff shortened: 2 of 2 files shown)


sgillies (approved):
Thanks @mwtoews !

PR #1112, opened 2022-06-03
Draft of a geometry operations applicator

### LINKED ISSUE #982: PROPOSAL: Segmentize
https://gdal.org/python/osgeo.ogr.Geometry-class.html#Segmentize

Would be useful to add an interface through fiona. Probably something like `transform_geom`. Thoughts?

sgillies:
Resolves #982

### DIFF
--- a/fiona/_transform.pyx
+++ b/fiona/_transform.pyx
@@ -11,6 +11,7 @@ from collections import UserDict
 
 from fiona cimport _cpl, _csl, _geometry
 from fiona.crs cimport OGRSpatialReferenceH, osr_set_traditional_axis_mapping_strategy
+from fiona._err cimport exc_wrap_pointer
 
 from fiona.compat import DICT_TYPES
 from fiona.crs import CRS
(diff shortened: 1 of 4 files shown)


fiona/_transform.pyx
```diff
@@ -263,3 +264,63 @@ def _transform_geom(src_crs, dst_crs, geom, antimeridian_cutting, antimeridian_o
     OSRRelease(dst)
 
     return out_geom
+
+
+from libcpp.unordered_map cimport unordered_map
+
+cdef OGRGeometryH segmentize(OGRGeometryH geom, double max_length):
```
sgillies, line 271: Because OGR_G_Segmentize is an oddball and returns nothing (`void`).

fiona/_transform.pyx
```diff
@@ -263,3 +264,63 @@ def _transform_geom(src_crs, dst_crs, geom, antimeridian_cutting, antimeridian_o
...
+    return new_geom
+
+
+ctypedef OGRGeometryH (*geometry_func)(OGRGeometryH, double)
+cdef unordered_map[int, geometry_func] func_map
+
+func_map[1] = <geometry_func>segmentize
+
+cdef geometry_func func_alias(int choice):
+    return func_map[choice]
```
sgillies, line 284: We can't just put `OGR_G_Segmentize` (for example) in a Python dict. https://stackoverflow.com/a/56708967 was very helpful in showing a solution.

fiona/_transform.pyx
```diff
@@ -263,3 +264,63 @@ def _transform_geom(src_crs, dst_crs, geom, antimeridian_cutting, antimeridian_o
...
+    Geometry
+
+    """
+    cdef OGRGeometryH ogr_geom1 = NULL
+    cdef OGRGeometryH ogr_geom2 = NULL
+    cdef int choice = 0
+
+    op_map = {
+        "segmentize": 1
+    }
```
sgillies, line 308: The map to the map of OGR functions.

fiona/_transform.pyx
```diff
@@ -263,3 +264,63 @@ def _transform_geom(src_crs, dst_crs, geom, antimeridian_cutting, antimeridian_o
...
+
+    for geom in geoms:
+        ogr_geom1 = _geometry.OGRGeomBuilder().build(geom)
+
+        for op_name, op_arg in operations:
```
sgillies, line 313: Basically it's a per-geometry pipeline implemented in C.

The other obvious design is to add segmentize &c as methods to our Geometry class. But performance won't be great until version 2.0 of fiona when Geometry is an extension class and carries an OGRGeometryH or a zero-copy data structure around with it.

sgillies:
@snowman2 I have a draft solution for #982 here. I'm going to remove it from the fiona 1.9a2 milestone since there's still some discussion to be had about the design.

fiona/_transform.pyx
snowman2: Might not be bad to wait until 2.0 for this as the `Geometry` class seems to have the potential to significantly impact on the design.

PR #1121, opened 2022-07-04
Trying out OGR_L_GetArrowStream

sgillies:
Towards resolving #469

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -1443,6 +1443,84 @@ cdef class WritingSession(Session):
         return GDALSetMetadataItem(obj, name, value, domain)
 
 
+cdef extern from "stdint.h":
+    ctypedef signed int int64_t
+
+
+cdef extern from "ogr_recordbatch.h":
+
+    cdef struct ArrowSchema:
+        const char* format
+        const char* name
+        const char* metadata
+        int64_t flags
+        int64_t n_children
+        ArrowSchema **children
+        ArrowSchema *dictionary
+        void (*release)(ArrowSchema *)
+
+    cdef struct ArrowArray:
+        int64_t length
+        int64_t null_count
+        int64_t offset
+        int64_t n_buffers
+        int64_t n_children
+        const void **buffers
+        ArrowArray **children
+        ArrowArray *dictionary
+        void (*release)(ArrowArray *)
+
+    cdef struct ArrowArrayStream:
+        int (*get_schema)(ArrowArrayStream *, ArrowSchema *out)
+        int (*get_next)(ArrowArrayStream *, ArrowArray *out)
+        const char* (*get_last_error)(ArrowArrayStream *)
+        void (*release)(ArrowArrayStream *)
+
+cdef extern from "ogr_api.h" nogil:
+
+    bint OGR_L_GetArrowStream(OGRLayerH hLayer, ArrowArrayStream *out_stream, char **papszOptions)
+
+
+cdef class BatchReader:
+    cdef collection
+
+    def __init__(self, collection):
+        self.collection = collection
+
+    def batches(self):
+        cdef Session session = self.collection.session
+        cdef OGRLayerH cogr_layer = session.cogr_layer
+
+        if not session or not session.isactive:
+            raise FionaValueError("Session is inactive, dataset is closed or layer is unavailable.")
+
+        cdef ArrowArrayStream stream
+
+        if not OGR_L_GetArrowStream(cogr_layer, &stream, NULL):
+            raise CPLE_AppDefinedError("OGR_L_GetArrowStream() failed")
+
+        cdef ArrowSchema schema
+
+        if stream.get_schema(&stream, &schema) == 0:
+            # print(schema.name)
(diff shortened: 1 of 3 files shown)


fiona/ogrext.pyx
```diff
@@ -1443,6 +1443,84 @@ cdef class WritingSession(Session):
...
+        cdef ArrowArray array
+
+        while True:
+            # Look for an error (get_next() returning a non-zero code), or
+            # end of iteration (array.release == nullptr)
+
+            if stream.get_next(&stream, &array) != 0 or array.release == NULL:
+                break
+
+            yield array.length
```
kylebarron, line 1517: Is this your issue? That instead of yielding pyarrow arrays, you're yielding ints?

Not an expert on this but you might need to use `pyarrow.Array._import_from_c` to construct the pyarrow table from the C memory? https://arrow.apache.org/docs/python/integration/python_r.html#r-to-python-communication-using-the-c-data-interface

It looks like the native OGR Python bindings use this as well: https://github.com/OSGeo/gdal/pull/5922/files#diff-738a05d250b30ce772021dcebe2793001a68c7548211b6424137b330133f5b8dR2874

sgillies: I'm yielding int there only as a baby step, to confirm that I'm otherwise using the API and structs properly without crashing.

Depend on pyarrow and the arrow lib or implement new, more limited data classes, that's the question.

kylebarron: My hunch would be that an optional dependency on `pyarrow` would be the most straightforward approach. Essentially everyone who is using Arrow memory in Python is using pyarrow already

kylebarron:
Watching @jorisvandenbossche talk about this now, giving an example of using the pure-GDAL API, and thought I should check back in on this issue 🙂. Anything I can help with on this PR to move it forward?

Screen Shot 2022-10-05 at 1 20 22 PM

sgillies:
@kylebarron wanna contribute some sort of acceptance test? I'm not working in this space right now and so I'm not confident I can identify good targets, but I'm confident that I can hit good targets :smile: If you still think an optional dependency on pyarrow is okay, I'm fine with that, too.

kylebarron:
Taking a look at https://github.com/geopandas/pyogrio/pull/155 might be wise as well. [This comment](https://github.com/geopandas/pyogrio/pull/155#issuecomment-1254983918) on GDAL dataset lifetimes makes my head hurt 😅. There are also some tests in that PR. I was going to use the raw python GDAL API to create example output data to test against, but realized I don't actually know how to build unreleased GDAL. I guess docker is the best route there?

sgillies:
@kylebarron thanks. The remaining question for me is: "is this worth the effort?" The potential user would be geopandas (Right? I'm not aware of anything else) and that project is already routing around fiona's lack of a columnar data API with pyogrio. Could be a waste of time to duplicate that effort. Why shouldn't Python projects use pyogrio to read data like this?

jorisvandenbossche:
> but realized I don't actually know how to build unreleased GDAL. I guess docker is the best route there?

@kylebarron docker is a good option (GDAL provides some docker images, https://gdal.org/download.html#linux-docker-images), but conda actually makes this easy as well if you would prefer that. "Just" create an environment with all GDAL's dependencies, clone GDAL, and doing a basic cmake build works out of the box for me (https://gdal.org/build_hints.html#cmake, I only needed to add `-DCMAKE_PREFIX_PATH=$CONDA_PREFIX -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX` to the `cmake` command to ensure to install it in the conda env)

kylebarron:
> Could be a waste of time to duplicate that effort. Why shouldn't Python projects use pyogrio to read data like this?

A larger, naive question is: for reading columnar data with GDAL 3.6+, what does pyogrio provide that isn't in this GDAL-native Arrow API? Obviously a ton of hard work has gone into pyogrio, and I don't want to denigrate that. But with GDAL's Arrow API, a simple binding to that API sounds way simpler than a custom cython row-to-columnar wrapper. If so, would it make sense for the community to coalesce on a single library? (Not sure where it makes sense to have this conversation)

> The potential user would be geopandas (Right? I'm not aware of anything else)

My interest in having Python bindings (in some library) to the GDAL Arrow API is foremost to improve the general geoarrow ecosystem in Python, but also for easy integration into [my side project](https://github.com/geopolars/geopolars) which uses a pure geoarrow memory layout.

jorisvandenbossche:
> A larger, naive question is: for reading columnar data with GDAL 3.6+, what does pyogrio provide that isn't in this GDAL-native Arrow API? Obviously a ton of hard work has gone into pyogrio, and I don't want to denigrate that.

No, that's certainly a good question, and I have wondered that myself. At the moment, the main thing is that we still have improved writing as well, although that is something that could be brought up to GDAL if they would want to support an inverse like "SetArrowStream" in addition to the read API. 

(we also have vectorized `read_bounds`, and reading by FID, tighter integration with geopandas. And a smaller API surface compared to Fiona. But at the same time we also duplicate quite some utility code from fiona ..)

Issue #1099, opened 2022-05-16
tests/test_drvsupport.py::test_no_append_driver_cannot_append[FlatGeobuf] segfault with GDAL 3.5.0


sgillies:
See https://github.com/OSGeo/gdal/issues/5739

snowman2:
I am able to reproduce with GDAL 3.4.0 (#1114) and 3.4.3:
```python
>>> import fiona
>>> fiona.show_versions()
Fiona version: 1.9a2
GDAL version: 3.4.3
PROJ version: 9.0.0

OS: Linux 5.13.0-44-generic
Python: 3.9.13
```

dotlambda:
With GDAL 3.5.1 I get
```
_______________ test_no_append_driver_cannot_append[FlatGeobuf] ________________

tmpdir = local('/build/pytest-of-nixbld/pytest-0/test_no_append_driver_cannot_a1')
driver = 'FlatGeobuf'
testdata_generator = <function testdata_generator.<locals>._testdata_generator at 0x7fffc4fc5000>
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7fffc4fe69b0>

    @pytest.mark.parametrize(
        "driver",
        [
            driver
            for driver, raw in supported_drivers.items()
            if "w" in raw and "a" not in raw
        ],
    )
    def test_no_append_driver_cannot_append(
        tmpdir, driver, testdata_generator, monkeypatch
    ):
        """
        Test if a driver that supports write and not append cannot also append

        If this test fails, it should be considered to enable append support for the respective driver in drvsupport.py.

        """

        monkeypatch.setitem(fiona.drvsupport.supported_drivers, driver, "raw")

        if driver == "BNA" and GDALVersion.runtime() < GDALVersion(2, 0):
            pytest.skip("BNA driver segfaults with gdal 1.11")

        path = str(tmpdir.join(get_temp_filename(driver)))
        schema, crs, records1, records2, test_equal, create_kwargs = testdata_generator(
            driver, range(0, 5), range(5, 10)
        )

        # If driver is not able to write, we cannot test append
        if driver in driver_mode_mingdal[
            "w"
        ] and get_gdal_version_num() < calc_gdal_version_num(
            *driver_mode_mingdal["w"][driver]
        ):
            return

        # Create test file to append to
        with fiona.open(
            path, "w", driver=driver, crs=crs, schema=schema, **create_kwargs
        ) as c:

            c.writerecords(records1)

        try:
[...]

PR #1122, opened 2022-07-06
Add gdal 3.5 and 3.6 to CI, update proj/gdal install scripts to use cmake, update drvsupport

rbuffat:
See https://github.com/Toblerity/Fiona/issues/1099#issuecomment-1176004721


mwtoews:
This pinned version of munch is incompatible with Python 3.11: https://github.com/Toblerity/Fiona/blob/9d06389755928a3ae4ffa55d4156700f5260ec1c/requirements.txt#L5
not sure which one to pick, but there are [several newer versions](https://pypi.org/project/munch/#history)

rbuffat:
Looks like I opened a can of worms. 

This PR:
- adds GDAL 3.5 to the CI matrix. 
- updates install scripts to use cmake. This is necessary for Proj 9, and will be necessary for GDAL 3.6 (which is on master already). 
- Update drvsupport:  flatgeobuf gained support to append with gdal 3.5, openfilegdb will gain support for write and append with gdal 3.6

The tests testing GDAL functionality (e.g. read / write / append support of drivers or support of drivers for time / datetime) are now marked with @pytest.mark.gdal. These tests should only be executed when adding the  `--test_gdal` flag to pytest. 

Thanks @mwtoews for having a look at this PR.

rbuffat:
There are still some regressions: 

The cmake configuration step emits the following warning and I'm not exactly sure why: 
```
CMake Warning:
  Ignoring extra path from command line:
   "/"
```

The FileGDB driver is not included with the cmake install script. But as gdal 3.6 seems to support writing GDB files with the OpenFileGDB driver this seems not be that important.

sgillies:
@rbuffat can we do without the `--test_gdal` option? Is there anything about it that isn't possible with `pytest -m gdal" or `pytest -m "not gdal"`?

rbuffat:
@sgillies Sorry I forgot about this PR.

> @rbuffat can we do without the --test_gdal option? Is there anything about it that isn't possible with pytest -m gdal" or pytest -m "not gdal"?

What I intended to implement is that the `gdal` marked tests are disabled by default and must be enabled manually. So that we can enable it on our CI and we can avoid situations where others have issues with them when they fail on exotic platforms.  Based on https://stackoverflow.com/questions/66315234/exclude-some-tests-by-default it should work also without an additional command line argument.  Lets see if this works.

rbuffat:
With the current changes, when pytest is run using `pytest -m "not wheel" ...` or `pytest ...` (not explicitly including the `gdal` marker) all tests marked with `@pytest.mark.gdal` are skipped:

```
tests/test_datetime.py ...sssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss.                                                                                                                             
tests/test_drvsupport.py ssssssssssssssssssssssssssssssssssssssssssssssssss. 
```

When using `pytest -m "not wheel or gdal"` (explicitly including the gdal marker), these tests are not skipped and are included:
```
tests/test_datetime.py .................................................................................................                                                                                                                             
tests/test_drvsupport.py ...............................x...................           
```

@sgillies I'm not sure this is the way this should be done. It looks too complicated, which often indicates that something is done wrong. I will look at it again when I have more brain capacity left. Or what is your opinion, should we avoid meddling with the markers in `conftest.py/pytest_collection_modifyitems`?

rbuffat:
@sgillies  This PR should now be ready for a review. 

- If you don't like the changes in  conftest.py/pytest_collection_modifyitems, let's just remove them. 
- pytest.mark.gdal is not the best name, maybe pytest.mark.ci_only would be better?

fiona/drvsupport.py
```diff
@@ -56,7 +56,7 @@
     # GeoJSON 	GeoJSON 	Yes 	Yes 	Yes
     ("GeoJSON", "raw"),
     # GeoJSONSeq 	GeoJSON sequences 	Yes 	Yes 	Yes 
-    ("GeoJSONSeq", "rw"),
+    ("GeoJSONSeq", "raw"),
```
rouault, line 59: "raw" only since GDAL 3.6

rbuffat: It is limited in the `driver_mode_mingdal` dict:  https://github.com/Toblerity/Fiona/blob/ac7aeeb47fd6db4733187c3117303a1bcbc4640a/fiona/drvsupport.py#L169

zklaus:
@rbuffat, friendly ping. Is this something you see moving forward soon?

rbuffat:
@zklaus The PR is ready to be reviewed. It is currently only for the 1.9 branch, but the same changes should also apply for the 1.8 branch. 

This PR does not change the functionality of Fiona (expect that without it Fiona limits some driver capabilities that GDAL gained with version 3.6)

pytest.ini
```diff
@@ -8,6 +8,7 @@ filterwarnings =
 markers = 
     iconv: marks tests that require gdal to be compiled with iconv
     network: marks tests that require a network connection
-    wheel: marks test that only works when installed from wheel
+    wheel: marks tests that only works when installed from wheel
+    gdal: marks tests that are only dependent on GDAL functionality (e.g. for drvsupport)
```
sgillies, line 12: :+1:

sgillies (approved):
@rbuffat thank you!

rbuffat:
@sgillies should this be backported to maint-1.8 for a 1.8.23 release?

sgillies:
@rbuffat if people use Fiona 1.8.22 with 3.6.0 they will see a few failing tests, but no breakage, right? I'm inclined to not backport and make a new release.

zklaus:
Fyi, I have effectively backported this in conda-forge/fiona-feedstock#203

PR #1128, opened 2022-07-24
ENH: Add support for PROJ_DATA environment variable

### LINKED ISSUE #1127: PROJ_DATA environment variable
Forward from: https://github.com/rasterio/rasterio/issues/2518

PROJ 9.1 recognizes the PROJ_DATA environment variable, as a replacement
for PROJ_LIB (both are still recognized by PROJ >= 9.1. PROJ_LIB will be
retired in PROJ 10, whose release is unplanned for now)

Cf OSGeo/PROJ#3253

snowman2:
Closes #1127

### DIFF
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -87,7 +87,7 @@ jobs:
 
           # Additional env variables
           echo "GDAL_DATA=$GDALINST/gdal-$GDALVERSION/share/gdal" >> $GITHUB_ENV
-          echo "PROJ_LIB=$GDALINST/gdal-$GDALVERSION/share/proj" >> $GITHUB_ENV
+          echo "PROJ_DATA=$GDALINST/gdal-$GDALVERSION/share/proj" >> $GITHUB_ENV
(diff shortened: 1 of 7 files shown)


sgillies:
@snowman2 I wonder how long we'll be supporting `PROJ_LIB`? A year or more? I think this will definitely hold for that long. Thank you!

snowman2:
> I wonder how long we'll be supporting `PROJ_LIB`? A year or more?

It will be dropped in PROJ 10 https://github.com/OSGeo/PROJ/issues/3255, though I am not sure on the planned release date.

jorisvandenbossche:
> I wonder how long we'll be supporting PROJ_LIB?

You should probably keep it as long as you support PROJ versions that use that env variable? (which I suppose will be longer than a year)

sgillies:
We have a numpy and GDAL version policy based on https://numpy.org/neps/nep-0029-deprecation_policy.html. We'll add PROJ. Even though fiona doesn't use the PROJ API (only via GDAL), there's no way around PROJ_LIB/PROJ_DATA.

snowman2:
Thanks @sgillies :+1:

PR #1139, opened 2022-09-15
Remove `setuptools` as a runtime dependency

cpcloud:
Depending on `setuptools` at runtime (`install_requires` is for
specifying runtime dependencies) creates problems for external packaging
tools such as [poetry2nix](https://github.com/nix-community/poetry2nix) due to
the need to use setuptools to bootstrap package installs in nix itself.

`setuptools` is already installed by the time this line of code runs,
and isn't imported anywhere in the codebase. There should be zero change
in behavior, while allowing external packaging tools like poetry2nix to
cleanly handle Fiona.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -223,7 +223,6 @@ requirements = [
     'cligj>=0.5',
     'click-plugins>=1.0',
     'munch',
-    "setuptools",
 ]
 
 extras_require = {


cpcloud:
@sgillies Any chance you could give this a review?

mwtoews:
This rolls back fab6e4e86bb3995a86e13de7ea0e274a201dba38 but I'll agree that it should not be a runtime dependency, as it is part of pyproject.toml.

@cpcloud could you re-target this PR to `maint-1.9` branch?

cpcloud:
@mwtoews Changed the base to `maint-1.9`.

mwtoews (approved):

sgillies:
Thanks @cpcloud !

PR #1147, opened 2022-10-11
Linux CI: Add Python 3.11

EwoutH:
Add a Python 3.11 job to the Linux CI.

Also updates the used actions to their latest versions.

### DIFF
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -48,6 +48,12 @@ jobs:
               PROJVERSION: "8.2.1",
               allow_failure: "false",
             }
+          - {
+              python: "3.11-dev",
+              GDALVERSION: "3.4.2",
+              PROJVERSION: "8.2.1",
+              allow_failure: "false",
+            }
 
           # Test GDAL master
           - {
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_linux.yml
```diff
@@ -48,6 +48,12 @@ jobs:
               PROJVERSION: "8.2.1",
               allow_failure: "false",
             }
+          - {
+              python: "3.11-dev",
+              GDALVERSION: "3.4.2",
+              PROJVERSION: "8.2.1",
```
mwtoews, line 54: We could try `GDALVERSION: "3.5.2"` and `PROJVERSION: "9.1.0"` here, but it could break things.

.github/workflows/ci_linux.yml
```diff
@@ -48,6 +48,12 @@ jobs:
...
+              allow_failure: "false",
```
mwtoews, line 55: give this a try:
```suggestion
              GDALVERSION: "3.5.2",
              PROJVERSION: "9.1.0",
              allow_failure: "true",
```

.github/workflows/ci_linux.yml
EwoutH: Already tried, and it indeed breaks things.

My idea was to first get Python 3.11 working, and then look at updating GDAL in another PR (like #1122).

EwoutH:
Thanks for reviewing! I would suggest leaving this PR to just getting Python 3.11 working, and then look at updating GDAL in another PR (like https://github.com/Toblerity/Fiona/pull/1122).

mwtoews (approved):

Issue #1158, opened 2022-10-29
slow layer writes when writing features with hundreds of columns


fcostin:
## Expected behavior and actual behavior.

I attempted to write a 35,000 record polygon dataset with 400 columns to FlatGeobuf using Fiona 1.8.21 (via geopandas). It took a couple of minutes to query and download this dataset from a remote server, and turn it into a geopandas GeoDataFrame in ram, so I expected it might take a minute or so to write. The write operation using geopandas engine="fiona" kept running for 10 minutes, until I interrupted it.

The current unreleased Fiona in master branch (rev 9c8fe736) appears to be substantially faster than Fiona 1.8.21.

## Steps to reproduce the problem.

I am unable to share the original dataset, but here is a script to create synthetic data with similar characteristics and write it:

```python3
# pip install geopandas
# pip install pyogrio

import numpy
import pandas as pd
import geopandas as gpd
import random
import shapely
import string

ALPHABET = list(string.ascii_lowercase)


def make_vocab(n, w):
    return numpy.asarray([''.join(random.sample(ALPHABET, w)) for _ in range(n)], dtype=object)


def make_random_data(n_rows: int, n_cols: int, w: int):
    k = 4321
    vocab = make_vocab(n=k, w=16)
    data = {}
    for i in range(n_cols):
        col_values = [vocab[((i + (i*j))%k)] for j in range(n_rows)]
        col_name = 'col_%d' % (i, )
        data[col_name] = pd.Series(col_values)

    x_coords = numpy.random.uniform(-1.0, 1.0, (n_rows, ))
    y_coords = numpy.random.uniform(-1.0, 1.0, (n_rows, ))

    geom = [shapely.geometry.Point((x, y)) for (x, y) in zip(x_coords, y_coords)]

    return gpd.GeoDataFrame(data=data, geometry=geom)


def main():
    print('making random data...')
    n_rows = 25000
    n_cols = 400
    w = 16
    gpd = make_random_data(n_rows, n_cols, w)
    print('writing...')
    driver='FlatGeobuf'
    engine='fiona'
    # engine='pyogrio'
    gpd.to_file('data.fgb', driver=driver, engine=engine)


if __name__ == '__main__':
    main()
```

## Operating system

Ubuntu 22.04.1 LTS

## Fiona and GDAL version and provenance

Slow write performance noticed with Fiona 1.8.21  installed from PyPI using pip version 9.0.1.

## Potential speedups

[...]

PR #1159, opened 2022-10-29
ogrext writerecords performance

fcostin:
This PR proposes three relatively simple opportunities to speed up writes by doing less work:

1.  ogrext - performance - cache OGR field indices  (can give 2x speedup when writing layers with 400 string columns)
2.  ogrext - performance - cache OGR normalized fields (can give another ~ 10% speedup)
3.  ogrext - writerecs - remove duplicated code (minor speedup)

I had some time to look at this after noticing writes seemed very slow for layers with large number of columns (ref: https://github.com/Toblerity/Fiona/issues/1158).

all timings collected:
- using cProfile
- writing a dataset via geopandas to_file

### benchmark - comparison on synthetic scenario with high column count

data:

-   25,000 records
-   random point geometries
-   400 string columns, each containing random length-16 strings

| library |  version                   |  function                             |  time to write layer (seconds) |
| ------- | -------------------------  | ------------------------------------- | ------------------------------ |
| Fiona   | master branch rev 9c8fe736 | fiona/collection.py:500(writerecords) |   39.719                       |
| Fiona   | proposed, rev bb5bd021     | fiona/collection.py:500(writerecords) |   15.208                       |

Please see https://github.com/Toblerity/Fiona/issues/1158 for the script used for this benchmark.

### benchmark - comparison on real data

data:

-  35,098 records (land parcels with attributes)
-  polygon geometries
-  subset of 15 of the full 400 columns in the data, (includes a mix of strings, missing values, flag columns, etc)

| library |  version                   |  function                             |  time to write layer (seconds) |
| ------- | -------------------------  | ------------------------------------- | ------------------------------ |
| Fiona   |  1.8.21                    | fiona/collection.py:355(writerecords) |  10.495                        |
| Fiona   |  1.9a3                     | fiona/collection.py:544(writerecords) |  11.497                        |
| Fiona   | master branch rev 9c8fe736 | fiona/collection.py:500(writerecords) |   7.473                        |
| Fiona   | proposed, rev bb5bd021     | fiona/collection.py:500(writerecords) |   7.092                        |
| pyogrio |  0.4.2                     | pyogrio/raw.py:163(write)             |   0.560                        |

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -1349,13 +1349,6 @@ cdef class WritingSession(Session):
                         record['geometry']['type'],
                         collection.schema['geometry'] ))
 
-            # Validate against collection's schema to give useful message
-            if set(record['properties'].keys()) != schema_props_keys:
-                raise SchemaError(
-                    "Record does not match collection schema: %r != %r" % (
-                        record['properties'].keys(),
-                        list(schema_props_keys) ))
-
             cogr_feature = OGRFeatureBuilder().build(record, collection)
             result = OGR_L_CreateFeature(cogr_layer, cogr_feature)
 
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -360,15 +360,11 @@ cdef class OGRFeatureBuilder:
         encoding = session._get_internal_encoding()
 
         for key, value in feature['properties'].items():
-            ogr_key = session._schema_mapping[key]
-
-            schema_type = normalize_field_type(collection.schema['properties'][key])
-
-            key_bytes = strencode(ogr_key, encoding)
-            key_c = key_bytes
-            i = OGR_F_GetFieldIndex(cogr_feature, key_c)
+            i = session._schema_mapping_index[key]
             if i < 0:
                 continue
+
+            schema_type = normalize_field_type(collection.schema['properties'][key])
 
             # Special case: serialize dicts to assist OGR.
             if isinstance(value, dict):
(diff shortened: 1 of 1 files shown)


fiona/ogrext.pyx
```diff
@@ -1342,13 +1349,6 @@ cdef class WritingSession(Session):
                         record['geometry']['type'],
                         collection.schema['geometry'] ))
 
-            # Validate against collection's schema to give useful message
-            if set(record['properties'].keys()) != schema_props_keys:
-                raise SchemaError(
-                    "Record does not match collection schema: %r != %r" % (
-                        record['properties'].keys(),
-                        list(schema_props_keys) ))
```
fcostin, line 1350: this duplicates lines 1338 -- 1343, with a different exception type, so i would expect this code to be unreachable.  perhaps these two blocks are an unintended result of a merge

fiona/ogrext.pyx
```diff
@@ -1285,6 +1283,15 @@ cdef class WritingSession(Session):
         self._schema_mapping = dict(zip(before_fields.keys(),
                                         after_fields.keys()))
 
+        # Mapping of the Python collection schema to OGR field indices.
+        # We assume that get_schema()['properties'].keys() is in the exact OGR field order
+        assert len(before_fields) == len(after_fields)
+        self._schema_mapping_index = dict(zip(before_fields.keys(), range(len(after_fields.keys()))))
```
fcostin, line 1289: instead of looking up the field index by `OGR_F_GetFieldIndex`, using the field name and cogr_feature in `OGRFeatureBuilder.Build()`, each time we write a record, the idea is to note the order of the field definitions from when we inspect the layer schema, and store that once for reuse when writing all records.

i am not very familiar with ogr, if field indices can vary arbitrary per record, this will be defective.

sgillies: @fcostin seems like a good strategy. Field indices are per-layer.

sgillies (approved):
@fcostin thank you!

PR #1162, opened 2022-11-17
drvsupport.py: changes for 3.6.0 compatibility

rouault:
This fixes the ``test_no_append_driver_cannot_append[GeoJSONSeq]`` failure of the libgdal36 rebuild of https://github.com/conda-forge/fiona-feedstock/pull/202

But locally I get other test failures which are suspicious. I didn't investigate more
```
FAILED tests/test_datetime.py::test_datefield[FlatGeobuf-datetime] - AssertionError: 2018-03-25T22:49:05 does not match 2018-03-25T22:49:05.230000
FAILED tests/test_datetime.py::test_datefield[GeoJSON-datetime] - AssertionError: 2018-03-25T22:49:05 does not match 2018-03-25T22:49:05.230000
FAILED tests/test_datetime.py::test_datefield[GeoJSONSeq-datetime] - AssertionError: 2018-03-25T22:49:05 does not match 2018-03-25T22:49:05.230000
```


rouault:
Ah, seeing that my pull request overlaps with https://github.com/Toblerity/Fiona/pull/1122  (mine has conditional capabilities depending on the GDAL version for the OpenFileGDB, FlatGeoBuf and GeoJSONSeq drivers though) which is actually required to have a CI testing GDAL master

rbuffat:
@rouault Fiona includes tests for certain driver characteristics maintained in drvsupport.py (e.g. support for different field types (date, datetime, etc) or modes (read, append, write)). The tests work by trying to write a file and then reading it again and checking if the expected result is read back. By nature, these tests are a bit messy (e.g. there are cases where the return values are dependent on the platform) and do not really test Fiona's functionality and only the functionality of the drivers. In https://github.com/Toblerity/Fiona/pull/1122 (which is unfortunately still a WIP) I try to mark all these tests with a marker so that these tests can easily be deactivated. (Actually, the initial goal was to deactivate these by default, so that e.g. fiona-feedstock or other package maintainers (https://github.com/Toblerity/Fiona/issues/935) do not have to worry about them as they need only to be run on Fiona's CI. But it turned out this is more complicated than I expected.) 

Regarding the FileGDB and OpenFileGDB not supporting `date`, I need to check why this is not detected. When I'm not mistaken FlatGeobuf seems to convert `time` and `date` to `datetime`.

rbuffat:
I included gdal 3.6 now also in https://github.com/Toblerity/Fiona/pull/1122

rouault:
closing as superseded per #1122

Issue #1155, opened 2022-10-21
UnsupportedGeometryTypeError in Fiona 1.8.22


Meresmata:
## Expected behavior and actual behavior.

Opening the file. Is working wit some files. Others raises the error below:

## Steps to reproduce the problem.

open the file with geopandas' read_file

## Operating system

OS Ubuntu 20.04 derivative

## Fiona and GDAL version and provenance

installed with mamba 0.27.0:
fiona 1.8.2.2
gdal 3.5.2
Example Shape File: [LoD1_23445938.zip](https://github.com/Toblerity/Fiona/files/9839473/LoD1_23445938.zip)
(1/3 of the LoD1 files (by Lower Saxony (open Data)) I tried did not open). But this example opens in QGIS.

```


  File "fiona/ogrext.pyx", line 1515, in fiona.ogrext.Iterator.__next__
  File "fiona/ogrext.pyx", line 317, in fiona.ogrext.FeatureBuilder.build
  File "fiona/_geometry.pyx", line 187, in fiona._geometry.GeomBuilder.build
  File "fiona/_geometry.pyx", line 168, in fiona._geometry.GeomBuilder._buildGeometryCollection
  File "fiona/_geometry.pyx", line 169, in fiona._geometry.GeomBuilder._buildGeometryCollection
  File "fiona/_geometry.pyx", line 149, in fiona._geometry.GeomBuilder._buildParts
  File "fiona/_geometry.pyx", line 182, in fiona._geometry.GeomBuilder.build
fiona.errors.UnsupportedGeometryTypeError: 16
```

jorisvandenbossche:
Do you know what type of geometries you have stored in the file? The error indicates that fiona received a geometry type it doesn't support (16 typically indicates TIN, https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary)

Meresmata:
QGIS states MultiPolygonZ, which makes sence for LoD1. I could open about 66 %. (I added an example in my first post)

jorisvandenbossche:
Using `ogrinfo -al LoD1_23445938_2_2021-01-19.shp` from the command line, there are indeed TIN geometries included in GeometryCollections (most of the geoemtries are MultiPolygons, a few of those mixed collections):

```
...
OGRFeature(LoD1_23445938_2_2021-01-19):102
[...]

Issue #1130, opened 2022-08-04
UnsupportedGeometryTypeError (16) with shape file from open data base.


uvchik:
## Expected behavior and actual behavior.

I expected to be able read a 3D shapefile (via zip) downloaded from the open geodata base of Lower Saxonia in Germany as I am able to open it in QGIS (v3.26.0).

https://opengeodata.lgln.niedersachsen.de/#lod2

By the way it works if I open the file with QGIS and store the file with QGIS, so it seems that QGIS is fixing something simple by using load/store.

I was not able to find the problematic geometry.

Geometry type 16 lead me to PR #679 so it should be fixed in my Fiona versions (1.8.21 or 1.9a2).
See: https://github.com/Toblerity/Fiona/blob/9c8fe736d4381526905c6976c399a33d10e3ecbf/fiona/ogrext.pyx#L301-L320

## Steps to reproduce the problem.

Download the following [zip-test-file](https://lod2.opengeodata.lgln.niedersachsen.de/25505804/2020-11-19/LoD2_25505804_2_2020-11-19.zip).

````python
import geopandas as gpd

fn = "path/to/downloaded/file/LoD2_25505804_2_2020-11-19.zip"
gpd.read_file(fn)

````

Reading the file leads to:

```python
Traceback (most recent call last):
  File "open_geo_data_test.py", line 72, in <module>
    maps.append(gpd.read_file(shp_filename))
  File "...\site-packages\geopandas\io\file.py", line 253, in _read_file
    return _read_file_fiona(
  File "...\site-packages\geopandas\io\file.py", line 340, in _read_file_fiona
    df = GeoDataFrame.from_features(
  File "...\site-packages\geopandas\geodataframe.py", line 640, in from_features
    for feature in features_lst:
  File "fiona\ogrext.pyx", line 1658, in fiona.ogrext.Iterator.__next__
  File "fiona\ogrext.pyx", line 384, in fiona.ogrext.FeatureBuilder.build
  File "fiona\_geometry.pyx", line 159, in fiona._geometry.GeomBuilder.build
  File "fiona\_geometry.pyx", line 140, in fiona._geometry.GeomBuilder._buildGeometryCollection
  File "fiona\_geometry.pyx", line 141, in fiona._geometry.GeomBuilder._buildGeometryCollection
  File "fiona\_geometry.pyx", line 121, in fiona._geometry.GeomBuilder._buildParts
[...]

PR #1163, opened 2022-11-18
Support GeometryCollection containing TIN or Triangle geometries

rbuffat:
This PR adds support for GeometryCollections containing  [RFC64](https://gdal.org/development/rfc/rfc64_triangle_polyhedralsurface_tin.html) geometries.

Should fix https://github.com/Toblerity/Fiona/issues/1155 and https://github.com/Toblerity/Fiona/issues/1130


fiona/_geometry.pyx
```diff
@@ -143,20 +146,49 @@ cdef class GeomBuilder:
...
+        # We convert special geometries (curves, TIN, Triangle, ...)
+        # to GeoJSON compatible geometries (LineStrings, Polygons, MultiPolygon, ...)
+        if 8 <= code <= 14:  # Curves.
+            # OGR_G_GetLinearGeometry: The ownership of the returned geometry belongs to the caller.
+            cogr_geometry = OGR_G_GetLinearGeometry(geom, 0.0, NULL)
+            code = base_geometry_type_code(OGR_G_GetGeometryType(cogr_geometry))
+            ndims = OGR_G_GetCoordinateDimension(cogr_geometry)
+            geom = cogr_geometry
+        elif 15 <= code <= 17:
+            cloned_geometry = OGR_G_Clone(geom)
```
rbuffat, line 166: Without cloning the geometry, (respectively initially stealing the geometry from the feature the code segfaults). I'm not entirely sure yet why.

fiona/_geometry.pyx
```diff
@@ -143,20 +146,49 @@ cdef class GeomBuilder:
...
-        self.geomtypename = GEOMETRY_TYPES[self.code]
-        self.ndims = OGR_G_GetCoordinateDimension(geom)
+        self.code = code
+        self.geomtypename = GEOMETRY_TYPES[code]
+        self.ndims = ndims
         self.geom = geom
         built = getattr(self, '_build' + self.geomtypename)()
+
+        # Cleanup geometries we have ownership over
+        if cogr_geometry is not NULL:
```
rouault, line 189: It is probably not necessary, but it would be more reassuring to nullify self.geom as you are actually destroying it just after...
```suggestion
        if cogr_geometry is not NULL:
            self.geom = None
```

fiona/_geometry.pyx
```diff
@@ -143,20 +146,49 @@ cdef class GeomBuilder:
 
     cdef build(self, void *geom):
         # The only method anyone needs to call
-        if geom == NULL:
-            raise ValueError("Null geom")
-
-        cdef unsigned int etype = OGR_G_GetGeometryType(geom)
+        cdef void *cogr_geometry = NULL
+        cdef void *cloned_geometry = NULL
```
rouault, line 150: I would merge those 2 declarations in a single "geometry_to_dealloc" one

fiona/_geometry.pyx
```diff
@@ -143,20 +146,49 @@ cdef class GeomBuilder:
...
+                # OGR_G_ForceToMultiPolygon: the converted geometry (ownership to caller).
+                cogr_geometry = OGR_G_ForceToMultiPolygon(cloned_geometry)
+            elif code == 17:
+                # OGR_G_ForceToPolygon: the converted geometry (ownership to caller).
+                cogr_geometry = OGR_G_ForceToPolygon(cloned_geometry)
+            code = base_geometry_type_code(OGR_G_GetGeometryType(cogr_geometry))
+            ndims = OGR_G_GetCoordinateDimension(cogr_geometry)
+            geom = cogr_geometry
+        else:
+            ndims = OGR_G_GetCoordinateDimension(geom)
```
rouault, line 177: ndims = OGR_G_GetCoordinateDimension(geom) could be moved outside all branches of the if / elif / else blocks.

rbuffat: True. Neither OGR_G_GetLinearGeometry, OGR_G_ForceToMultiPolygon or OGR_G_ForceToPolygon should have an influence on the number of dimensions.

fiona/_geometry.pyx
rbuffat: Thanks. This explains it!

rbuffat:
@rouault On a feature level there is the possibility of using `OGR_F_StealGeometry` to take ownership of a geometry. I assume performance-wise `OGR_G_Clone` will be slower compared to `OGR_F_StealGeometry`. Is there an equivalent function on a Geometry level?

For a PolyhedralSurface, TIN or Triangle that are not part of a GeoemtryCollection we could just steal the geometry from the feature, but this would need some more refactoring. 

Thanks for your review, it is much appreciated!

fiona/_geometry.pyx
rouault: I meant that this call is repeated in the 3 branches, and could be just done once after them

rouault:
> there an equivalent function on a Geometry level?

you can use:
- OGRGeometryH OGR_G_GetGeometryRef( OGRGeometryH hGeom, int iSubGeom )  where hGeom is a container geometry,  and iSubGeom the index of the sub geometry inside it, to get a reference to the geometry
- and OGR_G_RemoveGeometry( OGRGeometryH hGeom, int iGeom, int bDelete ) where hGeom is a container geometry, iGeom the index of the sub geometry inside it and bDelete to be set to FALSE, so that the previously get reference isn't destroyed
The combination of the 2 is the equivalent of stealing a sub-geometry to a container geometry

fiona/_geometry.pyx
```diff
@@ -143,20 +145,59 @@ cdef class GeomBuilder:
...
+        # We convert special geometries (Curves, TIN, Triangle, ...)
+        # to GeoJSON compatible geometries (LineStrings, Polygons, MultiPolygon, ...)
+        if code in (
+            OGRGeometryType.CircularString.value,
+            OGRGeometryType.CompoundCurve.value,
+            OGRGeometryType.CurvePolygon.value,
+            OGRGeometryType.MultiCurve.value,
+            OGRGeometryType.MultiSurface.value,
+            OGRGeometryType.Curve.value,
+            OGRGeometryType.Surface.value,
```
rouault, line 165: Curve and Surface are purely abstract / non-instanciable geometry types
```suggestion
```

rbuffat:
I found some time to work on it. But I will need to go over it again to tidy up everything.

fiona/_geometry.pyx
```diff
@@ -101,66 +110,150 @@ cdef class GeomBuilder:
...
-            raise ValueError("Null geom")
 
-        cdef unsigned int etype = OGR_G_GetGeometryType(geom)
+    cdef object build_from_feature(self, void *feature):
+        # Build Geometry from *OGRFeatureH
+        cdef void *cogr_geometry = NULL
+        cdef int code
+
+        cogr_geometry = OGR_F_GetGeometryRef(feature)
+        code = base_geometry_type_code(OGR_G_GetGeometryType(cogr_geometry))
```
rbuffat, line 177: Currently, the code of a geometry is determined two times: here and in build(). This could be optimized by passing the code to build.  What is the equivalent in Cython of build(..., code: Optional[int] = None)? Passing an int pointer that is NULL when no code was previously computed?

sgillies: @rbuffat don't worry about optimizing that. Getting the code is practically free compared to other stuff we're doing.

sgillies:
@rbuffat the immediate cause of those warnings is that _geometry.pyx imports gdal.pxi and at https://github.com/Toblerity/Fiona/pull/1163/files#diff-71c8a7a6ae4cca6dd2e3e2e0e98232b4c2decc714d27ed6bc972c0ff5cc66608R11 _err.pyx is imported and gdal.pxi is re-imported. We're redeclaring, but since we're not redefining we're okay. No bugs have arisen because of this. We can try to restructure in the future, but it's fine to ignore it for now. I'm going to merge this.

PR #1165, opened 2022-11-22
Move project metadata to pyproject.toml; add build-backend

mwtoews:
This PR moves the project metadata to pyproject.toml ([PEP 621](https://peps.python.org/pep-0621/)), as supported by [setuptools>61](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html), and specifies "setuptools.build_meta" as the build-backend ([PEP 517](https://peps.python.org/pep-0517/)).

A few other things to note:

- The `--no-use-pep517` option raises "ERROR: Disabling PEP 517 processing is invalid: project specifies a build backend of setuptools.build_meta in pyproject.toml", so is removed from CI workflow. More CI reworkings might be expected later, this is the minimal change needed for now.
- Add -v flag for pip install to show Cython compilation details
- VERSION.txt is removed from the base package. Was it needed?
- A pytest option `--import-mode=importlib` seems to be needed; see [pytest docs](https://docs.pytest.org/en/6.2.x/pythonpath.html#import-modes) for what it is.
- Updates to `requirements-dev.txt` were also required, for instance the previous `pytest==6.2.5` was problematic

Changes to the metadata:

- Add Python 3.11 classifier
- Upgrade versions in the optional dependencies closer to the requirements*txt files (which I may consider removing sometime later)
- Add project URLs table with multiple links
- Removed @sgillies email, and added "Fiona contributors" as generic maintainer
- Change license from "BSD" to more specific "BSD 3-Clause"

Note that package_data was kept in setup.py, as it is dynamically evaluated based on environment variables.


mwtoews:
The "name" metadata is kept in setup.py (for now) to maintain GitHub's [dependency graph](https://github.com/Toblerity/Fiona/network/dependents), currently showing 9818 repositories / 308 packages.

mwtoews:
One last change is to include gdal.pxi and *.pxd files with the wheels, so that someone writing their Cython modules could use (e.g.) `from fiona.crs cimport CRS`

rbuffat:
@mwtoews What are the implications regarding backward compatibility when changing to pyproject.toml? Setuptools 61.0.0 seems to have been released less than a year ago.

mwtoews:
Good question @rbuffat, this could potentially impact a few folks that use old setuptools versions. I'm not sure who they are, but generally this is overcome by using venv with the latest components from PyPI. Another implication was with inclusion for Debian/Ubuntu, which has been [recently resolved with Debian testing](https://tracker.debian.org/news/1374817/setuptools-6530-11-migrated-to-testing/), so this shouldn't be a blocker for downstream packaging.

From what I read from Python devs, pyproject.toml is the final destination for project metadata (among a swarm of other tool configurations), so this PR is for the long-term gain.

pyproject.toml
```diff
@@ -1,7 +1,85 @@
 [build-system]
 requires = [
-    "cython>=0.29.29",
+    "Cython~=0.29",
```
sgillies, line 3: We need a fix introduced in 0.29.29.

```suggestion
    "Cython~=0.29.29",
```

pyproject.toml
```diff
@@ -1,7 +1,85 @@
...
+    "cligj>=0.5",
+    "munch>=2.3.2",
+]
+
+[project.optional-dependencies]
+all = ["Fiona[calc,s3,test]"]
+calc = ["shapely"]
+s3 = ["boto3>=1.3.1"]
+test = [
+    "Cython~=0.29",
```
sgillies, line 50:
```suggestion
    "Cython~=0.29.29",
```

sgillies (approved):
@mwtoews thank you!

pyproject.toml
```diff
@@ -1,7 +1,85 @@
...
+    "cligj>=0.5",
+    "munch>=2.3.2",
+]
+
+[project.optional-dependencies]
+all = ["Fiona[calc,s3,test]"]
+calc = ["shapely"]
+s3 = ["boto3>=1.3.1"]
+test = [
+    "Cython~=0.29.29",
```
jorisvandenbossche, line 50: Is there a reason that `test` needs cython?

sgillies: No, this is a leftover from old versions, I think.

sgillies: I removed it in 10ce6b7

PR #1180, opened 2023-01-22
Invalid creation options are filtered and ignored

### LINKED ISSUE #504: Avoid passing encoding creation option to format drivers that don't support it

sgillies:
Resolves #504 in a more general way than previously, taking care of all invalid creation options, not only `ENCODING`.

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -1180,6 +1180,9 @@ cdef class WritingSession(Session):
             geometry_code = geometry_type_code(geometry_type)
 
             try:
+                # In GDAL versions > 3.6.0 the following directive may
+                # suffice and we might be able to eliminate the import
+                # of fiona.meta in a future version of Fiona.
                 with Env(GDAL_VALIDATE_CREATION_OPTIONS="NO"):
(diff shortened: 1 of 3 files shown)


fiona/ogrext.pyx
```diff
@@ -1131,40 +1146,45 @@ cdef class WritingSession(Session):
...
                 geometry_type = "Unknown"
+
             geometry_code = geometry_type_code(geometry_type)
 
             try:
-                self.cogr_layer = exc_wrap_pointer(
-                    GDALDatasetCreateLayer(
-                        self.cogr_ds, name_c, cogr_srs,
-                        <OGRwkbGeometryType>geometry_code, options))
+                with Env(GDAL_VALIDATE_CREATION_OPTIONS="NO"):
```
sgillies, line 1183: This directive might eventually take care of everything, but only works in GDAL 3.6.0 (See https://github.com/OSGeo/gdal/pull/6488).

PR #1182, opened 2023-01-22
Don't use deprecated distutils module.

sebastic:
`setup.py` still uses the deprecated distutils which will be removed in Python 3.12.

From [What’s New In Python 3.10](https://docs.python.org/3/whatsnew/3.10.html#distutils-deprecated):
> The entire `distutils` package is deprecated, to be removed in Python 3.12. Its functionality for specifying package builds has already been completely replaced by third-party packages `setuptools` and `packaging`, and most other commonly used APIs are available elsewhere in the standard library (such as [platform](https://docs.python.org/3/library/platform.html#module-platform), [shutil](https://docs.python.org/3/library/shutil.html#module-shutil), [subprocess](https://docs.python.org/3/library/subprocess.html#module-subprocess) or [sysconfig](https://docs.python.org/3/library/sysconfig.html#module-sysconfig)). There are no plans to migrate any other functionality from distutils, and applications that are using other functions should plan to make private copies of the code. Refer to [PEP 632](https://peps.python.org/pep-0632/) for discussion.

[Porting from Distutils](https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html#prefer-setuptools) suggests using [`logging`](https://docs.python.org/3/library/logging.html) instead of `distutils.log`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
-from distutils import log
+import logging
 import os
 import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


sgillies:
@sebastic thank you! I'm trying to remind myself why I was using the distutils logger. I think it was because of incompatibilities between that and the stdlib logging system. As far as I know, setuptools (still used by Fiona instead of distutils) now uses the Python logging system.

sebastic:
>  As far as I know, setuptools (still used by Fiona instead of distutils) now uses the Python logging system.

It does: https://github.com/pypa/setuptools/pull/2974

sgillies (approved):
Nevermind the failure, it's a different issue. Thanks @sebastic !

mwtoews:
@sgillies is it too late to cherry-pick this to `maint-1.9`?

sgillies:
@mwtoews probably not. It only affects those of us building from source if it affects anyone at all. I'll do it if you'll help me test :smile:

jorisvandenbossche:
It might actually be worth backporting, given that Python 3.12 is not that far away, and I expect that fiona 1.9 might still be used for some time?

mwtoews:
cherry-picked to `maint-1.9` with e344a16ce6af6b058fd8e4a5c3848f895ee85e8b

I've done a local build/install with this commit, and it seems to be clean enough to not need another beta release.

---

Also, a while back I started a setuptools discussion to see if there are "[best practices for logging in setup.py](https://github.com/pypa/setuptools/discussions/3710)". So far, it doesn't really seem so.

PR #1186, opened 2023-01-29
Run pyupgrade --py37-plus

mwtoews:
This is a 99% automated edit using [pyupgrade](https://github.com/asottile/pyupgrade):

    pyupgrade --py37-plus `find . -name "*.py"`

Despite the large number of changes, I don't expect any surprises here as the logic is the same. Most automated edits are removing stale code required for Python 2, or upgrading formatters to use f-strings or old % formatters to more modern styles. The code is guaranteed to run with a minimum Python 3.7 version.

### DIFF
--- a/benchmark-max.py
+++ b/benchmark-max.py
@@ -1,4 +1,3 @@
-
 import timeit
 from fiona import collection
 from osgeo import ogr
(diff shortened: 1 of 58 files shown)


sgillies (approved):
Thanks @mwtoews !

PR #1193, opened 2023-01-31
TST: don't rely on exact order in listdir tests

jorisvandenbossche:
On the conda-forge recipe, those tests were failing because it gave a different order (https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=650034&view=logs&j=4f922444-fdfe-5dcf-b824-02f86439ef14&t=b2a8456a-fb11-5506-ca32-5ccd32538dc0&l=3626). 
I don't know if that's platform or GDAL-version dependent, but it seems OK to not rely on the exact order in the tests. 

Let me know if the PR needs to be targeted against master or maint-1.9

### DIFF
--- a/tests/test_listing.py
+++ b/tests/test_listing.py
@@ -84,7 +84,9 @@ def test_listing_pathobj(path_coutwildrnp_json):
 
 def test_listdir_path(path_coutwildrnp_zip):
     """List directories in a path"""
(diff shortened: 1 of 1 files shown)


sgillies:
@jorisvandenbossche I'll cherry pick this for 1.9.1.

PR #1201, opened 2023-02-09
Log a warning message when identically named fields are found

### LINKED ISSUE #1178: Reading a layer having multiple properties with the same name

When trying to read a layer for which multiple properties have the same name, only the last one is loaded in the returned Collection.

This is definitely bad data format, but GDAL correctly displays all the properties even if they have the same name.

QGIS does append a suffix ("prop", "prop_1", "prop_2", ...).

I think the issue comes from python dictionary with constraints on uniqueness of the keys.

I would have expected to at least get a warning, or better to have the same behavior than in QGIS.

The data comes from the AIRBUS GeoStore.

```bash
wget "https://datadoors.intelligence-airbusds.com/export/v1/static/export/export-20230106-11165137.zip"
unzip -o "export-20230106-11165137.zip" -d "geostore_products"
```

With GDAL I can see all the properties:

```bash
$ ogrinfo geostore_products Export
...
geometry: String (254.0)
geometry: Integer64 (18.0)
geometry: Integer64 (18.0)
geometry: Integer64 (18.0)
geometry: Integer64 (18.0)
...
```

With fiona I can only see one:

```python
import fiona

fpath = "geostore_products"
c = fiona.open(fpath, layer="Export")

# Only one geometry property
print([p for p in c.schema["properties"].keys() if p.startswith("geo")])  # ['geometry']

# And this is not the interesting one
print(c.schema["properties"]["geometry"])  # int:18
```

## Attempt at tracking down the issue

I suspect that the other "geometry" properties were overridden during translation as a python dictionary, since keys must be unique.

I can trace back that this is true starting from the following line in [`fiona/collection.py::Collection.__init__#L236`](https://github.com/Toblerity/Fiona/blob/master/fiona/collection.py#L236):

```python
self.session.start(self, **kwargs)
```

which leads to the issue being located in one of the two following lines in `fiona/ogrext.pyx::Session.start` ([#L572](https://github.com/Toblerity/Fiona/blob/91c13ad8424641557a4e5f038f255f9b657b1bc5/fiona/ogrext.pyx#L572) or [#L577](https://github.com/Toblerity/Fiona/blob/91c13ad8424641557a4e5f038f255f9b657b1bc5/fiona/ogrext.pyx#L577))

```python
self.cogr_ds = gdal_open_vector(path_c, 0, drivers, kwargs)        #L572
self.cogr_layer = GDALDatasetGetLayerByName(self.cogr_ds, name_c)  #L577
```

if #L572 this further comes down to [`fiona/ogrtext.pyx::gdal_open_vector#L134`](https://github.com/Toblerity/Fiona/blob/91c13ad8424641557a4e5f038f255f9b657b1bc5/fiona/ogrext.pyx#L134) with the following call:

```python
cogr_ds = exc_wrap_pointer(
    GDALOpenEx(path_c, flags, <const char *const *>drvs, <const char *const *>open_opts, NULL)
)
```

So this is either `GDALOpenEx` or `GDALDatasetGetLayerByName`, which come from `include "gdal.pxi"`.

I can find them there, but my understanding of the code stops here:

- [`gcore/gdaldataset.cpp::GDALOpenEx#L3231`](https://github.com/OSGeo/gdal/blob/master/gcore/gdaldataset.cpp#L3231)
- [`gcore/gdaldataset.cpp::GDALDatasetGetLayerByName#L4380`](https://github.com/OSGeo/gdal/blob/master/gcore/gdaldataset.cpp#L4380)

## Operating system

Ubuntu 20.04.5 LTS x86_64

```bash
$ conda list
# Name                    Version                   Build  Channel
fiona                     1.8.22          py311h3f14cef_5    conda-forge
gdal                      3.6.1           py311hadb6153_2    conda-forge
libgdal                   3.6.1                he31f7c0_2    conda-forge
```

sgillies:
Resolves #1178

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -608,16 +608,26 @@ cdef class Session:
         self.collection = collection
 
         if self.collection.include_fields is not None:
-            self.collection.ignore_fields = list(set(self.get_schema()["properties"]) - set(collection.include_fields))
+            self.collection.ignore_fields = list(
+                set(self.get_schema()["properties"]) - set(collection.include_fields)
+            )
+
         if self.collection.ignore_fields:
             try:
                 for name in self.collection.ignore_fields:
                     try:
                         name_b = name.encode(encoding)
                     except AttributeError:
-                        raise TypeError("Ignored field \"{}\" has type \"{}\", expected string".format(name, name.__class__.__name__))
-                    ignore_fields = CSLAddString(ignore_fields, <const char *>name_b)
+                        raise TypeError(
+                            "Ignored field \"{}\" has type \"{}\", expected string".format(
+                                name, name.__class__.__name__
+                            )
+                        )
+                    else:
+                        ignore_fields = CSLAddString(ignore_fields, <const char *>name_b)
+
                 OGR_L_SetIgnoredFields(self.cogr_layer, <const char**>ignore_fields)
+
             finally:
                 CSLDestroy(ignore_fields)
 
@@ -696,12 +706,27 @@ cdef class Session:
         return driver_name.decode()
 
     def get_schema(self):
+        """Get a dictionary representation of a collection's schema.
+
+        The schema dict contains "geometry" and "properties" items.
+
+        Returns
+        -------
+        dict
+
+        Warnings
+        --------
+        Fiona 1.9 does not support multiple fields with the name
+        name. When encountered, a warning message is logged and the
+        field is skipped.
+
+        """
         cdef int i
-        cdef int n
+        cdef int num_fields
         cdef void *cogr_featuredefn = NULL
         cdef void *cogr_fielddefn = NULL
         cdef const char *key_c
-        props = []
+        props = {}
 
         if self.cogr_layer == NULL:
             raise ValueError("Null layer")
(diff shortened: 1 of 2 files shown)


fiona/ogrext.pyx
```diff
@@ -696,12 +706,27 @@ cdef class Session:
...
+        -------
+        dict
+
+        Warnings
+        --------
+        Fiona 1.9 does not support multiple fields with the name
+        name. When encountered, a warning message is logged and the
+        field is skipped.
+
+        """
```
sgillies, line 723: I took the opportunity to make get_schema more readable. We'll be making future changes to it.

fiona/ogrext.pyx
```diff
@@ -712,30 +737,46 @@ cdef class Session:
...
 
+            # See gh-1178 for an example of a pathological collection
+            # with multiple identically name fields.
+            if key in props:
+                log.warning(
+                    "Field name collision detected, field is skipped: i=%r, key=%r",
+                    i,
+                    key
+                )
+                continue
```
sgillies, line 776: This preserves existing behavior, field skipping, but adds a warning level log message. I decided against raising a warning, because there's nothing you can do about the fields and schema if you were to handle the warning.

PR #1220, opened 2023-03-19
Use importlib.metadata instead of pkg_resources

### LINKED ISSUE #1219: drop setuptools - replace pkg_resources with importlib.metadata
Hi there, this is a minor annoyance I've stumbled upon these days...

Using [jake](https://jake.readthedocs.io/en/latest/) to detect vulnerabilities in my dependencies, it warns me about a [medium vulnerability](https://ossindex.sonatype.org/vulnerability/sonatype-2014-0148) - that can be ignored in a secure enough environment, but makes my CI unhappy :sweat_smile: 

The official [setuptools documentation](https://setuptools.pypa.io/en/latest/pkg_resources.html) suggests to dismiss `pkg_resources` in favor of `importlib.metadata` anyway.

In `fiona`'s specific case, its just a matter of replacing

```python
from pkg_resources import iter_entry_points

for ep in iter_entry_points("fiona.fio_commands"):
    ...
```

with

```python
from itertools.metedata import entry_points


for ep in entry_points(group="fiona.fio_commands"):
    ...
```

For the [click decorator](https://github.com/Toblerity/Fiona/blob/44acaf01e64257ca25b12286ff05018563223492/fiona/fio/main.py#L25-L27), I would also suggest to use

```python
import itertools

itertools.chain(entry_points(group="fiona.fio_commands"), entry_points(group="fiona.fio_plugins"))
```

If python 3.7 support is still needed, one can use the [`importlib-metadata` backport](https://pypi.org/project/importlib-metadata/).

sgillies:
Resolves #1219

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,6 +39,7 @@ dependencies = [
     "click~=8.0",
     "click-plugins>=1.0",
     "cligj>=0.5",
+    "importlib-metadata",
     "munch>=2.3.2",
     "setuptools",
 ]
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -6,6 +6,7 @@ All issue numbers are relative to https://github.com/Toblerity/Fiona/issues.
 1.9.2 (TBD)
 -----------
 
+- Get command entry points using importlib.metadata (#).
 - Instead of warning, transform_geom() raises an exception when some points
   can't be reprojected unless the caller opts in to partial reprojection. This
   restores the behavior of version 1.8.22.
--- a/fiona/__init__.py
+++ b/fiona/__init__.py
@@ -126,7 +126,7 @@ __all__ = [
     "prop_width",
 ]
 
-__version__ = "2.0dev"
+__version__ = "1.9.2dev"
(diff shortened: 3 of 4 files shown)


pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
     "click~=8.0",
     "click-plugins>=1.0",
     "cligj>=0.5",
+    "importlib-metadata",
     "munch>=2.3.2",
     "setuptools",
```
jorisvandenbossche, line 44: Can setuptools be dropped now, or are there still other cases where setuptools is used during runtime?

sgillies: I'll see and will remove it if possible.

pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
     "click~=8.0",
     "click-plugins>=1.0",
     "cligj>=0.5",
+    "importlib-metadata",
```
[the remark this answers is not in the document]
sgillies: Until Python 3.10 the importlib.metadata API was in flux, so I'd want to use it for <3.10.

PR #1221, opened 2023-03-20
Fix or refactor string formatting

mwtoews:
This PR started off as a one line fix for `"Driver {} does not support virtual files."` (incomplete message), but it turns out other string formatting modernisation could make things nicer.

- Convert e.g. `"{x}".format(x=x)` to `f"{x}"`
- Convert old `%s` formatting to f-strings
- Convert some `{}` formatting for logging messages to `%s`
- Simplify `show_versions()` with basic `print()` functions

Some of these edits are automated using [flynt](https://github.com/ikamensh/flynt), others are manual (particularly `.pyx` files).


fiona/crs.py
```diff
@@ -64,7 +64,7 @@ def from_epsg(code):
     """
     if int(code) <= 0:
         raise ValueError("EPSG codes are positive integers")
-    return {'init': "epsg:%s" % code, 'no_defs': True}
+    return {"init": f"epsg:{code}", "no_defs": True}
```
sgillies, line 67: @mwtoews for the first time I realize that we have a crs Cython module and a crs Python module in the master branch! Let's not do anything about that here, this is a note to fix the situation in another PR.

mwtoews: This change is removed and issue raised at #1226

PR #1222, opened 2023-03-20
Remove benchmark scripts

mwtoews:
These Python 2 files have not been maintained. Should they? If not, they could be cleaned up.


sgillies:
Thanks @mwtoews !

PR #1218, opened 2023-03-07
Docker & conda tests

sgillies:
Based on Rasterio's tests. I'm focusing on getting them running first, then will see what we're lacking in comparison to Fiona's current tests.

Resolves #1206 not by migrating to the Ubuntu 20.04 runner, but by using GDAL and PROJ docker images and Conda Forge.

PR #1225, opened 2023-03-21
Extend CI test matrix

rbuffat:
See comments in https://github.com/Toblerity/Fiona/pull/1218#pullrequestreview-1350899043


rbuffat:
@sgillies  Fiona does not have any numpy dependency that I'm aware of.  I assume this is a copy-paste artifact from rasterio?

.github/workflows/tests.yaml
```diff
@@ -86,10 +90,10 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        os: [macos-latest]
-        python-version: ['3.9', '3.10']
```
rbuffat, line 90: I assume that there should not be any python specific issues on OS X. The different Python versions are already tested with the Docker tests.

.github/workflows/tests.yaml
```diff
@@ -86,10 +90,10 @@ jobs:
...
         include:
-          - os: ubuntu-latest
+          - os: macos-latest
+            python-version: '3.10'
+          - os: windows-latest
```
rbuffat, line 96: As we already have Linux tests with the Docker tests, it would be more interesting to test Windows here. Not sure how easy this is though.

rbuffat:
@sgillies It looks as if there are no appveyor checks showing up, but the configuration is still there. Is this unexpected or should the appveyor config be removed?

mwtoews:
The AppVeyor builds are running [here](https://ci.appveyor.com/project/sgillies/fiona/builds/46568549), but I guess some GitHub authorizations are needed to signal the check result back to the PR.

rbuffat:
@mwtoews  Thanks for checking. It looks as if they are broken.

sgillies:
Let's remove the AppVeyor config, but in a different PR.

.github/workflows/tests.yaml
```diff
@@ -99,29 +103,43 @@ jobs:
...
+      - name: Install Env (Windows)
+        if: matrix.os == 'windows-latest'
+        shell: bash
+        run: |
+          conda config --prepend channels conda-forge
+          conda config --set channel_priority strict
+          conda create -n test python=${{ matrix.python-version }} libgdal geos=3.10.3 cython=0.29
+          source activate test
+          GDAL_VERSION="3.5" python setup.py build_ext -I"C:\\Miniconda\\envs\\test\\Library\\include" -lgdal_i -L"C:\\Miniconda\\envs\\test\\Library\\lib" bdist_wheel
+          python -m pip install --ignore-installed dist\\Fiona-2.0.dev0-cp310-cp310-win_amd64.whl
```
sgillies, line 126: Maybe use a file glob on the wheel so that we're independent of the version?

rbuffat:
@sgillies I had a quick look at the new CI config and one thing led to another:

- Extending the Linux CI to all supported Python versions is a personal preference.  Most of Fiona's code should not really depend on the Python version, although dependencies might (and also e.g. type annotations support changed over time)

- As the conda tests are comparably slow I thought it would be more "interesting" to test on Windows instead of on Linux. For the same reason also to reduce the tests to just one per platform. Maybe it would make sense to just test the newest supported Python version to detect issues like the DLL loading on Windows? Or test the oldest and newest supported Python version for each platform?

- A while ago there was an issue that the rstcheck check broke due to unresolvable dependencies.  Pip-compile takes it to the next level by pinning all the dependencies to ensure the same environment with each test run. Enabling dependabot for the rstcheck dependencies ensures that rstcheck is not forgotten. It is configured to run once a month, this should keep the maintenance overhead minimal.

In case you don't like any of these changes I can revert them or you can just pick the changes you like.

rbuffat:
Numpy dropped support for Python 3.7, we could also drop the tests for Python 3.7.

mwtoews:
I suggest to keep the Python versions in this PR as they are now, and drop Python 3.7 in a separate PR against `master` (as it could have other non-CI changes). It would be best to keep Python 3.7 for `maint-1.9`. I can create a PR if you want.

sgillies (approved):
Thanks @rbuffat !

Issue #57, opened 2013-08-13
Ordered Field Names


dennytron:
Hi there! I read through the documents to see if there is a way to return an ordered list of the field names in a schema. Currently, the schema properties are returned as a dictionary, so the field names are returned out of order. An ordered list of fields is important when testing to see if a schema matches a specification (order, field widths, etc).

Thanks!
Dennis

PR #1235, opened 2023-04-02
Change schema properties from OrderedDict back to dict built-in

### LINKED ISSUE #1094: Replace OrderedDict with dict
This maintenance enhancement would remove internal use of [`collections.OrderedDict`](https://docs.python.org/3/library/collections.html#collections.OrderedDict) with [`dict`](https://docs.python.org/3/library/stdtypes.html#dict), which is easier to use and has a much more compact string representation (among [other things](https://www.youtube.com/watch?v=p33CVV29OG8)).

The history why OrderedDict was used is because older Python versions (before CPython 3.6) had non-ordered `dict` objects.

mwtoews:
Since Python 3.7 (and CPython 3.6), the `dict` built-in preserves insertion order, so there is no longer any need to use OrderedDict for feature properties in Fiona.

This PR undoes 2d51d932bcab4f2b20ff4bbc7c8a802960d85c88 discussed in #57 for Fiona 1.0.1.

Closes #1094

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -297,8 +297,7 @@ Features of a collection may also be accessed by index.
                                    (-4.663611, 51.158333)]],
                   'type': 'Polygon'},
      'id': '1',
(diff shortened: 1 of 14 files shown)


sgillies (approved):
Thanks @mwtoews !

PR #1237, opened 2023-04-03
Append mode for fio-load

### LINKED ISSUE #1236: Allow append, mode="a", for fio-load
I think it's mostly an oversight that fio-load only opens datasets in write mode. Different schemas will complicate things, but if you want to add features that are compatible with the existing dataset, and the format allows appending, there shouldn't be any problem.

sgillies:
Resolves #1236

### DIFF
--- a/fiona/fio/load.py
+++ b/fiona/fio/load.py
@@ -30,9 +30,22 @@ from fiona.transform import transform_geom
     "zero-based numbering when accessed by index.",
 )
 @options.creation_opt
+@options.open_opt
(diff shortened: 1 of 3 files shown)


fiona/fio/load.py
```diff
@@ -81,14 +94,19 @@ def feature_gen():
...
+            output,
+            "w",
+            driver=driver,
+            crs=dst_crs,
+            schema=schema,
+            layer=layer,
+            **creation_options
+        )
+
+    with opener as dst:
```
sgillies, line 110: I love how nice this is in Python!

PR #1246, opened 2023-04-05
Modernize the manual

### LINKED ISSUE #1245: Change examples in manual to use web-hosted coutwildrnp.zip
Like this:

Screenshot from 2023-04-04 13-13-51

sgillies:
Most of the work is switching over to using coutwildrnp.zip in the examples.

Resolves #1245


sgillies:
Here's the proposed new manual https://fiona--1246.org.readthedocs.build/en/1246/manual.html.

It's not entirely in sync with your work on the master branch @mwtoews , but not incompatible. My intention is to make it more approachable by 1) using a web-hosted dataset as a feature source in examples, and 2) using local GeoPackage files as the preferred output format.

Issue #1125, opened 2022-07-08
Add typing hints


rbuffat:
Recently it seems that more and more Python libraries include typing hints. Personally, I'm a fan of type hints while developing. Are there any plans to add typing hints to Fiona? 

I'm not 100% sure what the best way to deal with the dynamic nature of the dataset schemas is. Maybe the implementation of static typing of psycopg3 could be used as inspiration: https://www.psycopg.org/psycopg3/docs/advanced/typing.html

sgillies:
There wasn't much point in doing this before the Feature, Geometry, Properties classes. They will help quite a bit. Maybe a Schema class would, too?

StefanBrand:
In a naive attempt to model the parameters to `fiona.open` (v1.8) we created this:

```py
from typing import TypedDict

class FionaBaseDict(TypedDict):
    properties: dict[str, Any]


class FionaSchemaWithoutGeometry(FionaBaseDict):
    geometry: None


class FionaSchemaWithGeometry(FionaBaseDict):
    geometry: str


FionaSchema = Union[FionaSchemaWithoutGeometry, FionaSchemaWithGeometry]


class RecordGeometry(TypedDict):
    type: str
    coordinates: tuple


class RecordWithoutGeometry(FionaBaseDict):
    geometry: None


class RecordWithGeometry(FionaBaseDict):
    geometry: RecordGeometry


Record = Union[RecordWithoutGeometry, RecordWithGeometry]
```

I'm not yet familiar with the new Feature, Geometry, Properties classes. That would probably simplify things. In any case I would be interested to help get type hints into `fiona`, either by sharing opinion or contributing code.

sgillies:
@StefanBrand @rbuffat what would you think about collaborating on [stubs](https://typing.readthedocs.io/en/latest/source/stubs.html) for 1.9 and then possibly moving the stubs into the code at 2.0?

rbuffat:
@sgillies I can have a go at it but probably during the next few weeks I don't have too much time available, thus it can take a while.  I will also need to dive a bit deeper into the Python typing system and maybe look at how other projects are implementing dynamic types (Sqlalchemy or pydantic come to my mind here, perhaps it would also be worth asking somebody familiar with mypy/pyright for some inputs). Being able to static type check `writerecords()` or the collection iterator would probably be the most useful thing, but I guess also the hardest to implement.

StefanBrand:
> but probably during the next few weeks I don't have too much time available

[...]

PR #1259, opened 2023-05-05
Add type stubs

StefanBrand:
Ref #1125 

- Initial type stubs were generated using [Instagram/MonkeyType](https://github.com/instagram/MonkeyType)
- Manual fixes were applied until there were no errors
- For mypy, some dependencies are ignored in `pyproject.toml`
- Pipeline was added to Github Actions

The type stubs still contain many `Any`s. Now we would really like to gather feedback to make the type annotations stricter, especially around the important places:

> Being able to static type check `writerecords()` or the collection iterator would probably be the most useful thing


fiona/fio/helpers.pyi
```diff
@@ -0,0 +1,27 @@
...
+    List,
+    Optional,
+    Tuple,
+    Union,
+)
+from fiona.model import Geometry
+
+
+def eval_feature_expression(
+    feature: Dict[str, Union[Dict[str, Union[str, List[List[List[float]]]]], str, Dict[str, Optional[Union[float, str, int]]]]],
```
totycro, line 15: For complex combinations of basic types, it's probably useful to introduce names for (sub)expressions.

I like the style that the httpx library uses:
https://github.com/encode/httpx/blob/master/httpx/_client.py#L1111

fiona/rfc3339.pyi
```diff
@@ -0,0 +1,26 @@
...
+
+
+def parse_datetime(
+    text: str
+) -> Union[Tuple[int, int, int, int, int, int, int, float], Tuple[int, int, int, int, int, int, int, None], Tuple[int, int, int, int, int, int, int, int]]: ...
+
+
+def parse_time(
+    text: str
+) -> Union[Tuple[int, int, int, int, int, int, int, float], Tuple[int, int, int, int, int, int, int, None], Tuple[int, int, int, int, int, int, int, int]]: ...
```
totycro, line 21: This would also benefit from at least an alias.

sgillies:
@StefanBrand thank you for explaining the approach! I had not heard about MonkeyType.

Did adding type stubs find any Fiona bugs?

It will be a while before I can give my full attention to a review of this. The Fiona 1.9.4 release and some wheel building updates are higher priority for me, as are the next Rasterio release and some Shapely project governance work.

StefanBrand:
@sgillies Under my impression the type stubs derived from the tests by MonkeyType are still quite broad. We would now have to make the types stricter here and there to catch bugs if there are any. My approach would be to more correctly represent the `__geo_interface__` / GeoJSON spec. We can take inspiration from https://github.com/developmentseed/geojson-pydantic, as explained in this comment: https://github.com/jazzband/geojson/issues/167#issuecomment-1329499017

PR #1263, opened 2023-05-10
Encode bytes type feature properties using base64

### LINKED ISSUE #1262: Fiona allows bytes type properties but can't serialize them
Should be a fairly simple fix for 1.9.4.

sgillies:
Resolves #1262, which I discovered when trying to convert https://github.com/OSGeo/gdal/blob/master/autotest/ogr/data/parquet/test.parquet to GeoJSON. This bug has been lurking for a while.

@mwtoews what do you think?

@jorisvandenbossche what's the pyogrio approach to serializing bytes in GeoJSON?

### DIFF
--- a/fiona/model.py
+++ b/fiona/model.py
@@ -1,6 +1,6 @@
 """Fiona data model"""
 
-from base64 import b64encode
+from binascii import hexlify
 from collections.abc import MutableMapping
 from collections import OrderedDict
 from enum import Enum
@@ -382,7 +382,7 @@ class ObjectEncoder(JSONEncoder):
                 o_dict["properties"] = self.default(o.properties)
             return o_dict
         elif isinstance(o, bytes):
-            return b64encode(o)
+            return hexlify(o)
         else:
             return o
 
--- a/fiona/model.py
+++ b/fiona/model.py
@@ -1,9 +1,10 @@
 """Fiona data model"""
 
-import itertools
-from collections import OrderedDict
+from base64 import b64encode
 from collections.abc import MutableMapping
+from collections import OrderedDict
 from enum import Enum
+import itertools
 from json import JSONEncoder
 from warnings import warn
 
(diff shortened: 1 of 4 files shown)


fiona/model.py
```diff
@@ -367,21 +368,23 @@ def from_dict(cls, mapping=None, **kwargs):
 
 
 class ObjectEncoder(JSONEncoder):
-    """Encodes Geometry and Feature"""
+    """Encodes Geometry, Feature, and Properties."""
 
     def default(self, o):
         if isinstance(o, (Geometry, Properties)):
-            return {k: v for k, v in o.items() if v is not None}
+            return {k: self.default(v) for k, v in o.items() if v is not None}
```
sgillies, line 375: Here's where different types of properties will be handled.

fiona/model.py
```diff
@@ -367,21 +368,23 @@ def from_dict(cls, mapping=None, **kwargs):
...
+                o_dict["geometry"] = self.default(o.geometry)
             if o.properties is not None:
-                o_dict["properties"] = ObjectEncoder().default(o.properties)
+                o_dict["properties"] = self.default(o.properties)
             return o_dict
+        elif isinstance(o, bytes):
+            return b64encode(o)
         else:
-            return JSONEncoder().default(o)
+            return o
```
sgillies, line 387: Necessitated by the change at line 375

jorisvandenbossche:
> what's the pyogrio approach to serializing bytes in GeoJSON?

That's a good question, I don't think we already support this. 

We don't have custom JSON serialization, and so only support it when writing using the GeoJSON driver. And our writing code starts from numpy arrays, and anything that is not in the [list of mapped dtypes](https://github.com/geopandas/pyogrio/blob/6b07e7d8a150f341592a9330760b5aa9fdeb6ce3/pyogrio/_io.pyx#L60-L85) will either [fall back to using OTFString type](https://github.com/geopandas/pyogrio/blob/6b07e7d8a150f341592a9330760b5aa9fdeb6ce3/pyogrio/_io.pyx#L1443-L1452) (when it is numpy object dtype) or raise a not-implemented error. 
When starting from a geopandas dataframe, bytes will be stored in object dtype, and thus we currently write a "string repr" of the python bytes object (`"binary": "b'\\x00\\x01'"`, not particularly useful IMO)

When using ogr2ogr, it raises a warning about it: *Warning 1: The output driver does not natively support Binary type for field binary. Misconversion can happen. -mapFieldType can be used to control field type conversion.*, and then the result actually shows `"binary": "0001"` (which is the hex encoded version, I think?)

Checking with a quick patch to pyogrio to write a column of bytes to `OTFBinary` using `OGR_F_SetFieldBinary` just by passing the python bytes object, it seems that GDAL automatically does this? It also results in a GeoJSON file with `"binary": "0001"` (although without raising the warning, so that warning seems to be custom in ogr2ogr)

 I assume that the GeoJSON driver will call `GetFieldAsString` on each Field value, and this calls `CPLBinaryToHex` in case of `OTFBinary` type.

So in summary, I think if we would add support in pyogrio, we would let GDAL do its thing, and thus use hex encoding and not base64 encoding.

PR #1264, opened 2023-05-10
BUG: Handle getting geometry type with NULL geometries

snowman2:
Since this change #1163, this started showing up in the logs recently for a CSV file without a geometry:
```python
>>> fds = fiona.open("file.csv")
>>> next(iter(fds))
ERROR 10: Pointer 'hGeom' is NULL in 'OGR_G_GetGeometryType'.
```

Note: `OGR_G_GetGeometryType` returns 0 if NULL, so it works as is. Just adds a lot of error messages that can make you think there is an issue when there isn't one.


sgillies (approved):
Great! Thanks @snowman2

PR #1272, opened 2023-05-23
Exclude py files from docs, examples, scripts, tests from wheels

sgillies:


pyproject.toml
```diff
@@ -83,5 +83,5 @@ include-package-data = false
 version = {attr = "fiona.__version__"}
 readme = {file = ["README.rst", "CHANGES.txt", "CREDITS.txt"]}
 
-[tool.setuptools.packages]
-find = {}
+[tool.setuptools.packages.find]
+exclude = ["docs*", "examples*", "scripts*", "tests*", "venv*"]
```
mwtoews, line 87: An alternative is to use `include`:
```suggestion
include = ["fiona", "fiona.*"]
```
(`exclude` would be used to exclude things within the `include`)

mwtoews: `fiona._vendor` should be captured by `fiona.*`. It's worth confirming in a local build. Also, be sure to clean up `rm -rfv *.egg-info`, since older attempts may influence package results.

sgillies: @mwtoews the include route wasn't working for me before 1.9.4 but I'll try it again. I'm inclined to publish the 1.9.4.post1 wheels I've built as they are since they're equivalent to what would have been built by including rather than excluding. Okay with that?

mwtoews: I can confirm that `include` works fine, my guess why it didn't work before was influence from older `*.egg-info` which should be cleaned-up before builds.

sgillies: @mwtoews I think you're right. After ensuring that older egg info was deleted, including "fiona" and "fiona.*" seems sufficient :+1:

PR #1276, opened 2023-06-20
Preserve the null properties and geometry of a Feature

### LINKED ISSUE #1270: Unable to do round trip IO with GPKG files in geopandas

I have using GeoDataFrame objects in a project that involves IO and the GPKG format and I have come across a potential bug. geopandas uses fiona for IO, so I thought this would be the proper forum.

Previously (with Fiona 1.8), I was able to write a file with a GeoDataFrame column of Nones and read it back as Nones too.

Now (with Fiona 1.9.3), I read that file as NaN, which requires me to do pre-processing before writing the file. The previous behaviour makes more sense to me but perhaps there is a reason for this change.

Is this a bug or was something changed intentionally?

The issue can be reproduced as follows:

```
from geopandas import GeoDataFrame, read_file
from shapely.geometry import Point
gdf = GeoDataFrame(
    {
      'data': [None,None],
      'new': [3.4,4.5], 
      'geometry': [Point(1, 2), Point(3,4)],
      },
    index=['a','b']
    )
gdf.to_file('test.gpkg')
gdf2 = read_file('test.gpkg')
print(gdf)
print(gdf2)
```

And the output is:

```console
   data  new                 geometry
a  None  3.4  POINT (1.00000 2.00000)
b  None  4.5  POINT (3.00000 4.00000)
  index  data  new                 geometry
0     a   NaN  3.4  POINT (1.00000 2.00000)
1     b   NaN  4.5  POINT (3.00000 4.00000)
```

## Operating system

OS: Debian Bookworm
Python: 3.11.2

Fiona: 1.9.3 (via pip 23.0.1)
geopandas: 0.13.0

sgillies:
Resolves #1270

### DIFF
--- a/fiona/model.py
+++ b/fiona/model.py
@@ -384,15 +384,15 @@ class ObjectEncoder(JSONEncoder):
     """Encodes Geometry, Feature, and Properties."""
 
     def default(self, o):
-        if isinstance(o, (Geometry, Properties)):
-            return {k: self.default(v) for k, v in o.items() if v is not None}
-        elif isinstance(o, Feature):
-            o_dict = dict(o)
-            o_dict["type"] = "Feature"
-            if o.geometry is not None:
-                o_dict["geometry"] = self.default(o.geometry)
-            if o.properties is not None:
-                o_dict["properties"] = self.default(o.properties)
+        if isinstance(o, Object):
+            o_dict = {k: self.default(v) for k, v in o.items()}
+            if isinstance(o, Geometry):
+                if o.type == "GeometryCollection":
+                    _ = o_dict.pop("coordinates", None)
+                else:
+                    _ = o_dict.pop("geometries", None)
+            elif isinstance(o, Feature):
+                o_dict["type"] = "Feature"
             return o_dict
         elif isinstance(o, bytes):
             return hexlify(o)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,13 +1,14 @@
 [build-system]
 requires = [
-    "Cython~=0.29.29",
+    "cython~=3.0.2",
     "oldest-supported-numpy",
-    "setuptools>=61",
+    "setuptools>=67.8",
+    "wheel",
 ]
 build-backend = "setuptools.build_meta"
 
 [project]
-name = "Fiona"
+name = "fiona"
 dynamic = ["readme", "version"]
 authors = [
     {name = "Sean Gillies"},
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -38,13 +38,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9', '3.10']
-        gdal-version: ['3.5.3', '3.6.2']
         include:
-          - python-version: '3.9'
+          - python-version: '3.7'
+            gdal-version: '3.4.3'
+          - python-version: '3.8'
             gdal-version: '3.4.3'
           - python-version: '3.9'
-            gdal-version: '3.3.3'
+            gdal-version: '3.5.3'
+          - python-version: '3.10'
+            gdal-version: '3.6.4'
+          - python-version: '3.11'
+            gdal-version: '3.7.1'
+          - python-version: '3.12'
+            gdal-version: '3.7.1'
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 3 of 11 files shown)


jorisvandenbossche:
Thanks for looking into this!

I also see an error, which can be reproduced as:

```
In [3]: from fiona.model import Geometry

In [4]: Geometry(type="Point", coordinates=(1.0, 2.0), geometries=[])
Out[4]: <fiona.model.Geometry at 0x7f8f1ac8f590>

In [5]: Geometry(type="Point", coordinates=(1.0, 2.0), geometries=[]).__geo_interface__
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[5], line 1
----> 1 Geometry(type="Point", coordinates=(1.0, 2.0), geometries=[]).__geo_interface__

File ~/miniconda3/envs/geo-dev2/lib/python3.11/site-packages/fiona/model.py:244, in Geometry.__geo_interface__(self)
    242 @property
    243 def __geo_interface__(self):
--> 244     return ObjectEncoder().default(self)

File ~/miniconda3/envs/geo-dev2/lib/python3.11/site-packages/fiona/model.py:365, in ObjectEncoder.default(self, o)
    363 def default(self, o):
    364     if isinstance(o, Object):
--> 365         o_dict = {k: self.default(v) for k, v in o.items()}
    366         if isinstance(o, Geometry):
    367             if o.type == "GeometryCollection":

File ~/miniconda3/envs/geo-dev2/lib/python3.11/site-packages/fiona/model.py:365, in <dictcomp>(.0)
    363 def default(self, o):
    364     if isinstance(o, Object):
--> 365         o_dict = {k: self.default(v) for k, v in o.items()}
    366         if isinstance(o, Geometry):
    367             if o.type == "GeometryCollection":

File ~/miniconda3/envs/geo-dev2/lib/python3.11/site-packages/fiona/model.py:375, in ObjectEncoder.default(self, o)
    373     return o_dict
    374 else:
--> 375     return JSONEncoder().default(o)

File ~/miniconda3/envs/geo-dev2/lib/python3.11/json/encoder.py:180, in JSONEncoder.default(self, o)
    161 def default(self, o):
    162     """Implement this method in a subclass such that it returns
    163     a serializable object for ``o``, or calls the base implementation
    164     (to raise a ``TypeError``).
   (...)
    178 
    179     """
--> 180     raise TypeError(f'Object of type {o.__class__.__name__} '
    181                     f'is not JSON serializable')

TypeError: Object of type tuple is not JSON serializable

```

fiona/model.py
```diff
@@ -384,15 +384,15 @@ class ObjectEncoder(JSONEncoder):
...
-        if isinstance(o, (Geometry, Properties)):
-            return {k: self.default(v) for k, v in o.items() if v is not None}
-        elif isinstance(o, Feature):
-            o_dict = dict(o)
-            o_dict["type"] = "Feature"
-            if o.geometry is not None:
-                o_dict["geometry"] = self.default(o.geometry)
-            if o.properties is not None:
-                o_dict["properties"] = self.default(o.properties)
+        if isinstance(o, Object):
```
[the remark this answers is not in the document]
sgillies: `Object` is `fiona.model.Object`, not `object`.

pyproject.toml
```diff
@@ -1,13 +1,14 @@
 [build-system]
 requires = [
-    "Cython~=0.29.29",
+    "cython~=3.0.2",
```
adamjstewart, line 3: The package seems to build fine with cython 0, why is 3.0.2+ required?

sgillies: @adamjstewart I don't want to support different versions of Cython at the same time, that's all.

adamjstewart: Makes sense. It's not a big deal for pip but can be tricky for other package managers like spack. We don't yet have cython 3.0.2 packaged as it's quite new.

pyproject.toml
```diff
@@ -1,13 +1,14 @@
...
     "oldest-supported-numpy",
-    "setuptools>=61",
+    "setuptools>=67.8",
+    "wheel",
```
QuLogic, line 6: I don't think `wheel` is an actual build requirement. It's necessary _if you're building wheels_, but not for building Fiona itself. `wheel` should only be installed wherever needs to build a wheel (and this is automatic for e.g., `cibuildwheel` or using `build` with the default build isolation).

adamjstewart: Correct, PyPA used to recommend this but it was removed a while back: https://github.com/pypa/packaging.python.org/pull/1050

PR #1282, opened 2023-07-18
Add gdal 3.7 to CI

rbuffat:
GDAL 3.7 was released some time ago. Let's add it to the CI and see if there is a need to change drvsupport.py.


sgillies:
Thanks @rbuffat !

PR #1287, opened 2023-08-12
Use America/Denver instead of US/Mountain to fix pytz.exceptions.UnknownTimeZoneError.

sebastic:
tzdata (2023c-8) in Debian no longer provides the US timezones:
> ```
>    * Ship only timezones in tzdata that follow the current rules of geographical
>      region (continent or ocean) and city name. Move all legacy timezone symlinks
>      (that are upgraded during package update) to tzdata-legacy. This includes
>      dropping the special handling for US/* timezones. (Closes: #1040997)
> ```


mwtoews (approved):
Thanks!

PR #1288, opened 2023-08-17
Remove dated comment about Natural Earth encoding

lgolston:
There are two references to the Natural Earth datasets having Windows-1252 encoding, which are removed here since those transitioned to UTF-8 some years ago.


mwtoews (approved):
Thanks!

PR #1295, opened 2023-09-23
remove last bits of python2 / six

a-detiste:
this is a folowup over:

https://github.com/Toblerity/Fiona/pull/985/files


a-detiste:
(there must have been some collision because at some spots I even re-do what the previous PR did before)

sgillies (approved):
Thank you @a-detiste !

PR #1297, opened 2023-09-25
test: sort listdir before comp

akrherz:
This was missed with #1193 , sorry for the misfire on my previous PR


sgillies (approved):
Thanks @akrherz !

PR #1304, opened 2023-10-15
Normalize recursive extra requirements

QuLogic:
Since the package name is now all-lowercase, normalize the self-referential extras for consistency.


sgillies (approved):
Thanks @QuLogic !

Issue #1306, opened 2023-10-18
Broken online doc


loicdtx:
Looks like the API documentation on ReadTheDocs is broken; [here](https://readthedocs.org/projects/fiona/builds/21407518/) is the latest working build I could find 

The sphinx logs of more recent build are full of
```
WARNING: autodoc: failed to import module 'collection' from module 'fiona'; the following exception was raised:
No module named 'fiona'
```

sgillies:
@loicdtx recently, RTD stopped provisioning builds with common packages, and we needed to add sphinx_rtd_theme specifically. This has been fixed in the maint-1.9 branch and now that I've merge 1.9.5 to the master branch, I expect it will be fixed there as well.

sgillies:
Yep, there we go: https://readthedocs.org/projects/fiona/builds/22283040/.

loicdtx:
From the build logs it looks like fiona itself does not get installed.

How about adding this to `.readthedocs.yaml` (I just looked it up in rasterio)?

```yaml
python:
  install:
    - method: setuptools
      path: .
```

sgillies:
@loicdtx Do you have a minute to make a PR?

PR #1307, opened 2023-10-19
Attempt to fix the broken online documentation

loicdtx:
The "API documentation" section of the online documentation uses sphinx autodoc, which requires fiona to be install in the sphinx build environment. Apparently this is needs to be done explicitly via `.readthedocs.yaml` conf file. This is what this PR does

#1306

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,6 +6,8 @@ dependencies:
 - pip
 - python=3.9.*
 - libgdal=3.4.*
+- geos=3.10.3
+- cython=3
 - sphinx-click
 - sphinx-rtd-theme
 - pip:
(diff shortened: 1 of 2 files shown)


loicdtx:
The docs from that last build look fine. I added geos but wondering whether it's really required. (I copied from the unit testing environment)

environment.yml
```diff
@@ -6,6 +6,8 @@ dependencies:
 - pip
 - python=3.9.*
 - libgdal=3.4.*
+- geos
```
sgillies, line 9: @loicdtx the addition of geos doesn't hurt, but I'm sure it's already a dependency of libgdal.

sgillies (commented):
@loicdtx thank you! I'll merge later today. You can leave or remove the geos dependency, it's up to you.

PR #1314, opened 2023-12-02
Fix leak in set_proj_search_path

QuLogic:
The `CSLAddString` function takes a `char *`, not a `const char *`, which may turn into an error on new GCC. It also allocates a new array when given `NULL`, which is _copied_ by `OSRSetPROJSearchPaths`, so it is leaked.


sgillies (approved):
Thanks @QuLogic !

sgillies:
@sgillies Note to self to backport this.

PR #1315, opened 2023-12-03
Fix compiler warnings from extensions

QuLogic:
The fixes 3 sets of warnings:
- An unused `__hash__` function, caused by a duplicated definition on the Cython side.
- Various `const char*` -> `char*` warnings.
- A maybe-uninitialized variable warning.

See commit message for details.


sgillies:
@sgillies note to port to maint-1.9.

PR #1331, opened 2024-02-29
Add python opener from rasterio 1.4

### LINKED ISSUE #1328: Bring over Rasterio's python opener
https://rasterio.readthedocs.io/en/latest/topics/vsi.html#python-file-openers and figure out sidecar/sibling files too https://github.com/rasterio/rasterio/issues/3027.

sgillies:
And supporting changes:

* Modern _path module and updated usage
* A setup.cfg for editable installs
* A Dockerfile and Makefile for testing

See also https://github.com/rasterio/rasterio/pull/3032
Resolves #1328


Dockerfile
```diff
@@ -0,0 +1,30 @@
+ARG GDAL=ubuntu-small-3.6.4
```
sgillies, line 1: New default.

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+COPY requirements*.txt ./
+RUN python${PYTHON_VERSION} -m venv /venv && \
+    /venv/bin/python -m pip install -U build pip && \
+    /venv/bin/python -m pip install -r requirements-dev.txt && \
+    /venv/bin/python -m pip list
+
+FROM gdal
+COPY . .
+RUN /venv/bin/python -m build -o wheels
+RUN /venv/bin/python -m pip install --no-index -f wheels fiona[test]
```
sgillies, line 28: It's been good to have a chance to revisit the Dockerfile. This is the way to build and install packages now.

setup.cfg
```diff
@@ -0,0 +1,17 @@
...
+    collect = fiona.fio.collect:collect
+    distrib = fiona.fio.distrib:distrib
+    dump = fiona.fio.dump:dump
+    env = fiona.fio.env:env
+    filter = fiona.fio.filter:filter
+    info = fiona.fio.info:info
+    insp = fiona.fio.insp:insp
+    load = fiona.fio.load:load
+    ls = fiona.fio.ls:ls
+    rm = fiona.fio.rm:rm
```
sgillies, line 17: This duplicates config in pyproject.toml but is necessary. Without a setup.cfg we can't do `python setup.py develop` (via pip). And that's the only way we can have fast local tests with `make dockertest`.

tests/test_pyopener.py
```diff
@@ -0,0 +1,27 @@
...
+
+def test_opener_fsspec_zip_fs():
+    """Use fsspec zip filesystem as opener."""
+    fs = fsspec.filesystem("zip", fo="tests/data/coutwildrnp.zip")
+    with fiona.open("coutwildrnp.shp", opener=fs) as colxn:
+        profile = colxn.profile
+        assert profile["driver"] == "ESRI Shapefile"
+        assert len(colxn) == 67
+        assert colxn.schema["geometry"] == "Polygon"
+        assert "AGBUR" in colxn.schema["properties"]
```
sgillies, line 27: It works!

fiona/_vsiopener.pyx
```diff
@@ -0,0 +1,491 @@
...
+        return <void *>file_obj
+
+
+cdef int pyopener_eof(void *pFile) with gil:
+    cdef object file_obj = <object>pFile
+    if file_obj.read(1):
+        file_obj.seek(-1, 1)
+        return 1
+    else:
+        return 0
```
sgillies, line 253: Turns out that Shapefile needs the eof callback.

sgillies:
The test failures are with Python 3.7 and 3.8, which are being left behind by Numpy. Fiona 1.10 will drop support for them, too.

pyproject.toml
```diff
@@ -41,6 +35,7 @@ dependencies = [
     "click-plugins>=1.0",
     "cligj>=0.5",
     'importlib-metadata;python_version<"3.10"',
+    "numpy>=1.25,<2",
```
jorisvandenbossche, line 38: It's not _needed_ to require a numpy 1.25 as a minimum version. The goal of _building_ with a minimum version of 1.25 is that this ensures it also runs with older numpys  
(in contrast to older versions of numpy, where building with that specific version meant that the wheels were only compatible with that version or up, hence the use of "oldest-supported-numpy" for wheel building)

See https://numpy.org/doc/stable/release/1.25.0-notes.html#compiling-against-the-numpy-c-api-is-now-backwards-compatible-by-default in the 1.25 release notes, and https://numpy.org/doc/stable/dev/depending_on_numpy.html#build-time-dependency

pyproject.toml
```diff
@@ -1,10 +1,5 @@
 [build-system]
-requires = [
-    "cython~=3.0.2",
-    "oldest-supported-numpy",
-    "setuptools>=67.8",
-    "wheel",
-]
+requires = ["setuptools>=67.8", "cython~=3.0.2", "numpy>=1.25,<2"]
```
jorisvandenbossche, line 2: Another question, seeing numpy included here: do you know for what reason fiona actually did have a build requirement on numpy before? A search in the codebase doesn't actually show any use of numpy (before this PR)

It seems this was brought in with https://github.com/Toblerity/Fiona/commit/37da85db8858a67a8a06451013af9aafd956ea33, which copied some code from rasterio (and rasterio indeed depends on numpy, but AFAIK Fiona does not?)

fiona/_vsiopener.pyx
```diff
@@ -0,0 +1,491 @@
...
+    cdef int num_bytes = len(python_data)
+    # NOTE: We have to cast to char* first, otherwise Cython doesn't do the conversion properly
+    memcpy(pBuffer, <void*><char*>python_data, num_bytes)
+    return <size_t>(num_bytes / nSize)
+
+
+cdef size_t pyopener_write(void *pFile, void *pBuffer, size_t nSize, size_t nCount) with gil:
+    cdef object file_obj = <object>pFile
+    buffer_len = nSize * nCount
+    cdef np.uint8_t [:] buff_view = <np.uint8_t[:buffer_len]>pBuffer
```
jorisvandenbossche, line 279: As far as I can see, this is the only use of numpy in this PR. I _think_ this could be easily avoided by using a builtin cython type instead of a numpy type. Something like `unsigned char` instead of `np.uint8_t`

That would also avoid having to add numpy as a build dependency to start with?

sgillies: @jorisvandenbossche yes, I'll definitely look into removing the hard numpy requirement before the next pre-release.

fiona/__init__.py
```diff
@@ -60,7 +62,7 @@
     _remove,
     _remove_layer,
 )
-from fiona.path import ParsedPath, parse_path, vsi_path
+from fiona._path import _ParsedPath, _UnparsedPath, _parse_path, _vsi_path
```
jorisvandenbossche, line 65: FWIW, I see you added a deprecation warning inside the `fiona.path` module. But the above change directly breaks code that was using those functions from top-level `fiona` instead of from `fiona.path`

We are running into some issues with this in geopandas: https://github.com/geopandas/geopandas/issues/3207

sgillies: I commented at https://github.com/geopandas/geopandas/issues/3207. Fiona's path module and methods are only public by accident. Other projects shouldn't be coupled to fiona's internals, which are full of weird details. Removing this coupling between geopandas and fiona shouldn't be hard and I'm happy to help.

PR #1358, opened 2024-03-12
Add an "int16" property type

### LINKED ISSUE #1321: Write columns with "short" type in gdb with OpenFileGDB driver
Hello,

I am currently struggling to write a column with a `short` type in a GDB.
After some research, I found a way to write shapefiles with a difference between long a short integers with [this](https://stackoverflow.com/questions/62998994/geopandas-export-changing-data-type/68232025#68232025) solution (which works).

Both  `'int32:4'` and  `'int32:10'` types in schema give a `long` column in a GDB, instead of giving a `short` and a `long` column.

### Code
```python
import geopandas as gpd
import fiona

gdb_path = "my_gdb.gdb"
layer = "B1_observed_event_a"
observed_event = gpd.read_file(gdb_path , layer=layer)
with fiona.collection(gdb_path, layer=layer) as source:
     s = source.schema

s["properties"]["obj_desc"] = "int32:4"
s["properties"]["det_method"] = "int32:10"

observed_event.to_file("my_gdb_copy.gdb", layer=layer, driver="OpenFileGDB", schema=s)
```
PS: I know I am using geopandas, but they implement seamlessly the Fiona schemas, this is why I am iting the issue here.

### Output 
2024-01-10_11h52_50

## Operating system

Windows

Conda (conda-forge)

sgillies:
Resolves #1321

I might like to improve how the project enumerates its data types too.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -8,6 +8,8 @@ All issue numbers are relative to https://github.com/Toblerity/Fiona/issues.
 
 Bug fixes:
 
+- Add a 16-bit integer type "int16" based on OGR's OSFTInt16 integer sub-type
+  (#).
 - Allow a GeoJSON collection's layer name to be set on opening in write mode
   (#1352).
 - The legacy crs.py module which was shadowed by the new crs.pyx module has
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -9,7 +9,7 @@ All issue numbers are relative to https://github.com/Toblerity/Fiona/issues.
 Bug fixes:
 
 - Add a 16-bit integer type "int16" based on OGR's OSFTInt16 integer sub-type
-  (#).
+  (#1358).
 - Allow a GeoJSON collection's layer name to be set on opening in write mode
   (#1352).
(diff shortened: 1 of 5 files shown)


sgillies:
@mwtoews any thoughts on this?

mwtoews:
Sure, some questions / remarks:
- Should `int16` be described in [this doc](https://fiona--1358.org.readthedocs.build/en/1358/manual.html#keeping-schemas-simple) or specifically [`fiona.FIELD_TYPES_MAP`](https://github.com/Toblerity/Fiona/blob/a5dab31158599626c78b5246e760dd5a38c4c3e6/fiona/schema.pyx#L43)? (also `bool` is missing here too, so I'm not entirely sure if this subdata type should be added there.)
- Currently `fiona.schema.normalize_field_type("int16")` returns `int32`; should it? The same function with `bool` returns `bool`.

sgillies:
@mwtoews thank you! I'll check on the function, that looks like a bug.

sgillies:
@mwtoews I'm going to address your concerns in a follow up PR.

PR #1362, opened 2024-03-18
Bring onboard filter, map, and reduce from fio-planet

### LINKED ISSUE #1327: Add commands from fio-planet
I wrote https://fio-planet.readthedocs.io/en/latest/ and it might be a good idea to bring them into this project for care and feeding.

sgillies:
With some refactoring and conditional enablement of commands that require shapely.

TODO:

- [x] deprecation of old style filter expressions
- [x] docs

Resolves #1327


pyproject.toml
```diff
@@ -35,11 +35,13 @@ dependencies = [
...
     "cligj>=0.5",
     'importlib-metadata;python_version<"3.10"',
+    "pyparsing",
+    "shapely",
 ]
 
 [project.optional-dependencies]
 all = ["fiona[calc,s3,test]"]
-calc = ["shapely"]
+calc = ["pyparsing", "shapely"]
```
mwtoews, line 44: This line can be removed since these are project dependencies (doing `pip install fiona[calc]` would just show a warning).

sgillies: @mwtoews making them core project requirements in this PR was me being sloppy. I've now moved them properly to the "calc" extras set and followed up in code and docs. I imagine there are still some few folks who want to continue to use fiona without shapely.

sgillies:
@mwtoews do you have any comments on the new documentation? I think it looks okay and that this PR is ready for a final review. I feel good about allowing the new and old filter expressions to live side-by-side for a while.

https://fiona--1362.org.readthedocs.build/en/1362/cli.html#filter
https://fiona--1362.org.readthedocs.build/en/1362/cli.html#expressions-and-functions
https://fiona--1362.org.readthedocs.build/en/1362/cli.html#sizing-up-and-simplifying-shapes

sgillies:
@snorfalorpagus @perrygeo any objections to this?

perrygeo (approved):
Snuggs for vector data! Looks good and provides a better DSL than the old python expressions. I'm curious too see how far people can push this... writing an entire analysis pipeline with snuggs seems feasible.

Some Lisp programmers might have a heart attack at the optional outer parens but I think it makes the CLI easier on the eyes. +1

I'm not 100% clear on the logic for differentiating the type of expression. I'm imagining a case where the syntax is ambiguous and a previously valid python expression gets misinterpreted as a snuggs expression. I don't have a good example though, probably not an issue.

Looks great!

PR #1366, opened 2024-03-29
Fiona <-> OGR field mappings take 2 (refactor)

sgillies:
Follow up on #1358.

The project's original mappings predated OGR's sub-types (https://gdal.org/development/rfc/rfc50_ogr_field_subtype.html) and those were kinda bolted on. In this version we embrace them. We map `(fieldtype, subtype)` tuples to Fiona field types.

The extremely gross sections of `if/else` conditionals in the OGR and Fiona feature builders have been reduced to looking up a class in one of the maps and calling one of its methods. The increase in lines of code is largely explained by the new small classes and lack of normalization.

Original tests pass, save one, which had the wrong expected value.

Benchmarking: this script runs in 0.49 seconds with Fiona 1.9.6 and 0.44 seconds with the code in this PR. Measured by cProfile. So, a ~10% speedup for a 288 record dataset with features that have 60+ fields.

```python
import fiona

with fiona.open(
    "ne_10m_admin_0_countries_lakes/ne_10m_admin_0_countries_lakes.shp"
) as src:
    with fiona.open("/tmp/test_perf.shp", "w", **src.profile) as dst:
        dst.writerecords(src)
```

This PR is about making the code more readable and easier to maintain. A small speedup is a nice bonus.

When reviewing, use the "split" view, it's much easier to read for these particular changes.


fiona/ogrext.pyx
```diff
@@ -215,7 +205,142 @@ class TZ(datetime.tzinfo):
...
+                val = string_list[string_list_index]
+                try:
+                    val = val.decode(encoding)
+                except UnicodeDecodeError:
+                    log.warning(
+                        "Failed to decode %s using %s codec", val, encoding
+                    )
+                vals.append(val)
+                string_list_index += 1
+        return vals
```
sgillies, line 343: Big TODO: consider moving these methods to the fiona type classes?

fiona/ogrext.pyx
```diff
@@ -541,6 +603,94 @@ def featureRT(feat, collection):
...
+cdef class BinaryFieldGetter(AbstractFieldGetter):
+    cdef object get(self, OGRFieldDefnH fdefn):
+        return "bytes"
+
+
+cdef class StringListFieldGetter(AbstractFieldGetter):
+    cdef object get(self, OGRFieldDefnH fdefn):
+        return "List[str]"
+
+
```
sgillies, line 693: Big TODO: consider moving these methods to the fiona type classes?

fiona/schema.pyx
```diff
@@ -70,13 +150,11 @@ def normalize_field_type(ftype):
...
         An element from FIELD_TYPES
     """
-    if ftype in FIELD_TYPES:
+    if ftype in FIELD_TYPES_NAMES:
         return ftype
-    elif ftype == 'bool':
-        return 'bool'
     elif ftype.startswith('int'):
         width = int((ftype.split(':')[1:] or ['0'])[0])
-        if GDAL_VERSION_NUM >= 2000000 and (width == 0 or width >= 10):
```
sgillies, line 79: We require GDAL 3.4+

tests/test_collection.py
```diff
@@ -535,6 +535,7 @@ def test_shape_driver_rounds_float_property_to_requested_digits_number(self):
             assert 12.2 == rf1.properties["property1"]
             assert 12.9 == rf2.properties["property1"]
 
+    @pytest.mark.xfail(reason="Corner case not supported by new feature builder implementation")
```
sgillies, line 538: Currently, I'm xfail-ing 4 tests that assert some weird behavior. For example, I'm not sure that Fiona should be ensuring that when you write `"3.14"` (a string) to an int field, you get a `3` (int) in your shapefile. This would require wrapping the setter's input value like `int(float(value))` and that's not negligible:

```
python3 -m timeit "int(3.14)"
10000000 loops, best of 5: 26.8 nsec per loop
python3 -m timeit "int(float('3.14'))"
5000000 loops, best of 5: 53.5 nsec per loop
```

tests/test_props.py
```diff
@@ -24,9 +28,10 @@ def test_types():
     assert prop_type("str") == str
     assert isinstance(0, prop_type("int"))
     assert isinstance(0.0, prop_type("float"))
-    assert prop_type("date") == FionaDateType
+    assert prop_type("date") == str
```
sgillies, line 31: `FionaDateType` no longer derives from `str`.

tests/test_schema.py
```diff
@@ -200,15 +200,18 @@ def test_normalize_float():
...
 def generate_field_types():
     """
     Produce a unique set of field types in a consistent order.
 
     This ensures that tests are able to run in parallel.
     """
-    types = set(FIELD_TYPES)
-    types.remove(None)
-    return list(sorted(types)) + [None]
+    types = set(NAMED_FIELD_TYPES.keys())
```
sgillies, line 213: `FIELD_TYPES` remains, but internal usage now uses the more complete `NAMED_FIELD_TYPES` to get, for example `FionaBooleanType` by its name `"bool"`.

fiona/ogrext.pyx
```diff
@@ -215,7 +205,351 @@ class TZ(datetime.tzinfo):
...
+        """Get the short Fiona field name corresponding to the OGR field definition."""
+        raise NotImplementedError
+
+    cdef object get(self, OGRFeatureH feature, int i, object kwds):
+        """Get the value of a feature's field."""
+        raise NotImplementedError
+
+    cdef set(self, OGRFeatureH feature, int i, object value, object kwds):
+        """Set the value of a feature's field."""
+        raise NotImplementedError
```
sgillies, line 227: The classes that derive from this one replace the ugly if/def statements in the feature builders.

fiona/ogrext.pyx
```diff
@@ -225,7 +559,37 @@ cdef class FeatureBuilder:
     argument is not destroyed.
     """
 
-    cdef build(self, void *feature, encoding='utf-8', bbox=False, driver=None, ignore_fields=None, ignore_geometry=False):
+    cdef object driver
+    cdef object property_getter_cache
```
sgillies, line 563: I've added some state to the class so that we can reuse field getter and setters.

fiona/ogrext.pyx
```diff
@@ -248,142 +612,65 @@ cdef class FeatureBuilder:
...
-                    props[key] = bool(OGR_F_GetFieldAsInteger(feature, i))
-                else:
-                    props[key] = OGR_F_GetFieldAsInteger(feature, i)
-
-            elif fieldtype is int:
-                if fieldsubtype == OFSTBoolean:
-                    props[key] = bool(OGR_F_GetFieldAsInteger64(feature, i))
+            else:
+                if fieldkey in self.property_getter_cache:
+                    getter = self.property_getter_cache[fieldkey]
```
sgillies, line 656: Here we try to reuse an existing field getter.

fiona/ogrext.pyx
```diff
@@ -248,142 +612,65 @@ cdef class FeatureBuilder:
...
-                            val = val.decode(encoding)
-                        except UnicodeDecodeError:
-                            log.warning(
-                                "Failed to decode %s using %s codec", val, encoding
-                            )
-                        props[key].append(val)
-                        string_list_index += 1
-            else:
-                props[key] = None
+                props[key] = getter.get(feature, i, {"encoding": encoding})
```
sgillies, line 669: All that old crap replaced by this 😄

fiona/ogrext.pyx
```diff
@@ -429,89 +739,32 @@ cdef class OGRFeatureBuilder:
...
+                            fieldkey
+                        )
+                        continue
+
+                # Special case: serialize dicts to assist OGR.
+                if isinstance(value, dict):
+                    value = json.dumps(value)
+
+                setter.set(cogr_feature, i, value, {"encoding": encoding})
+
```
sgillies, line 767: I'm immoderately pleased about this.

fiona/ogrext.pyx
```diff
@@ -1367,14 +1597,15 @@ cdef class WritingSession(Session):
...
+        cdef OGRSFDriverH cogr_driver
+        cdef OGRFeatureH cogr_feature
         cdef int features_in_transaction = 0
-        cdef void *cogr_layer = self.cogr_layer
+        cdef OGRLayerH cogr_layer = self.cogr_layer
 
         if cogr_layer == NULL:
             raise ValueError("Null layer")
 
+        cdef OGRFeatureBuilder feat_builder = OGRFeatureBuilder(driver=collection.driver)
```
sgillies, line 1608: Here's an optimization that I've overlooked for a while: reusing the feature builder.

tests/test_collection.py
sgillies: Oh, now I understand the difference between the new and old code. Previously, we dispatched to `OGR_F_SetFieldXXX()` based on the type of the value, not on the type of the field. I may try that.

fiona/ogrext.pyx
```diff
@@ -689,18 +689,36 @@ cdef class OGRFeatureBuilder:
...
+        (OFTDate, OFSTNone, "str"): DateField,
+        (OFTTime, OFSTNone, "time"): TimeField,
+        (OFTTime, OFSTNone, "str"): TimeField,
+        (OFTDateTime, OFSTNone, "datetime"): DateTimeField,
+        (OFTDateTime, OFSTNone, "str"): DateTimeField,
+        (OFTBinary, OFSTNone, "bytes"): BinaryField,
+        (OFTBinary, OFSTNone, "bytearray"): BinaryField,
+        (OFTBinary, OFSTNone, "memoryview"): BinaryField,
+        (OFTStringList, OFSTNone, "list"): StringListField,
+        (OFTString, OFSTJSON, "dict"): JSONField,
```
sgillies, line 721: Dispatching to a field setter based on field type and value type restores the implicit behavior in fiona version 1.9.x.

sgillies:
@mwtoews @snorfalorpagus @perrygeo I'm tagging you for review of this refactor PR. Meanwhile I'm going to work on benchmarking.

sgillies:
Last call for review!

PR #1377, opened 2024-04-15
Restore behavior of setting fields to values derived from date, time

### LINKED ISSUE #1376: Fiona no longer recognizes pd.Timestamp (as datetime.datetime) since 1.10 a2
Example that works with released fiona, using a `pd.Timestamp` scalar in the data passed to `writerecords` for a "datetime" field:

```
In [1]: import fiona

In [2]: fiona.__version__
Out[2]: '1.9.6'

In [3]: schema = {'geometry': 'Point', 'properties': {'a': 'float', 'b': 'datetime'}}

In [4]: with fiona.open("test_fiona110.geojson", mode="w", schema=schema) as col:
   ...:     col.writerecords([{'id': '0', 'type': 'Feature', 'properties': {'a': 1.0, 'b': pd.Timestamp('2021-11-21 01:07:43.018000')}, 'geometry': {'type': 'Point', 'coordinates': (0.0, 0.0)}}])
   ...: 

In [5]: !cat test_fiona110.geojson
{
"type": "FeatureCollection",
"features": [
{ "type": "Feature", "properties": { "a": 1.0, "b": "2021-11-21T01:07:43.018" }, "geometry": { "type": "Point", "coordinates": [ 0.0, 0.0 ] } }
]
}
```

Starting with 1.10 a2, this seems to silently drop that column:

```
In [1]: import fiona

In [2]: fiona.__version__
Out[2]: '1.10a2'

In [3]: schema = {'geometry': 'Point', 'properties': {'a': 'float', 'b': 'datetime'}}

In [4]: with fiona.open("test_fiona110.geojson", mode="w", schema=schema) as col:
   ...:     col.writerecords([{'id': '0', 'type': 'Feature', 'properties': {'a': 1.0, 'b': pd.Timestamp('2021-11-21 01:07:43.018000')}, 'geometry': {'type': 'Point', 'coordinates': (0.0, 0.0)}}])
   ...: 

In [5]: !cat test_fiona110.geojson
{
"type": "FeatureCollection",
"name": "test_fiona110",
"features": [
{ "type": "Feature", "properties": { "a": 1.0 }, "geometry": { "type": "Point", "coordinates": [ 0.0, 0.0 ] } }
]
}
```

If I use `datetime.datetime.fromisoformat(..)` instead of `pd.Timestamp(..)` in the code above (i.e. passing a `datetime.datetime` object), then it works again. 
But `pd.Timestamp` is subclass of `datetime.datetime` (and so things like `isinstance(value, datetime.datetime)` should work), and this worked in the past.

We are running into this issue in GeoPandas' CI because we pass the result of `GeoDataFrame.iterfeatures()` to `writerecords`. And it seems that this is creating a json-like structure that uses `pd.Timestamp` instead of `datetime.datetime` objects.

sgillies:
Resolves #1376

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -760,26 +760,30 @@ cdef class OGRFeatureBuilder:
                 OGR_F_SetFieldNull(cogr_feature, i)
             else:
                 schema_type = session._schema_normalized_field_types[key]
-                fieldkey = (*FIELD_TYPES_MAP2[NAMED_FIELD_TYPES[schema_type]], type(value).__name__)
-                if fieldkey in self.property_setter_cache:
-                    setter = self.property_setter_cache[fieldkey]
+                val_type = type(value)
+
+                if val_type in self.property_setter_cache:
+                    setter = self.property_setter_cache[val_type]
                 else:
-                    try:
-                        setter = self.OGRPropertySetter[fieldkey](driver=self.driver)
-                        self.property_setter_cache[fieldkey] = setter
-                    except KeyError:
-                        log.warning(
-                            "Skipping field %s: invalid type %s",
-                            key,
-                            fieldkey
-                        )
+                    for cls in val_type.mro():
+                        fieldkey = (*FIELD_TYPES_MAP2[NAMED_FIELD_TYPES[schema_type]], cls.__name__)
+                        try:
+                            setter = self.OGRPropertySetter[fieldkey](driver=self.driver)
+                        except KeyError:
+                            continue
+                        else:
+                            self.property_setter_cache[val_type] = setter
+                            break
+                    else:
+                        log.warning("Skipping field because of invalid value: key=%r, value=%r", key, value)
                         continue
 
                 # Special case: serialize dicts to assist OGR.
                 if isinstance(value, dict):
                     value = json.dumps(value)
 
-                log.debug("Setting feature property: fieldkey=%r, setter=%r, i=%r, value=%r", fieldkey, setter, i, value)
+                log.debug("Setting feature property: key=%r, value=%r, i=%r, setter=%r", key, value, i, setter)
+
                 setter.set(cogr_feature, i, value, {"encoding": encoding})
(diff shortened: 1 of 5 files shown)


tests/test_datetime.py
```diff
@@ -827,3 +827,30 @@ def test_read_timezone_geojson(path_test_tz_geojson):
...
+
+def test_property_setter_lookup(tmp_path):
+    """Demonstrate fix for #1376."""
+
+    class MyDate(datetime.date):
+        pass
+
+    feat = Feature.from_dict(
+        {
+            "properties": {"when": MyDate(2024, 4, 15)},
```
sgillies, line 840: @jorisvandenbossche I believe this is for testing purposes equivalent to the situation with `pd.Timestamp`, yes?

fiona/ogrext.pyx
```diff
@@ -760,26 +760,30 @@ cdef class OGRFeatureBuilder:
                 OGR_F_SetFieldNull(cogr_feature, i)
             else:
                 schema_type = session._schema_normalized_field_types[key]
-                fieldkey = (*FIELD_TYPES_MAP2[NAMED_FIELD_TYPES[schema_type]], type(value).__name__)
-                if fieldkey in self.property_setter_cache:
-                    setter = self.property_setter_cache[fieldkey]
+                val_type = type(value)
+
+                if val_type in self.property_setter_cache:
+                    setter = self.property_setter_cache[val_type]
```
sgillies, line 766: In a nutshell: the code in 1.9 tested `elif isinstance(value, (date, time, datetime)` before calling a field setter. Now we register setters to names like "date" and use the MRO of `type(value)` to look up and cache a setter.

sgillies: Look okay to you @mwtoews ?

tests/test_datetime.py
jorisvandenbossche: Yes, that should cover that use case. Thanks!

PR #1385, opened 2024-04-28
Replace logging.warn usage with logging.warning

jelly:
logging.warn is an alias to logging.warning since Python 3.3 and will be removed in Python 3.13.


sgillies (approved):
Thank you @jelly ! The test failure is unrelated.

PR #1394, opened 2024-05-18
Don't attempt to test append to parquet dataset in test_append_memoryfile_drivers

manisandro:
The parquet driver does not appear to support append mode, and opening with `mode='a'` fails with

> fiona._err.CPLE_OpenFailedError: '/vsimem/f66391b5a3ba47bc93363c6b80a15e96/f66391b5a3ba47bc93363c6b80a15e96.parquet' not recognized as being in a supported file format.

Opening in regular write mode succeeds though.

### DIFF
--- a/tests/test_memoryfile.py
+++ b/tests/test_memoryfile.py
@@ -280,13 +280,20 @@ def test_append_memoryfile_drivers(driver, testdata_generator):
         with memfile.open(driver=driver, crs="OGC:CRS84", schema=schema) as c:
             c.writerecords(records1)
 
-        with memfile.open(mode='a', driver=driver, schema=schema) as c:
-            c.writerecords(records2)
+        # The parquet dataset does not seem to support append mode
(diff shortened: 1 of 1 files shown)


rbuffat:
@manisandro is this specific to memoryfile or in general? If the Parquet driver does not support append mode, it would be good to change [drvsupport.py](https://github.com/Toblerity/Fiona/blob/main/fiona/drvsupport.py#L106)  from `raw` to `rw` too.

manisandro:
For instance here [1] it states that 

> Once written, a Parquet file is immutable. No modification or appending can happen to that Parquet file. Instead, create a new Parquet file.

So I believe that it applies in general to parquet files. I've updated drvsupport.py accordingly.

[1] https://guide.cloudnativegeo.org/geoparquet/#files-are-immutable

sgillies (approved):
Thank you @manisandro !

PR #1415, opened 2024-07-24
Replace sketchy opener interfaces with ABCs in fiona.abc

sgillies:
Also adds support for GDAL VSI multi-range reads. This is an attempt to take https://github.com/rasterio/rasterio/pull/3117 all the way to a stable interface and API.

`io.open()`, fsspec filesystems, and to a lesser degree, `tiledb.VFS`, are privileged openers. There are built-in adapters for these objects.

Otherwise, users should implement `fiona.abc.FileContainer` and register their class as a virtual subclass within their application like `FileContainer.register(NewClass)`.

`fiona.abc.MultiByteRangeResourceContainer` exists for completeness, to provide a template for Rasterio, and for compatibility with future GDAL/OGR drivers that may exploit it. Currently no vectors drivers do and there is no point in implementing this interface.

### DIFF
--- a/fiona/_vsiopener.pyx
+++ b/fiona/_vsiopener.pyx
@@ -378,6 +378,16 @@ def _opener_registration(urlpath, obj):
     # Might raise.
     opener = to_pyopener(obj)
 
+    # Before returning we do a quick check that the opener will
+    # plausibly function.
+    try:
+        _ = opener.size("test")
+    except (AttributeError, TypeError, ValueError) as err:
+        raise OpenerRegistrationError(f"Opener is invalid.") from err
+    except Exception:
+        # We expect the path to not resolve.
+        pass
+
     registry = _OPENER_REGISTRY.get({})
 
     if key in registry:
--- a/fiona/_vsiopener.pyx
+++ b/fiona/_vsiopener.pyx
@@ -3,8 +3,11 @@
 Based on _filepath.pyx.
 """
 
+from abc import ABC, abstractmethod
+from collections.abc import Callable
 import contextlib
 from contextvars import ContextVar
+from functools import singledispatch
 import logging
 import os
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


tests/test_pyopener.py
```diff
@@ -161,3 +160,45 @@ def test_opener_fsspec_file_remove(data):
...
+
+        def mtime(self, path):
+            return pathlib.Path(path).stat().st_mtime
+
+        def size(self, path):
+            return pathlib.Path(path).stat().st_size
+
+    FileContainer.register(CustomContainer)
+
+    listing = fiona.listdir("tests/data", opener=CustomContainer())
```
sgillies, line 200: This is how a Fiona user creates a custom opener for their own application.

sgillies:
@vincentsarago @mwtoews @snorfalorpagus what do you think of a fiona.abc module?

Vincent, I think this is what we should do for Rasterio.

fiona/_vsiopener.pyx
```diff
@@ -353,7 +376,7 @@ def _opener_registration(urlpath, obj):
     cdef bytes prefix_bytes = f"/{namespace}/".encode("utf-8")
 
     # Might raise.
-    opener = _create_opener(obj)
+    opener = to_pyopener(obj)
```
vincentsarago, line 379: 👍

tests/test_pyopener.py
```diff
@@ -161,3 +160,45 @@ def test_opener_fsspec_file_remove(data):
...
+        def ls(self, path):
+            return list(pathlib.Path(path).iterdir())
+
+        def mtime(self, path):
+            return pathlib.Path(path).stat().st_mtime
+
+        def size(self, path):
+            return pathlib.Path(path).stat().st_size
+
+    FileContainer.register(CustomContainer)
```
vincentsarago, line 198: 🔥

PR #1441, opened 2024-09-05
docs: update text about python requirements

paulschreiber:
## Description
Update text about Python requirements. Fiona 1.9 is no longer "coming soon."

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -30,6 +30,8 @@ access to more optional format drivers (like GML).
 
 Fiona requires Python 3.7 or higher and GDAL 3.2 or higher.
 
+Fiona 2.0 (in development and unreleased) requires Python 3.8 or higher and GDAL 3.4 or higher.
+
 Advanced installation
 =====================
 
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -28,7 +28,7 @@ build and install Fiona from a source distribution.
 Many users find Anaconda and conda-forge a good way to install Fiona and get
 access to more optional format drivers (like GML).
 
-Fiona 1.9 (coming soon) requires Python 3.7 or higher and GDAL 3.2 or higher.
+Fiona requires Python 3.7 or higher and GDAL 3.2 or higher.
```
sgillies, line 31: @paulschreiber these versions are more accurate.

```suggestion
Fiona 2.0 (in development and unreleased) requires Python 3.8 or higher and GDAL 3.4 or higher.
```

When Fiona 2.0 is released, these will probably become Python 3.10 and GDAL 3.6.

sgillies (changes requested):
@paulschreiber thank you for updating this! I've suggested different versions.

sgillies (approved):

Issue #1451, opened 2024-09-20
JSONDecodeError with `fiona>=1.10.0`


veenstrajelmer:
## Expected behavior and actual behavior.
When reading a geojson with geopandas (engine fiona) from the web, I get `"JSONDecodeError: Extra data: line 1 column 2 (char 1)"`. This error only occurs with `fiona>=1.10.0`, so this only recently surfaced.

<details>
<summary>Click to expand traceback</summary>

```
JSONDecodeError                           Traceback (most recent call last)
Cell In[1], line 2
      1 import geopandas as gpd
----> 2 uhslc_gpd = gpd.read_file("https://uhslc.soest.hawaii.edu/data/meta.geojson", engine="fiona")
      3 time_min = uhslc_gpd["fd_span"].apply(lambda x: x["oldest"])

File ~\AppData\Local\miniforge3\envs\dfm_tools_env\Lib\site-packages\geopandas\io\file.py:306, in _read_file(filename, bbox, mask, columns, rows, engine, **kwargs)
    303     else:
    304         path_or_bytes = filename
--> 306     return _read_file_fiona(
    307         path_or_bytes,
    308         from_bytes,
    309         bbox=bbox,
    310         mask=mask,
    311         columns=columns,
    312         rows=rows,
    313         **kwargs,
    314     )
    316 else:
    317     raise ValueError(f"unknown engine '{engine}'")

File ~\AppData\Local\miniforge3\envs\dfm_tools_env\Lib\site-packages\geopandas\io\file.py:418, in _read_file_fiona(path_or_bytes, from_bytes, bbox, mask, columns, rows, where, **kwargs)
    414     df = pd.DataFrame(
    415         [record["properties"] for record in f_filt], columns=columns
    416     )
    417 else:
--> 418     df = GeoDataFrame.from_features(
    419         f_filt, crs=crs, columns=columns + ["geometry"]
    420     )
    421 for k in datetime_fields:
    422     as_dt = None

File ~\AppData\Local\miniforge3\envs\dfm_tools_env\Lib\site-packages\geopandas\geodataframe.py:732, in GeoDataFrame.from_features(cls, features, crs, columns)
    729     features_lst = features
    731 rows = []
--> 732 for feature in features_lst:
    733     # load geometry
    734     if hasattr(feature, "__geo_interface__"):
    735         feature = feature.__geo_interface__

File fiona\\ogrext.pyx:1996, in fiona.ogrext.Iterator.__next__()

[...]

PR #1452, opened 2024-09-22
In a weird case, don't fail utterly, but add an error message

sgillies:
Follow up on #1451.

GeoJSON feature properties are not tightly constrained. GDAL/OGR depends on constraints, and would prefer a single type for field values. When GDAL receives features where a named property (or field) may be a string, or a number, or something else, the behavior is not well defined. In versions before 3.6, it seems like GDAL interpreted these as String type fields. After, it seems that these are interpreted as String(JSON) fields.

When GDAL has determined that the field is type: String(JSON) (rightly or wrongly), if a JSONDecodeError occurs, instead of raising an exception and bringing processing to a halt, we might return a error message in place of a normal value.

I honestly have no idea what the best UX is here, I don't see obvious precedent yet.

### DIFF
--- a/fiona/ogrext.pyx
+++ b/fiona/ogrext.pyx
@@ -386,8 +386,21 @@ cdef class JSONField(AbstractField):
         return "json"
 
     cdef object get(self, OGRFeatureH feature, int i, object kwds):
+        cdef void *fdefn = OGR_F_GetFieldDefnRef(feature, i)
+        cdef char *key_c = OGR_Fld_GetNameRef(fdefn)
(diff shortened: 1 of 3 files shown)


sgillies:
Resolution of [GDAL #10902](https://github.com/OSGeo/gdal/issues/10902) and [GDAL #10943](https://github.com/OSGeo/gdal/issues/10943) means that Fiona doesn't need a special solution of its own. GDAL will make better guesses about GeoJSON field types and will allow users to override these guesses.

PR #1462, opened 2024-10-29
Small performance improvements for GeomBuilder._buildCoords

groutr:
Some minor improvements to help build coordinate lists a little more efficiently.

1. Lift the ndim check out of the loop. This only needs to be checked once per call.
2. Build tuples directly instead of creating a list and converting to tuples.

### DIFF
--- a/fiona/_geometry.pyx
+++ b/fiona/_geometry.pyx
@@ -124,15 +124,20 @@ cdef class GeomBuilder:
     cdef list _buildCoords(self, void *geom):
         # Build a coordinate sequence
         cdef int i
+        cdef int npoints
+        cdef list coords
+
         if geom == NULL:
             raise ValueError("Null geom")
         npoints = OGR_G_GetPointCount(geom)
         coords = []
-        for i in range(npoints):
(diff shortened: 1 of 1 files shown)


fiona/_geometry.pyx
```diff
@@ -124,15 +124,20 @@ cdef class GeomBuilder:
     cdef list _buildCoords(self, void *geom):
         # Build a coordinate sequence
         cdef int i
+        cdef int npoints
+        cdef list coords
```
sgillies, line 128:
```suggestion
        cdef list coords = []
```

And then delete line 133?

sgillies (approved):
@groutr thank you! I'll backport to maint-1.10 after it's merged.

I made one suggestion.

Would a list comprehension be any faster?

fiona/_geometry.pyx
groutr: Seems fine to me.

groutr:
I would be shocked if a list comprehension performed any different from a for-loop since they both generate identical C code.

sgillies (approved):

sgillies:
Back ported to maint-1.10.

PR #1467, opened 2024-11-18
TST: Mark test_opener_fsspec_zip_http_fs as using the network

QuLogic:
As with other tests that use the network.


sgillies (approved):
Thank you @QuLogic !

PR #1470, opened 2024-12-03
Fallback to raw string value, when json field inside geojson isn't decodable

jbylina:

I expected to successfully read the following geojson file by Fiona and leave handling mixed types geojson properties to upstream.
```json
{
"type": "FeatureCollection",
"features": [
{"type": "Feature", "properties": { "a": [ ]}, "geometry": { "type": "Point", "coordinates": [ [-74, 5]] } },
{"type": "Feature", "properties": { "a": ""}, "geometry": { "type": "Point", "coordinates": [ [-74, 5]] } }
]
}
```

Actual behavior: 
```
  File "<input>", line 1, in <module>
  File "fiona/ogrext.pyx", line 1996, in fiona.ogrext.Iterator.__next__
  File "fiona/ogrext.pyx", line 668, in fiona.ogrext.FeatureBuilder.build
  File "fiona/ogrext.pyx", line 390, in fiona.ogrext.JSONField.get
  File "/Users/jacek/.pyenv/versions/3.12.7-debug/lib/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jacek/.pyenv/versions/3.12.7-debug/lib/python3.12/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jacek/.pyenv/versions/3.12.7-debug/lib/python3.12/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Run the following script using the latest Fiona.

```python
from pathlib import Path
import fiona

Path("test.geojson").write_text("""\
{
"type": "FeatureCollection",
"features": [
{"type": "Feature", "properties": { "a": [ ]}, "geometry": { "type": "Point", "coordinates": [ [-74, 5]] } },
{"type": "Feature", "properties": { "a": ""}, "geometry": { "type": "Point", "coordinates": [ [-74, 5]] } }
]
}
""")

list(fiona.open("test.geojson", "r"))
```

## Operating system

macOS 15.1.1

Fiona 1.10.1 from pip
GDAL 3.10.0 installed via Homebrew

## Comment

When I flip lines, so that `[ ]` geojson property isn't first, the code doesn't fail.
```json
{
"type": "FeatureCollection",
"features": [
{"type": "Feature", "properties": { "a": ""}, "geometry": { "type": "Point", "coordinates": [ [-74, 5]] } },
{"type": "Feature", "properties": { "a": [ ]}, "geometry": { "type": "Point", "coordinates": [ [-74, 5]] } }
]
}
```

## Fix

I propose to fallback to raw string value if json field property isn't decodable.


fiona/ogrext.pyx
```diff
@@ -412,14 +412,10 @@ cdef class JSONField(AbstractField):
...
-        # String(JSON) is the fallback for GDAL in the case of
-        # properties that are a mix of strings and numbers. See gh-1451.
-        # In GDAL 3.11, users will be able to override this and declare
-        # a field type.
         try:
             return json.loads(val)
-        except json.JSONDecodeError as error:
-            raise FieldError(f"String(JSON) field could not be decoded: {val=}, {error=}")
+        except json.JSONDecodeError:
+            return val
```
sgillies, line 418:
```suggestion
        except json.JSONDecodeError as error:
            warnings.warn(f"JSON field value not decoded, returning as string: {val=}, {error=}", FeatureWarning)
            log.info("JSON field value not decoded, returning as string: val=%r, error=%r", val, error)
            return val
```

sgillies (changes requested):
@jbylina Thanks for the suggestion!

How about a log message to say that we are defaulting to different behavior? And a warning so that users can raise an exception instead if they want to be very strict.

PR #1475, opened 2024-12-27
manual: use consistent dict style

shenriotpro:
Possibly because it used to be an OrderedDict https://github.com/Toblerity/Fiona/pull/1246/files#diff-7d78d5cd5602b1e22507b370b43360c79eeec171775948ce62fa85b5a23209b8L940


shenriotpro:
@sgillies in case you'd like to review

mwtoews (approved):
Thanks, this is indeed an OrderedDict hangover.

Issue #1487, opened 2025-02-20
Fiona throws TypeError when comparing Object to values that can't be unpacked


alexgware:
## Background

Hello,
We recently did a major upgrade of Fiona in our codebase and ran into this issue. We have a default behaviour of checking imported fields for empty strings to convert to None, however this causes  a TypeError when it runs into the new Fiona Geometry object. The old behaviour was to return false when comparing the empty string to the old dict geometry representation. We've temporarily fixed this by forcing the Geometry object back to dict, but would like to ensure our codebase can work with the new native implementation as it seems like a better approach overall.

### Expected
* Fiona Object should return false when asked if it is equal to a differing object, regardless of if that object can be natively unpacked

### Actual
* Fiona Object throws a TypeError when asked if it is equal to an object that can't natively be unpacked

## Steps to reproduce the problem
```python
from fiona.model import Object

obj = Object(g=1)
assert obj != ""  # TypeError: dict() argument after ** must be a mapping, not str
```

## Versions
* Fiona==1.10.1
* GDAL==3.10.1
* python==3.10.16

## PR
Suggested fix can be found in PR here https://github.com/Toblerity/Fiona/pull/1488. Happy to discuss and/or make changes

Thanks!

PR #1488, opened 2025-02-20
Issue 1487 - Suggested Fix - Return false when comparing Object against values that can't be unpacked 

agware:
This is a suggested fix for [Issue 1847](https://github.com/Toblerity/Fiona/issues/1487)

In cases where eq is used for an object that cannot be unpacked (i.e. is not an instance of Mapping) simply default to returning false.

Depending on style/preference, another way this could be implemented is:
```python
try:
    return dict(**self) == dict(**other)
except TypeError:
    return False
```
I went with the Mapping approach as it seemed to tie in more nicely with the inheritance from MutableMapping for Object, but both are pythonic and I'm happy to switch the fix over if you have a preference.

I made the assumption that the decision to do equality explicitly against the unpacked versions of the objects was in part to maintain backwards compatibility with old style Fiona representations of geometries (or similar). So I tried to stay away from anything that would create more direct comparisons between object types (e.g. return false if it's not a dict/Object), leaving the option open for other Mapping based objects to also resolve to True when compared.

### DIFF
--- a/fiona/model.py
+++ b/fiona/model.py
@@ -1,7 +1,7 @@
 """Fiona data model"""
 
 from binascii import hexlify
-from collections.abc import MutableMapping
+from collections.abc import Mapping, MutableMapping
 from enum import Enum
 import itertools
 from json import JSONEncoder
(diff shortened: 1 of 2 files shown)


sgillies (changes requested):
Thank you @agware ! This is right on the mark and very well done.

sgillies (approved):

PR #1500, opened 2025-08-19
Remove duplicate click option

QuLogic:
The most recent click is now warning about the duplication.

PR #1509, opened 2025-12-29
Drop duplicated --where click.option for fio cat

### LINKED ISSUE #1508: Duplicated `--where` click option for `fio cat`
```
$ fio cat --help
/opt/miniconda3/envs/prod/lib/python3.13/site-packages/click/core.py:1223: UserWarning: The parameter --where is used more than once. Remove its duplicate as parameters should be unique.
  parser = self.make_parser(ctx)
```

Creating a PR now for this.

akrherz:
closes #1508


akrherz:
oops, dup of #1500

